Issue #4, opened 2014-11-05
BAP Python API


geohot:
In order to make BAP usable to the majority of people doing binary analysis today, it needs to have a functional, stable, and documented Python API.

PR #40, opened 2014-12-08
A proposal for BIL Python representation.

ivg:
This is a proposed python interface to BAP BIL for discussion with those who are interested, especially for those who are subscribed to #4. 

The API follows closely the OCaml definitions and tries to provide comparable facilities, with a `Visitor` pattern.

Although, it is not strictly a part of the interface, but I'm going to dump data from the OCaml side in a python representation, so that it 
can be loaded on the python side using `eval`. In other words, I will print BIL as valid python expression, like David has proposed. 

I've evaluated other solutions, like using json and xml, and found them to costly to use and to implement. 

Note: this is a discussion PR, I'm not going to merge it.


src/server/mmap_server.ml
```diff
@@ -0,0 +1,69 @@
...
+let url_of_file ?query sub path : Uri.t =
+  let length = Bigsubstring.length sub in
+  let offset = Bigsubstring.pos sub in
+  let url = Uri.make ~scheme ~path ~query:[
+      "length", [Int.to_string length];
+      "offset", [Int.to_string offset];
+    ] () in
+  match query with
+  | None -> url
+  | Some v -> Uri.with_query' url ["q", v]
```
ivg: bug! this erases previous query

PR #61, opened 2015-01-12
BAP Public API

ivg:
==============

Abstract
--------

This patch adds JSON Public API to BAP. It also introduces some bug
fixes and extensions. As an example of API usage a small python binding
is provided.

Infrastructure Changes
----------------------

Since BAP now has more that 50 dependencies (most are optional), opam
needs further guidance, otherwise it can't find a proper solution. This
patch adds such guidance, properly specifying packages versions.

Also, this patch fixes warnings issued by a build system on a newer
compiler version.

Core Lwt library
----------------

This patch introduces a Core Lwt library, that is a thin wrapper around
Lwt library to provide a concise Core-like interface. I'm going to
remove this library from BAP sooner or later, and push it to opam as a
separate entity.

Consistent Constructor Names
----------------------------

In order to satisfy the requirements of ADT data format, and for
consistency purposes, I've title-cased all constructor names. This is a
breaking change.

Debugging Support
-----------------

Many were extended to support `sexp_of` or `sexp` protocols. Also, all
regular types got a plethora of new primitives, that can be used with
different format specifiers.

Changes to Bitvector
--------------------

Functions `of_int32` and `of_int64` now accepts optional `width`
parameter. Also, a new `string_of_value` function was added, that
converts to string only vector value, dropping information about the
size.

Changes to Image
----------------

Switched to bap's size type instead of core's.

Python Bindings
---------------

This PR introduces a python binding to BAP. This binding allows one to
disassemble arbitrary strings, and to load and analyze binary files. The
library is packed with distutils, that means, that in order to use, you
can just `pip install` it. The binding is rather deep, that means that
instead of textly typed dictionaries you will get first-class python
value of the appropriate type. All known to us ARM operands, registers
and instructions are lifted to python classes. See documentation for
more info.

Public API
----------

BAP now can be called from any language, using JSON API. This API is
implemented in a new program called `bap-server`. The server will also
take care on storage and persistence problems, i.e., it will store date
for you. Currently, the following set of protocols are implemented:

- http
- mmap
- file
- zmq

This set allows to use the following physical media for interaction:

- tcp and udp sockets
- unix sockets
- regular files
- shared memory


python/arm.py
```diff
@@ -0,0 +1,244 @@
...
+
+class Insn(ADT)     : pass
+class Move(Insn)    : pass
+class Bits(Insn)    : pass
+class Mult(Insn)    : pass
+class Mem(Insn)     : pass
+class Branch(Insn)  : pass
+class Special(Insn) : pass
+
+class ADCri(Move) : pass
```
maurer: Where's the script to autogenerate this? If there isn't one, there should be. If there is one, it should be indicated in a comment int his file so people don't edit it.

python/adt.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python
+"""
+Algebraic Data Types (ADT) is used to represent two kinds of things:
```
maurer: This is the first way I considered doing ADTs in python too, but every professional python programmer I knew told me that discriminating using type() or similar functions for any purpose other than optimization is considered in extremely poor form.

This may be OK anyways, but the python folks are unlikely to be happy with this representation.

python/arm.py
ivg: This script sits on his chair in 2205 =) 
It is not autogenerated in other words. An actually you can even edit it, i.e. add attributes or methods.

python/bap.py
```diff
@@ -0,0 +1,342 @@
...
+                if attempt + 1 == RETRIES:
+                    raise
+                else:
+                    time.sleep(0.1 * attempt)
+
+        if not "capabilities" in self.__dict__:
+            raise RuntimeError("Failed to connect to BAP server")
+
+        self.data = {}
+        self.temp = NamedTemporaryFile('rw+b')
```
maurer: If we support shm as was stated in the README, and we're spawning on the same server, why are we not using it?

python/bil.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
+
+"""BAP BIL Python representation"""
+
+from adt import *
+
+
+class Exp(ADT)  : pass     # Abstract base for all expressions
```
maurer: Again, expect the python folk to get mad about this.

python/adt.py
ivg: "professional" "python" and "programmer" shouldn't ever be used together))

actually, they may dislike it of course, but, in this case I do not discriminate over `type`, but using the visitor. The same approach is used in python compiler libs itself, so I assumed that python programmers can hold it. What concerning, discriminating over type, `isinstance` is ubiquitous in python code base.

python/bap.py
ivg: we're using, `NamedTemporaryFile` is used only for name. `mmap` needs a file system entry to share data between processes.

python/bap.py
```diff
@@ -0,0 +1,342 @@
...
+                    msg = "Expected {0} msg but got {1}".format(
+                        self._name, msg)
+                    raise RuntimeError(msg)
+
+    def get(self, child):
+        self.load()
+        return self.msg[self._name].get(child)
+
+
+class Image(Resource):
```
maurer: Does Image/Section/Symbol really go in here?
I think it probably goes in a nested module. That module might want to be pulled up to the base namespace, but code for "Here is how we will connect to the bap server" should probably be separated from "Here are a number of translations for specific queries we want to make"

python/bap.py
```diff
@@ -0,0 +1,342 @@
...
+        mm.close()
+        return url
+
+    def _load_resource(self, res):
+        rep = self.call(res).next()
+        if 'error' in rep:
+            raise ServerError(rep)
+        return Id(rep['resource'])
+
+def spawn_server(**kwargs):
```
maurer: Why are we using HTTP by default instead of shm? We know we're on the same machine, since we're spawning the server via Popen.

python/bap.py
ivg: the `Image`/`Section`/`Symbol` are mostly helpers, that are thin wrappers on top off dictionaries. Their main purpose is to provide lazy behavior, fetching data on-demand, that's why they are in this module. Also, this is python, no one should even know about this classes, and they're indeed hidden. In the help string I've specified their attributes.

python/adt.py
```diff
@@ -0,0 +1,168 @@
...
+        def args():
+            if isinstance(self.val, tuple):
+                return ", ".join(qstr(x) for x in self.val)
+            else:
+                return qstr(self.val)
+
+        return "{0}({1})".format(self.name, args())
+
+
+class Visitor(object):
```
maurer: Why not just make ADT iterable instead?

ivg: because it is tree

python/bap.py
ivg: `HTTP` is used for control information. `shm` is used to share binary data. Currently bap supports only `HTTP` as a control information transport. It can be easily extended, but I do not see any reason to pass requests and response via `shm` or some other faster mechanisms. Actually, I've traces, most of the time we spend in python, trying to understand bap's answer, not in sending or receiving data.

lwt/core_lwt_extra.mli
```diff
@@ -0,0 +1,18 @@
+(** Overlay over some Lwt functions  *)
+open Lwt
+
+val catch : (unit -> 'a t) -> exn:(exn -> 'a t) -> 'a t
+
+val try_bind : (unit -> 'a t) -> ok:('a -> 'b t) -> exn:(exn -> 'b t) -> 'b t
+
+val protect : (unit -> 'a t) -> finally:(unit -> unit t) -> 'a t
+
+val don't_wait : 'a t -> unit
```
maurer: B|

lwt/core_lwt.ml
```diff
@@ -0,0 +1,30 @@
+open Core_kernel.Std
```
maurer: Honest question: What is the behavior of Lwt if a cooperative thread performs a blocking syscall? (I really don't know, all I know about Lwt is that it can't use multiple cores and uses cooperative threading, which normally has a problem with this.)

Worry cases:
1.) Bap as a service - need an entirely separate bap instance for every connection to avoid accidental DOS on one thing blocking.
2.) Bap server hardlocking on an operation (e.g. file open or read from a network share?) and being unable to report to the client.

ivg: Actually, the things are completely opposite here. Usage of the cooperative frameworks like `lwt` or `async` allows one to solve problems with availability in a high load environment. When cooperative thread performs a blocking call (I'm talking about a call lifted to `lwt` not to the real system call) it  will automatically yield to another thread. 

1) No, each connection to bap is very cheap, and bap can hold tons of its without worry. We should only worry on memory footprint and actual disassembly. For that case, I've added push back mechanism, that will not accept any new requests, until the older one are resolved.

2) Nope. It will respond and very fast.

PR #75, opened 2015-01-28
Import deprecated x86 lifter

feseal:
Ivan, please make code review of deprecated lifter.


_oasis
```diff
@@ -211,6 +211,7 @@ Library disasm
                  Bap_disasm_arm_shift,
                  Bap_disasm_arm_types,
                  Bap_disasm_arm_utils,
+                 Bap_disasm_x86_lifter,
```
ivg: please, keep the entries sorted. This will ease the process of merging.

lib/bap_disasm/bap_disasm_x86_lifter.mli
```diff
@@ -0,0 +1,6 @@
+open Core_kernel.Std
+open Bap_types.Std
+
+(** [insn mem basic] takes a basic instruction and a memory and
+    returns a sequence of BIL statements. *)
+val insn : Bap_memory.t -> ('a,'k) Bap_disasm_basic.insn -> stmt list Or_error.t
```
ivg: You will also need an arch parameter or a mode or something else, to distinguish between 32 and 64 addressing. I would suggest to prepend a parameter of type `addr_size` to the parameters.

PR #77, opened 2015-02-04
add viz

maverickwoo:
https://app.asana.com/0/14602640551203/25080518462222


maverickwoo:
I note that extjs, graphviz, and prettify all have their own licenses.

ivg:
1200 files, 40000 LOC. Are you kidding on me? This not what git is about. Only code that we actually develop should be on the repository.

PR #91, opened 2015-02-13
Fixes issue in lifter where register offsets were not accounted for.

tim-becker:
When working with bap instructions in Qira, I came across the following instruction that did not have BIL associated with it:
```
ldr	r2, [r3, r2]
```

The lifter was failing because it falsely assumed that offsets cannot be registers in loads. This PR fixes that bug.


ivg:
Great, now no more lifting errors! 
Thanks

PR #93, opened 2015-02-13
bap-objdump: fix "--syms" parsing bug from last commit, autogenerate dummy symbol for each Memory if necessary

ddcc:
bap-objdump: fix bug from last commit where passing '--binary' causes '--syms' to not be parsed
bap-objdump: generate a dummy symbol for each Memory if none are present

bap_fileutils: add missing eof newline


src/readbin/readbin.ml
```diff
@@ -8,10 +8,16 @@ open Image
...
-    let disasm = match roots with
-      | Some x -> disassemble ~roots:x arch mem
-      | None -> disassemble arch mem in
+  let disassemble ?img arch mem =
+    let syms = match options.symsfile with
+      | Some filename ->
+        Symtab.read ?demangle:options.demangle ~filename mem
+      | None -> match img with
+        | Some x -> Table.map (symbols x) ~f:Symbol.name
+        | None -> Table.singleton mem "bap_autogen" in
```
ivg: can you please rename "bap_autogen" to "text"
This will result in consistent behavior, since if we don't know to what symbol address is belong, we just use "text" segment as a big symbol.

ddcc: done

Issue #67, opened 2015-01-16
pluginize transports for bap-server


ivg:
Transports should be plugins, loadable at runtime.

PR #95, opened 2015-02-17
>= deps

percontation:
It'd be nice to be able to install bap without downgrading core_kernel, and the newer versions of libraries seem to work fine. So unless there's any reason to fix bap to the old versions of packages I propose using >= dep constraints.

Also added the `rm` for uninstalling binaries.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -70,12 +70,8 @@ Flag zmq
   Default: false
 
 Flag llvm
- Description: Build with llvm backend
- Default: true
-
-Flag server
- Description: Build BAP server
- Default: true
+  Description: Build with llvm backend
+  Default: true
 
 Library bap
   Path:            lib/bap
@@ -305,10 +301,13 @@ Library core_lwt
                   Core_lwt_pool,
                   Core_lwt_stream
 
-Library zmq
+Library zmq_transport
   Path:           src/server
   Build$:         flag(server) && flag(zmq)
   CompiledObject: best
+  FindlibParent:  bap
+  FindlibName:    zmq_transport
+  BuildDepends:   lwt-zmq
   Install:        true
   Modules:        Zmq_client, Zmq_server
 
--- a/opam
+++ b/opam
@@ -8,13 +8,14 @@ bug-reports: "https://github.com/BinaryAnalysisPlatform/bap/issues"
 dev-repo: "git://github.com/BinaryAnalysisPlatform/bap/"
 license: "MIT"
 build: [
-  ["./configure" "--prefix=%{prefix}%" "--enable-docs"]
+  ["./configure" "--prefix=%{prefix}%" "--enable-docs" "--%{lwt-zmq:enable}%-zmq"]
   [make]
 ]
 install: [make "install"]
 remove: [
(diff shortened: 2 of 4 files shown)


ivg:
I think that Travis can give a better answer than me))

Anyway, we're stick with this old version of core_kernel library because this is the latest library version that still works for version 4.01.0.  So the proper way would be to specify in `opam` file something like this `core_kernel  { = "111.28.0" & ocaml-version "4.01.0" | >= "111.28.0" & ocaml-version > "4.01.0"}` But, the problem is that we're still trying to support `opam 1.1` that doesn't support such queries, and that means that users using it can't install bap with either `4.01` or `4.02` (depending on what `opam-admin` will left from the expression in a process of simplifying the expression). So, the only solution, that I see is to remove dependency from the OCaml 4.01 at all, but not sure that this is a very good idea. 

What concerning using `>=` dependencies in general, I also do not think that it is right. A good constraints, should specify only version that are tested to work. Moreover `>= 1.0` specifies, that it should work even with `2.0` and `3.0`, but according to semantic versioning rules, changes in major number introduce breaking changes. So maybe `{ >= 1.0 & < 2.0}` would be a more correct recipe (if opam correcly compares semantic versions. But, we can't apply this tactics to versions less then `1.0` because, according to semantic versioning rules, every new release starting with `0.` can introduce breaking changes without further notice (and they do).

Yet another problem, you've touched `opam.deps` file, that actually is not used in `opam` bundling, and it is written mostly for compiling bap from sources (usefull for testing and developing). If you're doing this, than it can be easier for you to create your own `opam.deps` or just not to use it at all and write dependencies manually.

What concerning `rm` then `ocamlfind remove` will properly remove all dependencies, that he installed, because it maintains a propert manifest. The only issue, is that we're installing `baptop` and `bapbuild` utilities, that aren't properly removed. 

If you find embracing, that `opam` downgrades all your packages, and you personally are not interested in using old `opam` or OCaml versions, then I would suggest you, to clone `opam-repository` and add change `bap` opam file to your needs, and then use `opam add repository <url-to-your-repo>`, where `<url-to-your-repo>` can be a local path or your github repo. 

Thank for feedback!

percontation:
Hmm, I feel there's a significant hole in your logic when more than one person in the universe decides that they're special and want to fix a core_kernel version... but whatever, glad to see that the situation has been considered :)

Unrelatedly, you still might want to fix the small copy-paste error in the _oasis file, wherein the `server` flag is specified twice.

ivg:
> Hmm, I feel there's a significant hole in your logic when more than one person in the universe decides that they're special and want to fix a core_kernel version...

There is no hole here, if I understand you correctly. Everyone can clone `opam-repository` and fix it locally, and use it, it won't clash with any other's repositories. Moreover, you can even decide to switch it several times and have different copies on your laptop. 

> Unrelatedly, you still might want to fix the small copy-paste error in the _oasis file, wherein the server flag is specified twice.

Thanks, sharp eye! Will fix, and will also properly add a `zmq` transport. Issue #67 is about this.

PR #97, opened 2015-02-17
Nuke opam.deps

rvantonder:


ivg:
You've forgotten the most important part, our CI test. It uses this files (actually the were created for Travis). Look at `.travis-ci-sh`

.travis-ci.sh
```diff
@@ -47,7 +46,7 @@ opam --version
 opam --git-version
 opam repository list
 
-opam install ${OPAM_DEPENDS}
+opam install --deps-only bap
```
ivg: you should also pin bap to our repository, otherwise we will have a variant of an egg and a chicken problem. 

if we add new dependency we will need to add it to the `opam-repository` first, so we need to release, before we can pass the CI test.

rvantonder:
Catch-22 creates more problems than it solves, abandoning.

PR #98, opened 2015-02-17
add objdump verbose flag and fix empty symbol table, make Disasm.Error include Printable

ddcc:


lib/bap_disasm/bap_disasm.ml
```diff
@@ -210,6 +210,19 @@ module Disasm = struct
     | `Failed_to_lift of mem * insn * Error.t
   ] with sexp_of
 
+  module Error = Printable(struct
+    open Format
+    type t = error
+
+    let module_name = "Bap_disasm.Error"
```
ivg: I expect that this won't work, the correct path should be `Bap_disasm.Disasm.Error`
this should came up as an error when you open `Bap.Std` in `baptop`

ddcc: You're right, fixed.

PR #104, opened 2015-02-26
Preparation for x86 integration.

ivg:
This PR contains fixes, that will allow to pass CI test for x86 and
x86_64 lifter.

It includes the ability to retrieve signatures for x86, that was
uploaded to github (For some reason, github doesn't allow us to release
files with other extensions, so this PR renames `sigs.db` to
`sigs.zip`). Also, there're some small fixes, like hardcoded 32 bits
address widths in readbin.


ivg:
close since it is unmergable

PR #127, opened 2015-03-13
slight change to ADT repr

percontation:
Recurse into tuple args when printing. This looks better especially for empty tuples, which show up a lot as clauses in If statements.

For example, before this change: 
`If(EQ(AND(Int(0x8049a00, 0x20), Int(0xf, 0x20)), Int(0x0, 0x20)), "()", CpuExn(0xd))` 
and after: 
`If(EQ(AND(Int(0x8049a00, 0x20), Int(0xf, 0x20)), Int(0x0, 0x20)), (), CpuExn(0xd))`

### DIFF
--- a/python/adt.py
+++ b/python/adt.py
@@ -69,6 +69,8 @@ class ADT(object):
                 return '0x{0:x}'.format(x)
             elif isinstance(x, ADT):
                 return str(x)
+            elif isinstance(x, tuple):
+                return "(" + ", ".join(qstr(i) for i in x) + ")"
             else:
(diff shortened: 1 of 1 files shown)


ivg:
this representation is used not for pretty-printing, but for data serialization, have you checked that this didn't break interaction with qira?

percontation:
I don't see where this is used for data serialization. The serialization step is the string that ocaml emits, and this change actually fixes the python `__repr__` to match the strings that ocaml produces.

PR #129, opened 2015-03-13
FLAGS registers fixes

percontation:
Typo is giving FLAGS regs the wrong lengths. For example, when lifting the `\x9c` `pushf` instruction, the generated BIL shows a 32-bit store of a 33-bit value.


ivg:
Thanks!

PR #139, opened 2015-03-20
As per small commit messages edited bap-mc for disassembling on command line and revived piqi translation ability

KennethAdamMiller:
Also, this requires that you update your piqi to the latest version, built so that the interpreter can find the function it needs dynamically. A pull request was sent to the piqi repo maintainer in order that this be done, when it's finished do an opam upgrade piqi.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -45,7 +45,7 @@ Document bap
 
 Flag serialization
   Description: Build serialization library
-  Default: false
+  Default: true
 
 Flag llvm_static
   Description: Links with llvm in a static mode
@@ -106,7 +106,6 @@ Library top
             Bap_install_printers,
             Bap_load_plugins,
             Bap_toplevel
-
   BuildDepends: compiler-libs, bap, bap.plugins
 
 Library types
@@ -154,7 +153,7 @@ Library serialization
   BuildTools: ocamlbuild, piqi
   DataFiles:    *.piqi
   CompiledObject: best
-  BuildDepends: piqirun, bap.types
+  BuildDepends: piqilib,piqirun,piqirun.pb,piqirun.ext,bap.types
   Modules:   Bil_piqi,
              Stmt_piqi,
              Stmt_piqi_ext
@@ -375,7 +374,7 @@ Executable "bap-mc"
   MainIs:         bap_mc.ml
   Install:        true
   CompiledObject: best
-  BuildDepends:   bap, bap.plugins, cmdliner
+  BuildDepends:   bap, bap.plugins, cmdliner, bap.types.serialization
 
 Executable "bap-byteweight"
   Path:           src/byteweight
--- a/baptop
+++ b/baptop
@@ -1,3 +1,4 @@
 #!/bin/sh
-
-exec utop -require bap.top "$@"
+# for bil serialization do: -required "bap.types.serialization"
+# disabled by default to reduce breakage liklihood
+exec utop -require "bap.top" "$@"
--- a/opam
+++ b/opam
@@ -24,6 +24,7 @@ install: [
 remove: [
   ["ocamlfind" "remove" "bap"]
   ["ocamlfind" "remove" "core_lwt"]
+  ["ocamlfind" "remove" "piqi"]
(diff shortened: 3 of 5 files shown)


ivg:
Great work! 

But as far as I understand `baptop` (or `utop` underneath the hood) will work only on your system, because you have locally fixed piqi, to deliver shared libraries? In that case we need firstly to fix the piqi, installation procedure, or still leave baptop without piqi stuff.

KennethAdamMiller:
Baptop will break for a short time, until the piqi is fixed and a new
version tagged and shipped in opam. Either route is fine, we don't have to
merge immediately, just wait until the piqi guy gets his change in and
ships the new version.

KennethAdamMiller:
Also, thanks! :)

ivg:
you also need to add `piqi` to `opam.deps` file, that is used to install opam based dependencies in our CI system

KennethAdamMiller:
I can fix it! :)

ivg:
I'm ready to merge this, but there're left to more things to be done:

1. Can you rebase onto master squashing all to one commit
2. Temporary remove loading piqi into baptop

Rationale for the 1: 

We're trying to keep our history as tight as possible. This will allow us to bisect bugs faster. Also all commits must pass the CI.

Rationale for the 2: 

Currently piqi doesn't work with toplevel as it doesn't provide necessary library files, even after the upstream will be fixed the old version will be still on the horizon, thus breaking baptop here and there. Moreover, I want in future going to make piqi optional dependency, to reduce amount of dependencies. Those who would like to run `baptop` with piqi can always provide `-pkg bap.types.serializations` flag to `baptop` invocation.

KennethAdamMiller:
Absolutely, I will do as you request. In the baptop script, I will add a
comment explaining the package and flag to use to get such support.

KennethAdamMiller:
I have squashed the commits, completed a merge. I'm not sure if/how github picks up the merge request differences, but in any case, I think it's done as you wanted.

ivg:
squashed branch should contain only one commit, here are [instructions](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html). You done something very wrong, and I'm expecting that you canhave problems now with it. I would suggest you to reset your last action and start from scratch following the instructions.

ivg:
And yet one more thing, just recalled, don't forget to update `opam` file, with a new dependency.

KennethAdamMiller:
I'm pretty sure the reason why it looks like a stack of commits is because the squashed commit is in between some commits that were made since the original pull request was made. I don't know what happened to drop the opam edit, because previously, I had had the edit in the opam file. I don't know how it disappeared when I did the rebase. I'm sorry for the complication; after I had done the rebase, it rejected my push, saying that I had to pull. When I pulled, it created an automatic merge (what you see at the top). I will try and revise again and get the opam edit in, but I don't know that that will make the long line of commits go away. Sorry again.

ivg:
That's ok, git is the most complicated part in the binary analysis))  
What concerning `opam`, there're two places that should be updated: `opam.deps` that is used in `CI` system, and `opam` (without deps extension) is a file, that describes to `opam` package manager how to install `bap`. There is some sort of duplication due to the fact, that we still trying to support old version of opam.  What concerning squashing, you have two choices, either create a new PR, or push with `--force` into this one. 
P.S. if you find yourself completely lost with this rebasing, just say it to me, I can do it myself. It is easy, since I have emacs, and it will do everything for me with a few keystrokes.

KennethAdamMiller:
Hey I use Emacs too, man. I love magit! I've used git a lot in the past,
I'll see if I can't simplify the look of things by rearranging the order of
the commits. It's not that I'm completely lost, but that opam edit just
disappeared, like I don't know where it went.

dbrumley:
Git is a programming language unto itself. Your pain is felt by me too.

KennethAdamMiller:
*sigh* Ok I think I fixed it. 

I are baboon!

ivg:
Thanks a lot for the contribution!

PR #141, opened 2015-03-21
This PR implements iter, iteri, fold for Memmap.

dbrumley:
The implementation is modeled after map, mapi, and so on.


ivg:
I think that `fold` as well as `iter` aren't very useful. Usually one wants `foldi`, and `iteri`. As well as `find`, `find_map` and all other usual stuff. I think that better approach would be to write on function `to_sequence` (c.f., `Table.to_sequence`), that would grant access to all operations at one.  Another option, is to use `Container.Make` functor to create a set of iterators. Or, both approaches can be implemented.

PR #142, opened 2015-03-22
Fixed conficting name bug in ARM lifter

tim-becker:
Two variables were named `t`, which caused conflicts in some instructions. For example, the following:
```
	Instrction:  	subs	r4, r4, r12, lsr #1
	BIL:
	Move(Var("t", Imm(0x20)), Var("R12", Imm(0x20)))
	Move(Var("s", Imm(0x20)), Var("R4", Imm(0x20)))
	Move(Var("t", Imm(0x20)), RSHIFT(Var("t", Imm(0x20)), Int(0x1, 0x20)))
	Move(Var("R4", Imm(0x20)), MINUS(Var("R4", Imm(0x20)), RSHIFT(Var("t", Imm(0x20)), Int(0x1, 0x20))))
	Move(Var("CF", Imm(0x1)), LE(Var("t", Imm(0x20)), Var("s", Imm(0x20))))
	Move(Var("VF", Imm(0x1)), HIGH(0x1, AND(XOR(Var("s", Imm(0x20)), Var("t", Imm(0x20))), XOR(Var("s", Imm(0x20)), Var("R4", Imm(0x20))))))
	Move(Var("NF", Imm(0x1)), HIGH(0x1, Var("R4", Imm(0x20))))
	Move(Var("ZF", Imm(0x1)), EQ(Var("R4", Imm(0x20)), Int(0x0, 0x20)))

```

After the third statement, the unshifted value is incorrect, but is referenced in the computation for `R4` and some flags.

### DIFF
--- a/lib/bap_disasm/bap_disasm_arm_mov.ml
+++ b/lib/bap_disasm/bap_disasm_arm_mov.ml
@@ -19,7 +19,7 @@ let lift ?dest src1 ?src2 (itype ) ?sreg ?simm raw ~wflag cond =
     | Some src -> exp_of_op src
     | None     -> zero reg32_t in
 
-  let unshifted = Env.new_tmp "t" in
+  let unshifted = Env.new_tmp "u" in
 
   (* Do the register shift *)
   let s1, s2, stmts, carry =


ivg:
every temporary variable is unique, as it is postfixed with unique number.

tim-becker:
@ivg Is this supposed to be explicit, or am I to assume that every non-register variable has this property?

ivg:
every variable has this property, but non-temporary variables are printed without the unique number.

ivg:
you should take a look at the place, where bap-server emits the ADT representation for BIL, and fix the issue their. I.e.., to make it to emit temporary variables with full name, instead of just a human-readble part. If you don't know where to look or how to fix, just create an issue, and I will take care of this.

tim-becker:
I'm still not sure I understand. Both of these variables named `t` were created as temporary, and as you can see in my BIL statement, they do not have unique names.

dbrumley:
It looks like the line you changed is new_tmp "t" to new_tmp "u". What @ivg is saying is the bug isn't where you fixed it; it is elsewhere (I think).  In BAP, we give it a unique number, e.g., think `new_tmp t; new_tmp t` as producing two temporaries `t1` and `t2`. See `bap_disasm_arm_env.ml`, where we have `let new_tmp name = Var.create ~tmp:true name reg32_t`. This says to create a new variable. If you look in `bap_var.ml`, the function `create` assigns a fresh identifier to each variable.  I think what @ivg may be saying is the bug is that fresh identifier isn't communicated when bap-server emits the ADT.

tim-becker:
Ah, I understand now! I'll look into that.

PR #155, opened 2015-03-26
Elf/MachO/COFF binary files loader backend.

feseal:
We have made the BAP plugin for loading Elf, MachO and COFF binaries. We are not completely sure of the correctness of the extraction of the parameters for MachO and COFF. It would be great to hear Your comments.

### DIFF
--- a/lib/bap_image/llvm_binary_stubs.cpp
+++ b/lib/bap_image/llvm_binary_stubs.cpp
@@ -1,0 +1,169 @@
+#include <string>
+
+#include "llvm_binary_stubs.h"
+#include "llvm_binary_stubs.hpp"
+
+extern "C" {
+#include <caml/mlvalues.h>
+#include <caml/memory.h>
+#include <caml/fail.h>
+#include <caml/alloc.h>
+#include <caml/custom.h>
+#include <caml/bigarray.h>
+}
+
+void llvm_binary_fail [[ noreturn ]](const char* message) {
+    ::caml_failwith(message);
+}
+
+namespace impl {
+
+using namespace llvm;
+using namespace llvm::object;
+
+Binary* from_value (::value v) {
+    return *reinterpret_cast<Binary**>(Data_custom_val(v));
+}
+
+::value to_value (Binary* b) {
+    CAMLparam0();
+    CAMLlocal1(v);
+    static struct custom_operations binary_ops = {
+        const_cast<char*>("llvm.caml.llvm_binary"),
+        [](::value v) { CAMLparam1(v); delete from_value(v); CAMLreturn0;},
+        custom_compare_default,
+        custom_hash_default,
+        custom_serialize_default,
+        custom_deserialize_default
+    };
+    v = ::alloc_custom(&binary_ops, sizeof(Binary*), 0, 1);
+    *reinterpret_cast<Binary**>(Data_custom_val(v)) = b;
+    CAMLreturn(v);
+}
+
+
+::value to_value(Triple::ArchType arch) {
+    CAMLparam0();
+    CAMLlocal1(v);
+    typedef Triple::ArchType AT;
+    switch(arch) {
+    case AT::arm: v = ::caml_hash_variant("arm"); break;
+    case AT::aarch64: v = ::caml_hash_variant("aarch64"); break;
+    case AT::hexagon: v = ::caml_hash_variant("hexagon"); break;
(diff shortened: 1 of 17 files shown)


lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,172 @@
...
+    CAMLlocal1(result);
+    result = caml_alloc(7, 0);  
+    Store_field (result, 0, caml_copy_string(s.name().c_str()));
+    Store_field (result, 1, caml_copy_int64(s.offset()));
+    Store_field (result, 2, caml_copy_int64(s.addr()));
+    Store_field (result, 3, caml_copy_int64(s.size()));
+    Store_field (result, 4, Val_bool(s.is_readable()));
+    Store_field (result, 5, Val_bool(s.is_writable()));
+    Store_field (result, 6, Val_bool(s.is_executable()));
+    return result;
```
ivg: You should use `CAMLreturn` in allocating stubs

lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,172 @@
...
+    case AT::nvptx64: return ::caml_hash_variant("nvptx64");
+    default:
+        std::string message("Unknown arch: ");
+        message += std::to_string(static_cast<int>(arch));
+        ::caml_failwith(message.c_str());
+    }
+}
+
+
+::value to_value(const seg::segment& s) {
```
ivg: All allocating function must start with `CAMLparam*` macro:

   > The function body must start with one of the [CAMLparam] macros.
   If the function has no parameter of type [value], use [CAMLparam0].
   If the function has 1 to 5 [value] parameters, use the corresponding
   [CAMLparam] with the parameters as arguments.
   If the function has more than 5 [value] parameters, use [CAMLparam5]
   for the first 5 parameters, and one or more calls to the [CAMLxparam]
   macros for the others.
   If the function takes an array of [value]s as argument, use
   [CAMLparamN] to declare it (or [CAMLxparamN] if you already have a
   call to [CAMLparam] for some other arguments).

   > If you need local variables of type [value], declare them with one
   or more calls to the [CAMLlocal] macros at the beginning of the
   function, after the call to CAMLparam.  Use [CAMLlocalN] (at the
   beginning of the function) to declare an array of [value]s.

lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,172 @@
...
+    };
+    ::value v = ::alloc_custom(&binary_ops, sizeof(Binary*), 0, 1);
+    *reinterpret_cast<Binary**>(Data_custom_val(v)) = b;
+    return v;
+}
+
+
+::value to_value(Triple::ArchType arch) {
+    typedef Triple::ArchType AT;
+    switch(arch) {
```
ivg: It is better to use strings here. For example, what if some of the variants has clashing hashes?

lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+namespace impl {
+
+using namespace llvm;
+using namespace llvm::object;
+
+Binary* from_value (::value v) {
+    CAMLparam1(v);
+    return *reinterpret_cast<Binary**>(Data_custom_val(v));
```
ivg: There still should be `CAMLreturn`. `CAMLparam` must always match with `CAMLreturn`

ivg: But in this function you can just remove `CAMLparam`, since it allocates nothing.

dbrumley:
I didn't see any test cases. Can you please let us know what the status of testing is? I would like to see tested on macho, PE (when you say coff, you mean just coff or pe/coff?), and elf compiled with gcc and llvm if possible. We have various git repos set up with some testing at https://github.com/binaryanalysisplatform/. Thanks!

feseal:
We have simple test, which just read a file and print content of Img.t record. We tried to read macho ls, mv and paint.exe from windows. We will prepare set of required test after discussing it with @ivg.

ivg:
@tiffanyb, do you have a set of test files for mac and windows platforms? Is there is a chance, that this [script](https://github.com/BinaryAnalysisPlatform/x86_64-binaries/blob/master/build.py) will work on other platforms?

feseal:
Closed for testing

PR #163, opened 2015-04-01
Elf/MachO/COFF binary files loader backend with tests

feseal:
LLVM loader backend with tests.
We tested elf coreutils (arm, x86, x86_64) and macho coreutils (x86, x86_64)


lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,172 @@
...
+    CAMLreturn(v);
+}
+
+
+::value to_value(Triple::ArchType arch) {
+    CAMLparam0();
+    CAMLlocal1(v);
+    typedef Triple::ArchType AT;
+    switch(arch) {
+    case AT::arm: v = ::caml_copy_string("arm"); break;
```
ivg: There is a function to transfer arch to string

lib/bap_image/llvm_binary_stubs.cpp
```diff
@@ -0,0 +1,172 @@
...
+    case AT::aarch64: v = ::caml_copy_string("aarch64"); break;
```
ivg: And where are subarches?

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -0,0 +1,449 @@
...
+
+    std::vector<sec::section> sections() const {
+        return sec::read(obj_);
+    }
+protected:
+    const T *obj_;
+};
+
+template <typename T>
+std::shared_ptr<image_base> create_image(const ObjectFile* obj) {
```
ivg: it is not a create function, it is a cast function. Otherwise, calling `create_image` in each getter shocks the reader

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -0,0 +1,449 @@
...
+} //namespace sec
+
+namespace img {
+using namespace llvm;
+using namespace llvm::object;
+
+struct image_base {
+    virtual uint64_t entry() const = 0;
+    virtual Triple::ArchType arch() const = 0;
+    virtual std::vector<seg::segment> segments() const = 0;
```
ivg: is there any good reason to copy segments every time?

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -0,0 +1,449 @@
...
+    StringRef data_ref(data, size);
+    MemoryBuffer* buff(MemoryBuffer::getMemBufferCopy(data_ref, "binary"));
+    OwningPtr<object::Binary> binary;
+    if (error_code ec = createBinary(buff, binary))
+        llvm_binary_fail(ec);
+    return binary.take();
+}
+
+Triple::ArchType llvm_binary_arch(Binary* binary) {
+    return img::create_image(binary)->arch();
```
ivg: is it good to create image wrapper in each accessor function? Why don't you call this in `llvm_binary_create`?

ivg:
Please, remove the query to environment variables in the plugin registration. Plugins should be registered unconditionally. In CI `test.sh` Pass a value of `BAP_LOADER` variable to the `--loader` command-line option, and don't forget to set this variable to `bap-elf` on first five rows of the build matrix. And, please, rename your backend just to "llvm". 

Everything else will be addressed lately, now I want to have it to start testing IRL. 

Thanks in advance!

P.S. I've reverted mach-o from the test artifacts. Looks like that we need to redesign our test suites. It already pushes to hard on Travis.

PR #168, opened 2015-04-03
Updating to LLVM 3.5

perror:
Finally got it to compile with LLVM 3.4 and 3.5. I also tried with 3.6 but there a few more changes (nothing untracktable but a bit annoying because it keep segmenting the code with `#ifdef` macros).

Test it first before merging because I did not tried it in depth.

### DIFF
--- a/lib/bap_disasm/llvm_disasm.cpp
+++ b/lib/bap_disasm/llvm_disasm.cpp
@@ -1,4 +1,6 @@
-#include <llvm/ADT/OwningPtr.h>
+#if LLVM_VERSION_MAJOR == 3 && LLVM_VERSION_MINOR < 5
+  #include <llvm/ADT/OwningPtr.h>
+#endif
 #include <llvm/MC/MCAsmInfo.h>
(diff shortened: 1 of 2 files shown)


ivg:
You may noticed, that we have got recently new updated concerning LLVM, like bindings to LLVM Binary library. Have you tried to compile it with newer LLVM?

perror:
No, I didn't try it with the new added code. I'll try it if I can find a slot for that (but I'm quite busy right now).

perror:
By any chance, are you at ETAPS or LLVM European Conference in London this week ?

ivg:
unfortunately not

sjcappella:
Any update to this issue?

ivg:
nope, from that time we got more llvm 3.4 dependent code (bindings to BinaryObject), so this PR will not work

sjcappella:
Ok, thanks for the answer. I'll just keep depends on my machine pointing to llvm 3.4 instead of default. That should work out for me.

PR #186, opened 2015-04-21
Final refining before release.

ivg:
1. Refined `project` type

This PR is major refining of a project type, that is used in a plugin
system. First of all, some fields are deprecated, since we're going to
move away from the low-level representation to a more high-level
intermediate representation. Also, we're gradually moving away from the
incorrect abstractions of symbol (where the symbol must be contiguous in
space and every block belongs only to one symbol).

This PR also introduce a storage map, that can be used to exchange data
between plugins.

2. Renamed `bap-objdump` to `bap`

bap-objdump outgrew its name

3. Put Travis on a diet

Waiting for three hours is more than enough! Reduced the set of testing
files, but now we're testing on more platforms.

4. Removed examples folder

Since it is dead anyway.


dbrumley:
Recall one thing people ask for is access to the various sections, e.g., rodata to find constants such as strings.  Is this data easily accessible now?

ivg:
sure, even officially. But it is named rather strange, `region`. Cite from documentation:

```
    Depending on the analysis performed and input parameters, one can
    expect that memory may be annotated with the following tags:

    - [Image.region] -- for regions of memory, that had some
      particular name in the original binary. For example, in ELF
      sections have names, that will be used to annotate corresponding
      memory regions.

    - [Image.section] -- if the binary data was loaded from a binary
      format that contains sections (aka segments), then corresponding
      memory regions would be marked. Sections gives you access
```

PR #188, opened 2015-04-22
Fix building documentation.

ivg:
Releasing revealed a bug in documentation, that wasn't caught by CI,
since we're not compiling documentation on Travis. I've fixed this
bug in opam file, that is in the official repository, and bumped changes
back here. Also, I fixed the source of the error.

Since, that time we no longer build documentation as a part of
installation, since it takes too much time, very faulty (this is not a
first time), and worthless (the output is of very low quality). When
codoc will mature, we will return to this issue.

### DIFF
--- a/lib/bap/bap_project.mli
+++ b/lib/bap/bap_project.mli
@@ -34,7 +34,12 @@
 
     - [Image.section] -- if the binary data was loaded from a binary
       format that contains sections (aka segments), then corresponding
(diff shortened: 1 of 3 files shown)


computereasy:
Hello Ivan, may I ask on what platform do you usually use to develop bap?

ivg:
I personally use Ubuntu Trusty 14.04 or Mac OS X Mavericks with macports

PR #189, opened 2015-04-22
wait for IDA autoanalysis to complete before exporting symbols

ddcc:


ivg:
Wouldn't be to much from my side to ask you to fix `extract_script` function also?
 As it may hit us later...

ddcc:
Fixed

ivg:
Thank you!

PR #197, opened 2015-04-28
fixed PE-64 entry point address

gitoleg:
Now, it's available to get entry point address from 64-bit PE executables

### DIFF
--- a/lib/bap_image/llvm_binary_stubs.hpp
+++ b/lib/bap_image/llvm_binary_stubs.hpp
@@ -331,12 +331,40 @@ uint64_t image_entry(const MachOObjectFile* obj) {
 }
 
 uint64_t image_entry(const COFFObjectFile* obj) {
-    const pe32_header* hdr = 0;
-    if (error_code ec = obj -> getPE32Header(hdr))
-        llvm_binary_fail(ec);
-    if (!hdr)
-        llvm_binary_fail("PE header not found");
-    return hdr->AddressOfEntryPoint;
+    if (obj->getBytesInAddress() == 4) {
+        const pe32_header* hdr = 0;
+        if (error_code ec = obj -> getPE32Header(hdr))
+            llvm_binary_fail(ec);
+        if (!hdr)
+            llvm_binary_fail("PE header not found");
+        return hdr->AddressOfEntryPoint;
+    } else {        
+        // llvm version 3.4 doesn't support pe32plus_header,
+        // but in version 3.5 it does. So, later one will be 
+        // able to write obj->getPE32PlusHeader(hdr) for 64-bit files.
+        uint64_t cur_ptr = 0;
+        const char * buf = (obj->getData()).data();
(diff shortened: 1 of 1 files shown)


lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -331,12 +331,35 @@ uint64_t image_entry(const MachOObjectFile* obj) {
...
+        if (!hdr)
+            llvm_binary_fail("PE header not found");
+        return hdr->AddressOfEntryPoint;
+    } else {        
+        // llvm version 3.4 doesn't support pe32plus_header,
+        // but in version 3.5 it does. So, later one will be 
+        // able to write obj->getPE32PlusHeader(hdr) for 64-bit files.
+        uint64_t cur_ptr = 0;
+        StringRef data = obj->getData();
+        MemoryBuffer *buf = MemoryBuffer::getMemBuffer(data, "", false);
```
ivg: 1. memory leak
2. return value is not checked
3. why are you creating this buffer at all, if you only need a pointer to the start, that you already have as it is equal to `data`

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -331,12 +331,35 @@ uint64_t image_entry(const MachOObjectFile* obj) {
...
+        const uint8_t *start = reinterpret_cast<const uint8_t *>(buf->getBufferStart());
+        uint8_t b0 = start[0];
+        uint8_t b1 = start[1];
+        if (b0 == 0x4d && b1 == 0x5a) { // Check if this is a PE/COFF file.
+            cur_ptr += *reinterpret_cast<const uint16_t *>(start + 0x3c);
```
ivg: and what if the input file is less than `3e` in size? segfault?

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -331,12 +331,35 @@ uint64_t image_entry(const MachOObjectFile* obj) {
...
+            // 3c is offset for coff_file_header in executables
+            cur_ptr += 4; // Skip the PE magic bytes.
+            cur_ptr += sizeof(coff_file_header);
+            const pe32plus_header *hdr = 
+                reinterpret_cast<const pe32plus_header *>(start + cur_ptr);
```
ivg: yet another segfault

lib/bap_image/llvm_binary_stubs.hpp
```diff
@@ -331,12 +331,35 @@ uint64_t image_entry(const MachOObjectFile* obj) {
...
+            return hdr->AddressOfEntryPoint;
```
ivg: again segfault

gitoleg:
All checks for file size are performed in COFFObjectFile constructor, so I still haven't added such checks. But if it's necessary, I'll do it. And I've added one more check to be sure that we have got exactly PEplus header.

PR #198, opened 2015-04-28
fixed bug with sum of bitvectors with different width

gitoleg:
Fixed bug with sum of bitvectors with different width. This bug occured during work with 64-bit PE files

### DIFF
--- a/lib/bap_disasm/bap_disasm_x86_lifter.ml
+++ b/lib/bap_disasm/bap_disasm_x86_lifter.ml
@@ -969,7 +969,8 @@ module ToIR = struct
             let shift = Bil.(offset land int_exp Pervasives.(t' - 1) t') in
             reg, shift
           | Oaddr a ->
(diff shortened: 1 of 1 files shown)


lib/bap_disasm/bap_disasm_x86_lifter.ml
```diff
@@ -969,7 +969,8 @@ module ToIR = struct
             let shift = Bil.(offset land int_exp Pervasives.(t' - 1) t') in
             reg, shift
           | Oaddr a ->
-            let byte = load (size_of_typ reg8_t) Bil.(a + (offset lsr int_exp 3 t')) in
+            let offset = Bil.(Cast (LOW, width_of_mode mode, offset)) in
```
ivg: Why `LOW`? `LOW` casting is used for downcasting usually. It means take `n` lower bits. In your case `n` is greater than the word bitwidth. For upcasting use `SIGNED` and `UNSIGNED`. Morover, use functional constructors, as they do not clobber the code with punctuation marks.

PR #200, opened 2015-04-30
As per discussion, grammar names conflict with cpp vernacular keywords, and some options were needed in bap-mc

KennethAdamMiller:
For our specific use cases, we needed to be able to know the opcode length, stop disassembling after some number of instructions and lastly specify a memory offset in order that certain BIL semantics be generated correctly.

Additionally, we would like to be able to deserialize via any method supported, so I went about changing the BIL spec to rename only what had forced keyword conflicts, I then propagated those changes throughout related ml code, and it compiles

### DIFF
--- a/src/bap_mc/bap_mc.ml
+++ b/src/bap_mc/bap_mc.ml
@@ -1,4 +1,4 @@
-open Core_kernel.Std
+ open Core_kernel.Std
 open Or_error
 open Format
 open Bap.Std
@@ -15,10 +15,10 @@ exception Unknown_arch
 exception Can't_lift of Error.t
 exception Trailing_data of int
 
-let no_disassembly state boff =
+let no_disassembly state (start_addr, boff) =
   let mem = Dis.memory state in
-  let addr = Dis.addr state in
-  let stop = Addr.to_int addr |> ok_exn in
+  let addr = Addr.((Dis.addr state) - start_addr) in
+  let stop = (Addr.to_int addr |> ok_exn) in
   raise (Bad_insn (mem, boff, stop))
 
 let escape_0x =
@@ -51,10 +51,9 @@ let read_input input =
     | "0x" ->  to_binary ~escape:escape_0x input
     | x -> to_binary ~escape:prepend_slash_x input
 
-let create_memory arch s =
-  let width = Arch.addr_size arch |> Size.to_bits in
+let create_memory arch s addr =
   let endian = Arch.endian arch in
-  Memory.create endian Addr.(of_int ~width 0) @@
+  Memory.create endian addr @@
   Bigstring.of_string s |> function
   | Ok r -> r
   | Error e -> raise (Create_mem e)
@@ -63,6 +62,11 @@ let print_kinds insn =
   Dis.Insn.kinds insn |>
   List.map ~f:sexp_of_kind |>
   List.iter ~f:(printf "%a@." Sexp.pp)
+
+let print_insn_size should_print mem =
+  if should_print then
+    let len = Memory.length mem in
+    printf "%#x\n" len
 
 let print_insn insn_formats insn =
   let insn = Insn.of_basic insn in
@@ -103,31 +107,48 @@ let print_bil lift bil_formats mem insn =
   List.iter bil_formats ~f:(fun fmt ->
       printf "%s@." (string_of_bil fmt (bil insn)))
 
-let make_print lift insn_fmt bil_fmt show_kinds mem insn =
-  print_insn insn_fmt insn;
+let make_print lift insn_fmt bil_fmt show_kinds show_size mem insn =
+  print_insn_size show_size mem;
+  print_insn insn_fmt insn; 
   print_bil lift bil_fmt mem insn;
   if show_kinds then print_kinds insn
 
-let step print state mem insn _ =
-  print mem insn;
-  Dis.step state (Dis.addr state |> Addr.to_int |> ok_exn)
+let check max_insn counter = match max_insn with
+  | None -> true
+  | Some max_insn -> counter < max_insn
+  
+let step max_insn print state mem insn (addr, counter) =
+  if (check max_insn counter) then (
+    print mem insn;
+    Dis.step state (Dis.addr state, counter+1)
+  ) else Dis.stop state (Dis.addr state, counter)
 
-let disasm src arch show_insn show_bil show_kinds =
+let disasm src addr max_insn arch show_oplen show_insn show_bil show_kinds =
   let arch = match Arch.of_string arch with
     | None -> raise Unknown_arch
     | Some arch -> arch in
+  let extension = match arch with
+    | `x86 -> ":32"
+    | `x86_64 -> ":64" in
+  let addr = Addr.of_string (addr ^ extension) in
   let module Target = (val target_of_arch arch) in
-  let print = make_print Target.lift show_insn show_bil show_kinds in
+  let print = make_print Target.lift show_insn show_bil show_kinds show_oplen in
   let input = read_input src in
   Dis.create ~backend:"llvm" (Arch.to_string arch) >>= fun dis ->
-  let invalid state mem off = no_disassembly state off in
-  let pos =
-    Dis.run dis ~return:ident
-      ~stop_on:[`Valid] ~invalid ~hit:(step print) ~init:0
-      (create_memory arch input) in
-  if pos <> String.length input then
-    raise (Trailing_data (String.length input - pos));
-  return ()
+  let invalid state mem (r_addr, off) = no_disassembly state (r_addr, off) in
+  let pos, dis_insn_count  =
+    Dis.run dis ~return:(fun x -> x)
+      ~stop_on:[`Valid] ~invalid
+      ~hit:(step max_insn print) ~init:(addr, 0)
+      (create_memory arch input addr) in
+  let bytes_disassembled = Addr.(pos - addr) |> Addr.to_int |> ok_exn in
+  let len = String.length input in
+  match max_insn with
+  | None -> 
+     if bytes_disassembled <> len then
+       raise (Trailing_data (len - bytes_disassembled));
+     return ()
+  | _ -> return ()
 
 module Cmdline = struct
   open Cmdliner
@@ -139,6 +160,14 @@ module Cmdline = struct
   let show_kinds =
     let doc = "Output instruction kinds." in
     Arg.(value & flag & info ["show-kinds"] ~doc)
+
+  let show_insn_size =
+    let doc =
+      "Output recognized opcode length (including potential data)\
+       as in the number of bytes EIP is incremented upon having\
+       executed said opcode."
+    in
+    Arg.(value & flag & info ["show-size"] ~doc)
 
   let show_insn =
     let formats = [
@@ -170,6 +199,21 @@ module Cmdline = struct
     Arg.(value & opt_all ~vopt:`bil (enum formats) [] &
          info ["show-bil"] ~doc)
 
+  let addr =
+    let doc = "Specify an address of first byte, as though \
+	       the instructions occur at a certain address, \
+	       and accordingly interpreted. Be careful that \
+	       you appropriately use 0x prefix for hex and \
+	       leave it without for decimal." in
+    Arg.(value & opt  string "0x0" &  info ["addr"] ~doc)
+
+  let max_insns =
+    let doc = "Specify a number of instructions to disassemble.\
+	       Good for ensuring that only one instruction is ever\
+	       lifted or disassembled from a byte blob. Default is all" in
+    Arg.(value & opt (some int) None & info ["max-insns"] ~doc)
+
+       
   let src =
     let doc = "String to disassemble. If not specified read stdin" in
     Arg.(value & pos 0 (some string) None & info [] ~docv:"STRING" ~doc)
@@ -193,7 +237,7 @@ module Cmdline = struct
            bap-mc  --show-inst --show-asm");
       `S "SEE ALSO";
       `P "llvm-mc"] in
-    Term.(pure main $src $arch $show_insn $show_bil $show_kinds),
+    Term.(pure main $src $addr $max_insns $arch $show_insn_size $show_insn $show_bil $show_kinds),
     Term.info "bap-mc" ~doc ~man ~version:Config.pkg_version
 
   let parse main = Term.eval (cmd main) ~catch:false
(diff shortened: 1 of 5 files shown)


src/bap_mc/bap_mc.ml
```diff
@@ -15,10 +15,10 @@ exception Unknown_arch
 exception Can't_lift of Error.t
 exception Trailing_data of int
 
-let no_disassembly state boff =
+let no_disassembly state boff offset =
   let mem = Dis.memory state in
   let addr = Dis.addr state in
-  let stop = Addr.to_int addr |> ok_exn in
+  let stop = (Addr.to_int addr |> ok_exn) - offset in
```
ivg: the arithmetics should be done in `Addr` algebra, not in `Int`, otherwise all negatives would be incorrect. With this said, `offset` is a bad name - as it is not an offset in the data, it is starting address of the first byte.

src/bap_mc/bap_mc.ml
```diff
@@ -170,6 +201,22 @@ module Cmdline = struct
     Arg.(value & opt_all ~vopt:`bil (enum formats) [] &
          info ["show-bil"] ~doc)
 
+  let addr_offset =
+    let doc = sprintf
```
ivg: This is not an offset, it is an address of a fist byte. Better name would be just `--addr`

ivg: also, there is no need to use `sprintf` here

src/bap_mc/bap_mc.ml
```diff
@@ -170,6 +201,22 @@ module Cmdline = struct
...
+    Arg.(value & opt ~vopt:0 (int) 0 &
+	 info ["offset"] ~doc)    
+
+  let stop_at =
+    let doc = sprintf
+		"Specify a number of instructions to disassemble.\
+		 Good for ensuring that only one instruction is ever\
+		 lifted or disassembled from a byte blob. 0 connotes all" in
+    Arg.(value & opt ~vopt:0 (int) 0 &
+	   info ["ins_limit"] ~doc)
```
ivg: it is not clear whether it supposed to be minimum or maximum number of instructions? Also, command line arguments usually use hyphen `-` instead of underscores. So better name, would be `max-insns` or `min-insns`.

src/bap_mc/bap_mc.ml
```diff
@@ -170,6 +201,22 @@ module Cmdline = struct
...
+	 interpreted. Bases can be hex or decimal " in
+    Arg.(value & opt ~vopt:0 (int) 0 &
+	 info ["offset"] ~doc)    
+
+  let stop_at =
+    let doc = sprintf
+		"Specify a number of instructions to disassemble.\
+		 Good for ensuring that only one instruction is ever\
+		 lifted or disassembled from a byte blob. 0 connotes all" in
+    Arg.(value & opt ~vopt:0 (int) 0 &
```
ivg: Instead of using `0` as a sentinel for infinity it is better to use `None` to specify the absence of the parameter, and use it as a flag that we should dissassemble as many as possible.

src/bap_mc/bap_mc.ml
```diff
@@ -140,6 +158,19 @@ module Cmdline = struct
     let doc = "Output instruction kinds." in
     Arg.(value & flag & info ["show-kinds"] ~doc)
 
+  let show_oplen =
+    let formats = [
+	"hex", `hex;
+	"dec", `dec;
+      ] in
+    let doc = sprintf
+		"Output recognized opcode length, optional \
```
ivg: this is not opcode length, this is length of the overall instruction including operands. And, are you sure that we really need to different formats to output length? As it clobber the interface, without really buying to much functionality?

src/bap_mc/bap_mc.ml
```diff
@@ -64,6 +64,13 @@ let print_kinds insn =
   List.map ~f:sexp_of_kind |>
   List.iter ~f:(printf "%a@." Sexp.pp)
 
+let print_insn_oplen oplen_bases mem =
+  let len = (Memory.length mem) in
+  (List.iter oplen_bases
+	     ~f:(function
+		  | `hex -> printf (format_of_string "%x\n") len
```
ivg: there is no need to call `format_of_string` on string literals when using `printf`. Constant string literals are already of type format when needed (they are overloaded). Moreover, it is better to use `%#x` specifier, since without it numbers will be printed without leading `0x`, that is very confusing. So, the better version would be:
```ocaml
printf "%#x" len
```
Also, all parenthesis in this function are excessive.

src/bap_mc/bap_mc.ml
```diff
@@ -15,10 +15,10 @@ exception Unknown_arch
 exception Can't_lift of Error.t
 exception Trailing_data of int
 
-let no_disassembly state boff =
+let no_disassembly state boff offset =
```
ivg: All arithmetics on addresses should be done in `Addr` algebra, not in `Int`, otherwise all negatives would be incorrect. Also many addresses won't fit into OCaml int, making lots of confusion. And yes, offset is a bad name - as it is not an offset in the data, it is starting address of the first byte.

src/bap_mc/bap_mc.ml
ivg: so, better name would be just `show-size`

src/bap_mc/bap_mc.ml
```diff
@@ -103,30 +110,41 @@ let print_bil lift bil_formats mem insn =
...
     | None -> raise Unknown_arch
     | Some arch -> arch in
   let module Target = (val target_of_arch arch) in
-  let print = make_print Target.lift show_insn show_bil show_kinds in
+  let print = make_print Target.lift show_insn show_bil show_kinds show_oplen in
   let input = read_input src in
   Dis.create ~backend:"llvm" (Arch.to_string arch) >>= fun dis ->
-  let invalid state mem off = no_disassembly state off in
+  let invalid state mem off = no_disassembly state off offset in
+  let hit_count = ref 0 in
```
ivg: There is no need to use a reference here. The last parameter of this function is specially reserved for user data. Currently it is the address, where disassembler machine is pointing, that is used to check, that we consumed all data. You can extend the state to tuple, of address and counter, and increment it in each step.

src/bap_mc/bap_mc.ml
```diff
@@ -103,30 +110,41 @@ let print_bil lift bil_formats mem insn =
...
+  print_insn insn_fmt insn; 
   print_bil lift bil_fmt mem insn;
   if show_kinds then print_kinds insn
 
-let step print state mem insn _ =
-  print mem insn;
+
+let check disasm_len current_total =
+  if disasm_len=0 then true
+  else current_total < disasm_len
```
ivg: this is usually written as `disasm_len = 0 || current_total < disasm_len`, but of course using `0` as a sentinel is a bad style

src/bap_mc/bap_mc.ml
```diff
@@ -103,30 +110,41 @@ let print_bil lift bil_formats mem insn =
...
+
+let step disasm_len hit_count print state mem insn _ =
+  if (check disasm_len !hit_count) then (
+    hit_count := !hit_count+1;
+    print mem insn;
+  ) else ();
```
ivg: First of all there is no need to provide `else` part if `then` part evaluates to zero. Second, you have encoded quite misleading, behavior. You promised a user to disassemble only `N` instructions, but actually, you disassemble all the instructions, but print only first `N`. That is not nice, what if they gave you a string of length gigabyte? Or worse, what if after the first instruction, there is no valid instructions at all, but just data, as it is quite usual for `jmp` instructions in `x86`?
In order to stop disassembler, you need to call `Disasm.stop` instead of step. And yes, this means, that you need also update the logic of exhaustiveness check.

src/bap_mc/bap_mc.ml
KennethAdamMiller: I thought that this was a good side step to the compiler errors that I mention in other comment. So this issue arose because of other issues. It's clearly not right, I will fix it.

src/bap_mc/bap_mc.ml
KennethAdamMiller: I will make the change.

At times, I have to use parenthesis to make sure that it passes the compiler, but then don't know how to realize the minimum number in OCaml.

src/bap_mc/bap_mc.ml
KennethAdamMiller: I agree about the bad sentinel. I will change it.

src/bap_mc/bap_mc.ml
KennethAdamMiller: The compiler didn't admit it if just the else (); part was taken off. 

Thanks, I was searching for a way to halt the disassembler, but then thought this was a workaround, because in our use case we only pass 32 bytes at most. The thinking was "whatever is not printed is not decoded". But I agree with you, that is correct and the best way to go about it.

Thank you for the criticism.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Yes, now that I know about Disasm.stop and the other alternative ways to solve the other problems, I can recode it and many of these alternative routes fall by the wayside. A few problems was causing other problems.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Ok, yes show-size is more precise. Yeah we don't have to add hex or decimal, I will just remove those. Int of string will deserialize each case correctly anyway, so it is irrelevant. I will fix it.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Will do.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Will do.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Yeah, I was thinking about how to name that. max-insns it is.

KennethAdamMiller:
All changes adopted, new bap-mc options tested.

src/bap_mc/bap_mc.ml
```diff
@@ -140,6 +161,14 @@ module Cmdline = struct
     let doc = "Output instruction kinds." in
     Arg.(value & flag & info ["show-kinds"] ~doc)
 
+  let show_insn_size =
+    let doc =
+      "Output recognized opcode length (including potential data)\
+       as in the number of bytes EIP is incremented upon having\
```
ivg: we shouldn't reference `EIP`, as it is x86-specific, just say "program counter"

src/bap_mc/bap_mc.ml
```diff
@@ -170,6 +199,20 @@ module Cmdline = struct
     Arg.(value & opt_all ~vopt:`bil (enum formats) [] &
          info ["show-bil"] ~doc)
 
+  let addr =
+    let doc = "Specify an address of first byte, as though \
+	       the instructions occur at a certain address, \
+	       and accordingly interpreted. " in
+    Arg.(value & opt ~vopt:"0" (string) "0" &
```
ivg: Looks like that you've misunderstood the role of this `vopt` parameter. If `vopt` is used in an argument specification, that means, that user can omit the value of the option, i.e., the grammar would be:
```EBNF
addr = "--addr", [ value].
```
Example, the following invocation would be valid,
```
$ bap-mc --addr 
```
And it is not only non-intuitive, but also introduces the ambiguity in the grammar, as now we can distinguish between unspecified address valued and positional parameter, e.g., 
```shell
$ bap-mc --addr DEADBEEFBADC0DE
```
here the argument would be treated as an address, not as data, as user expected. 

Another issue, is that your default value is incorrect, "0:32" is not valid string representation of the bitvector, as per grammar the base part is required. 

With all this said, the correct specification is:

```ocaml
Arg.(value & opt  string "0x0" &  info ["addr"] ~doc)
```

src/bap_mc/bap_mc.ml
```diff
@@ -170,6 +199,20 @@ module Cmdline = struct
...
+	 info ["addr"] ~doc)    
+
+  let max_insns =
+    let doc = "Specify a number of instructions to disassemble.\
+	       Good for ensuring that only one instruction is ever\
+	       lifted or disassembled from a byte blob. Default is all" in
+    Arg.(value & opt ~vopt:(Some 0) (some int) (Some 0) &
```
ivg: The same abuse of `vopt` parameter. And using `Some 0` as a special value for infinity is actually the same as using just `0`. The idea was to use `None` value for infinity, and `Some n` for some finite `n`. Here is a better way to do things:
```ocaml
Arg.(value & opt  (some int) None & info ["max-insns"] ~doc)
```
Of course you need to update the code, that handle this parameter.

src/bap_mc/bap_mc.ml
KennethAdamMiller: Will update.

src/bap_mc/bap_mc.ml
KennethAdamMiller: When I used baptop to test out Addr.of_string on some samples, it interprets a lack of 0x prefix as being decimal, and with it as being hex. So in each case it will function, but for clarity, I've also edited the doc for this option so that users won't have to look up that spec. :)

Dang, I've missed the spec in each case. I never saw that 0x prefix was a requirement, or I would have added it. The doc for [Addr.of_string](http://binaryanalysisplatform.github.io/bap/api/v0.9.6/experimental/codoc/index.html#/module:Std/type:addr) doesn't mention those...

Anyways, advice taken, appreciated and update applied.

src/bap_mc/bap_mc.ml
KennethAdamMiller: In my case, if I omit a value to max-insns option or use --max-insns 0, there are no instructions disassembled because check evaluates to false in that case. Also, check does regard None to be a moniker for infinity, because it always returns true.

I don't dispute the argument defaults are error prone and that this is more correct, I've adopted the advice in argument handling in each case :) Thanks!

ivg: No, you shouldn't ask users to input `--max-insn=0` for the infinity. The interface is as follows:
1. if `--max-insns` option is not specfied that means that there is no upper bound
2. otherwise, use the provided value as an upper-bound.

E.g.,
```
$ echo "DEADBEEFC0DE" | bap-mc 
```
will decode all instructions, but, 
```
$ echo "DEADBEEFC0DE" | bap-mc --max-insns=1
```
will stop after the first one.

PR #213, opened 2015-05-13
Make substitution available to plugins. Includes a few tests

rvantonder:
Make the handy `Project.substitute` function available to plugins, which is convenient for outputting BIL (and many other properties) of tagged memory.

Do check the `lib_test` structure: `lib_test/bap` contains the main test-runner `run_tests.ml`, so I created a separate directory `lib_test/bap_project`.


lib/bap/bap_project.ml
```diff
@@ -2,6 +2,10 @@ open Core_kernel.Std
 open Bap_types.Std
 open Image_internal_std
 open Bap_disasm
+open Format
```
ivg: We now have `Bap_image_std` and `Bap_disasm_std`, that should be opened instead of concrete modules from a library. In other words, just open this two modules and remove line 3,4,7,8

lib_test/bap_project/test_project.ml
```diff
@@ -0,0 +1,92 @@
...
+let test_substitution ctxt ~arch ~kind ~expect ~hexstring =
+  let project = create_project arch in
+  let result = substitute project ~kind ~hexstring in
+  let msg = sprintf "Incorrect %s substitution" kind in
+  assert_equal ~ctxt ~printer:str_printer ~msg expect
+    (match List.hd result with
+     | Some x -> x
+     | None -> assert_failure "Empty result")
+
+let expect_x86_mov_mem_esp_esi =
```
ivg: this tests are very fragile... we change printers quite often. What about regular expressions?

ivg: or, instead of regular expressions, you can normalize string by removing all whitespaces (including newlines) and compare normalized string. e.g., 
```
{
    mem32 := mem32\
            with [low:32[ESP], el]:u32 <- low:32[ESI]\
}
```
would be transformed to:

```
{mem32:=mem32with[low:32[ESP],el]:u32<-low:32[ESI]}
```

rvantonder: Sounds good. Another problem with testing output would happen if temp variables show up in the disassembly, so that's avoided.

ivg:
great work!

PR #214, opened 2015-05-13
Introduce Intermediate Representation.

ivg:
This PR adds long-awaited BAP IR. It is still work-in-progress (IR is
itself ready, but it is still far from being released). So I'm not going
to advertise the change right now. But everyone is welcome to review,
anyway.

Also, this PR moved stuff from `Tag` module to `Value`, and adds
`Value.Map` data structure.


lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+      type t
+      val create : ?tid:tid  -> ?subs:int -> unit -> t
+      val add_sub : t -> sub term -> unit
+      val result : t -> program term
+    end
+
+    include Regular with type t := t
+  end
+
+  (** Function.  *)
```
dbrumley: maybe call it subroutine instead of Sub, which is a commonly overloaded name?  Perhaps add a comment that of course we cannot use the type name Function.

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+  end
+
+  (** Block.  *)
+  module Blk : sig
+    (** Logically block consists of a set of {{!Phi}phi nodes}, a
+        sequence of {{!Def}definitions} and a sequence of out-coming
+        edges, aka {{!Jmp}jumps}. A colloquial term for this three
+        entities is a "block element".
+
+        The order of $\Phi$-nodes is unspecified. Definitions are stored
```
dbrumley: Be explicit on why the order is unspecified, e.g., The order of $\Phi$-nodes is left unspecified because the order does not affect the execution semantics.

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+          If [tid] parameter is specified, then the new block will have
+          this tid. If any of [phis], [defs] or [jmps] parameters are
+          specified, the provtided number would be used as a hint of the
+          expected amount of the corresponding entries. Since it is the
+          hint, it can mismatch with the actual size. The hint must be
+          a positive number.
+      *)
+      val create : ?tid:tid -> ?phis:int -> ?defs:int -> ?jmps:int -> unit -> t
+      val add_def : t -> def term -> unit
+      val add_jmp : t -> jmp term -> unit
```
dbrumley: Is a jump added to the beginning or end of the list?

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+    type t = arg term
+
+    val create : ?name:string -> typ -> t
+    val of_var : var -> t
+    val to_var : t -> var
+    val name : t -> string
+    include Regular with type t := t
+  end
+
+  module Call : sig
```
dbrumley: Are we above the ABI and calling convention here?  Does it make sense to put this information in a call, e.g., because the calling convention may be mixed in the same library (at least that is what I've heard, I can't claim to have observed it).

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+      val create : ?tid:tid  -> ?subs:int -> unit -> t
+      val add_sub : t -> sub term -> unit
+      val result : t -> program term
+    end
+
+    include Regular with type t := t
+  end
+
+  (** Function.  *)
+  module Sub : sig
```
dbrumley: I have the same question here as I do about return, would it be helpful to include abi information explicitly.

lib/bap/bap.mli
ivg: all builders functions append

ivg: c.f., OCaml Buffer, Java's and C# StringBuilder

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+
+    val create : ?name:string -> typ -> t
+    val of_var : var -> t
+    val to_var : t -> var
+    val name : t -> string
+    include Regular with type t := t
+  end
+
+  module Call : sig
+    type t = call
```
dbrumley: Shall we include ABI information here explicitly? Also, do we distinguish between an inter-procedural jump statement and call statement (they have two different stacks).

lib/bap/bap.mli
```diff
@@ -3822,6 +3860,377 @@ module Std : sig
...
+      val add_blk : t -> blk term -> unit
+      val add_arg : t -> arg term -> unit
+      val result : t -> sub term
+    end
+
+    include Regular with type t := t
+  end
+
+  (** Block.  *)
+  module Blk : sig
```
dbrumley: I'm probably missing it, but wouldn't `pred` and `succ` be really handy with blocks?

ivg: you're missing the `Term` interface. All terms have type `'a term`, and thus all functions from `Term` module are applicable.

lib/bap/bap.mli
ivg: ABI is used to create IR. This is, kind of, the end product.

lib/bap/bap.mli
ivg: I never heard anything about inter-procedural jumps with their own stack, google also didn't help me. 
We have `goto` for local jumps, and `call` for leaving the procedure. Without explicit return edge, it will not be possible to write intra-procedural analysis.

lib/bap/bap.mli
dbrumley: doh!

lib/bap/bap.mli
dbrumley: I didn't mean it's own stack. I meant that a call instruction pushes a return address (on x86), and a inter-procedural jump does not.  One has a return address to overwrite, one does not, which seems like a fact an analysis may want to consider.

lib/bap/bap.mli
dbrumley: Hmm. I was thinking a bit differently of the abstraction where it wouldn't hide the lower level details, just simplify them.  ABI seems like it would be relevant to many analysis, e.g., an analysis on the IR may want to update how many arguments there really are based on its results, which in turn may be improved by considering the abi.  I'm not sure it's useful to hide this detail in the abstraction, rather than provide a summary.  We can wait and see, but that's my bet for now.

lib/bap/bap.mli
ivg: are you talking about `setjmp`, `longjmp` and exceptions, when you're referring to "inter-procedural" jump?

PR #216, opened 2015-05-19
Test image is fixed for a 32-bit OS.

nikita-git:
Fix test_image for a 32-bit OS in i686.


ivg:
Thanks for fixing this!

PR #217, opened 2015-05-20
fix for PUNPCKHDQ, e.g. "660f6ac9"

percontation:


ivg:
hmm, can you elaborate, what did you fix?

percontation:
There's a typo in the source that prevents lifting of the PUNPCKHDQ instruction. For example, lifting the bytes "660f6ac9" causes an disasm_x86_utils.Arch_exception("impossible") to be raised.

ivg:
oh, thanks! What a weird name for opcode...

PR #261, opened 2015-07-07
Fix #260 and other compatibilty issues

### LINKED ISSUE #260: BAP not compatible with core_kernel.112.35.00
Hey,

I think the latest core_kernel update broke the bap build. More specifically, they "renamed `Monad.ignore` as `Monad.ignore_m`" in 112.35.00.

My build breaks at https://github.com/BinaryAnalysisPlatform/bap/blob/master/lwt/core_lwt_container.ml#L88, as Pervasives.ignore is now called instead of Monad.ignore.

To fix the build, consider renaming the uses of Mondad.ignore, or pinning core_kernel to 112.24.00.

Thanks,
Alex

ivg:
112.35.00 version of core introduces some breaking changes,
like changing a monad interface, and removing some dependencies.

This PR resolves this issue in a such manner, that it should work
for both - old and newer versions of core.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -304,7 +304,7 @@ Library types_test
   Build$:         flag(tests)
   Install:        false
   CompiledObject: best
-  BuildDepends:   bap, bap.conceval
+  BuildDepends:   bap, bap.conceval, oUnit
   Modules:        Test_bitvector,
                   Test_conceval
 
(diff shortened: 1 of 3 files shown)


maverickwoo:
Would this also mean we can lift the upperbound in the opam file?

ivg:
@maverickwoo, yep, we can.

PR #294, opened 2015-08-25
Add `r128 and `r256 sizes

percontation:
See #126 and #130; this restores support for lifting a variety of memory-related SSE instructions (e.g. `f3 0f 7f 04 38 movdqu %xmm0,(%eax,%edi,1)`), while breaking less stuff than #126.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -899,6 +899,8 @@ module Std : sig
       | `r16
       | `r32
       | `r64
+      | `r128
+      | `r256
     ] with variants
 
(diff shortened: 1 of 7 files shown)


percontation:
I can only speak for the x86-lifting use case, so someone should make sure that all fallout of #126 is indeed resolved before considering a merge.

ivg:
Looks ok, lets put into master and see the result. I've spend most of the time verifying correctness of the readers, they looks ok. Also, as far as I understand this is only a first step. The next one would be to fix the lifter itself.

PR #295, opened 2015-08-27
fix div instruction

percontation:
Small change earlier was accidentally creating Move(1-bit flag, 32-bit Unknown) BIL for `div` instructions (also maybe `bsf`).


ivg:
Thanks!

PR #297, opened 2015-09-14
Small client/server fixes. 

rvantonder:
Server should still respond when client does not supply id. E.g. `curl --data "{\"init\": {\"version\": \"0.1\"}}" 127.0.0.1:8080` does not get a response.

Client should break on successful connection.

### DIFF
--- a/src/server/server.ml
+++ b/src/server/server.ml
@@ -250,8 +250,10 @@ let run_exn (requests, replies) : unit Lwt.t =
           Response.create id msg |> reply
         | Error err' ->
           let err = Error.of_list [err; err'] in
-          let msg = Error.to_string_hum err in
-          warning_f ~section "Ignoring junk request: %s" msg)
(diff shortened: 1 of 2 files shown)


src/server/server.ml
```diff
@@ -250,8 +250,10 @@ let run_exn (requests, replies) : unit Lwt.t =
           Response.create id msg |> reply
         | Error err' ->
           let err = Error.of_list [err; err'] in
-          let msg = Error.to_string_hum err in
-          warning_f ~section "Ignoring junk request: %s" msg)
+          let str = Error.to_string_hum err in
+          warning_f ~section "Ignoring junk request: %s" str |> ignore;
```
ivg: a small issue here: you shouldn't ignore thread value, either use `ign_warning_f` function or bind the return value of `warning_f` with `reply` thread.

PR #299, opened 2015-09-24
A proposal to change BIL.

ivg:
This PR proposes to disallow nested load and store operations in BIL.
Essentially, this is the core part of a change:

```diff
- | Load    of exp * exp * endian * size (** load from memory *)
- | Store   of exp * exp * exp * endian * size (** store to memory  *)
+ | Load    of var * exp * endian * size (** load from memory *)
+ | Store   of var * exp * exp * endian * size (** store to memory  *)
```

Motivation
==========

Before showing a motivating example, let me refresh your memory with
syntax and operating semantics of BIL. (I will drop sizes and endianess
from discussion for simplicity, they really do not affect the described
problem). Whenever I refer to an operational semantics, or typing rules,
I'm actually referring to a BAP handbook, from old BAP.

We denote a load from address `a` of memory object `m` with the abstract
entity,

   `Load(m,a)`

or with the following syntax:

   `m[a]`

We denote a storage of value `v` to address `a` of memory object `m` as,

   `Store(m,a,v)`

or with the following syntax:

   `m[a <- v]`.

A memory object `m` can be any expression. In a well-formed program,
this expression should be of type `mem_t`. So let us narrow our
discussion only to well-typed programs, to exclude expressions like
`(1+2)[3]`.

There're three kinds of expressions that may result in a well formed
memory object:

  - Var v, that will be looked up in context D
  - Store(m,a,v), that extends memory object m with binding `a <- v`
  - Let (v,x,m), that substitutes `v` for `x` in the expression `m`,
    that may result in a valid memory object.

Again, for simplicity, let us reason only about `Var` and `Store`.

Since memory object is an expression we can encode nested (or parallel)
store operations, like this:

   `Store (Store (m,a1,x1),a2,x2)`

or in our concrete syntax:

   `m[a1 <- x1][a2 <- x2]`

Since the above expression is a proper memory object, it can be used
in the load instruction:

   `m[a1 <- x1][a2 <- x2][a1]`

that should, according to the operational semantics, evaluate to `x1`.

But consider the following expression:

   `m[a1 <- x1][a2 <- m[a1 <- x2][a1]][a1]`

This expression is also well formed and it also must evaluate to
`x2`, according to our operational semantics.

The problem arises, when we try to model semantics of a BIL with some
simple model of computation, like register machine with RAM.

Expressions in BIL don't have side effects. When we store a side
effect, visible to a machine, occurs when we assign a new memory object
to an old one. (Usually we limit ourselves to machines with only one
memory). For example,

  `m := m[a1 <- x1][a2 <- x2]`

will result in an ordered sequence of two side effects:

   1. store x1 at a1
   2. store x2 at a2

For load expressions there is no place for side effects at all.
(This is another problem, as memory access is also a side effect
that affects system, but this is another story).

Now let's try to figure out what side-effects will have the following
statement:

   `m := m[a1 <- x1][a2 <- m[a1 <- x2][a1]]`

Maybe:

   1. store x1 at a1
   2. store x2 at a1
   3. store x2 at a1

This will lead to an incorrect memory, as we have `x2` at `a1` instead
of `x1`. We can enumerate other possibilities, and it looks like that we
will end up with the only choice, that is valid - ignore nested write,
as it wasn't "commited" to a memory with an assignment statement.

The problem, actually, arose from a fact that `m[a1 <- x2]` from
mathematical perspective, creates a fresh new clone of memory `m` and
extends it with binding `a1 <- x2`. So that, loads

   `(m[a1 <- x1][a2 <- m[a1 <- x2][a1]])[a1]`

and

   `m[a1 <- x2][a1]`

are created from two different memory objects that have diverged at some
point of their history.

We can highlight the problem by desugaring this parallel match into a
series of assignments, while preserving SSA form (just for clarity).

`m := m[a1 <- x1][a2 <- m[a1 <- x2][a1]]` is transformed to:

```
  1: m1 = m[a1 <- x1]
  2: m2 = m[a1 <- x2]
  3: m3 = m1[a2 <- m2[a1]]
```

Here we clearly see that at program point (3) we simultaneously have two
different versions of memory, m1 and m2, that are clones of original
memory `m` with the same cell holding different values.

Unfortunately, we cannot model this memory cloning, not because it is
hard to implement (it is not), but because it moves us significantly far
away from real machines, that we would like to model. They are
deterministic and can't clone itself. There is only one memory (or
finite amount of designated memories), and each load and store
operations corresponds to a particular physical effect on a particular
memory address of real memory chip.

So, the real problem here is that we trying to model load and stores
with side effectless expressions. Although this is mathematically nice,
this indeed moves us away from real machines. But I'm not proposing to
raise them into statements, this is a too big change.

Instead, I'm claiming that if we limit the expressiveness of load and
store expressions we can get clean side effects that model memory
effects close to real machines, while still preserving flexibility of
memory objects as expressions [1].

By limiting `mem` in `Load(mem,addr)` to be a variable, we attach a side
effect to occur simultaneously with a context lookup.  Whenever we look
a `mem` variable we atomically lookup for the `addr` offset of this
memory. We're also protected from any side-effects, occuring in
evaluating of `addr` expression, since the only way to modify memory is
to perform explicit assignment.

By limiting `mem` in `Store(mem,addr,x)` expression, we allow to update
memory only explicitly via `Move` operation, and with one storage
operation per Move statement. This provides us with an excellent place
for a side-effect (at the end of the statement).

With a new scheme, we cannot encode our motivating example.  And  that's
right, since it is impossible to implement it on a real machine (as it
is impossible that the same address has two different values at the same
time).

[1]: Unfortunately, limiting load and stores will not solve all our problems,
as we still have `let` expression, where memory object produced by a
store expression can occur without requiring explicit side-effect, e.g.,

      `let m = m[a <- x1] in m[a] + 1`

That's why I still think that we need to get rid of let expressions. To
continue let's just assume, that already got rid of them.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -167,7 +167,7 @@ module Std : sig
       of the same type we usually disambiguate them with keywords,
       e.g., [Bil.Load of (exp,exp,endian,size)] has function
       {{!Bil.load}Bil.load} with type:
-      [mem:exp -> addr:exp -> endian -> size -> exp]
+      [mem:var -> addr:exp -> endian -> size -> exp]
 
       {3:value Value}
 
@@ -1511,8 +1511,8 @@ module Std : sig
 
       (** BIL expression variants  *)
       type exp =
-        | Load    of exp * exp * endian * size (** load from memory *)
-        | Store   of exp * exp * exp * endian * size (** store to memory  *)
+        | Load    of var * exp * endian * size (** load from memory *)
+        | Store   of var * exp * exp * endian * size (** store to memory  *)
         | BinOp   of binop * exp * exp  (** binary operation  *)
         | UnOp    of unop * exp         (** unary operation *)
         | Var     of var                (** variable *)
@@ -1626,8 +1626,8 @@ module Std : sig
     val sle : binop
     val neg : unop
     val not : unop
-    val load : mem:exp -> addr:exp -> endian -> size -> exp
-    val store : mem:exp -> addr:exp -> exp -> endian -> size -> exp
+    val load : mem:var -> addr:exp -> endian -> size -> exp
+    val store : mem:var -> addr:exp -> exp -> endian -> size -> exp
     val binop : binop -> exp -> exp -> exp
     val unop : unop -> exp -> exp
     val var : var -> exp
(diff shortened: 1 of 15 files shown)


ivg:
After a long discussion inside the team we decided to put this PR on hold.

But the problem is still there. BIL semantics and syntax allows us to formulate programs, that can't be compiled directly and doesn't model real physical machines.

# A summary of a discussion
 
Trying to fix the problem in a proposed way is a sort of jury rigging. The actual problem is that `load` and `store` are expressions and shouldn't have side-effects, but when we compile them to real machine (or model real machine) we would like to attach side-effects to loads and stores. To solve the problem, we can lift `load` and `store` operation into `stmt` type. But this will have several drawbacks: 
- too much impact on existing code;
- too much effort and risks;
- incompatibility with SMTP lib expressions;

PR #308, opened 2015-10-23
Bap trace library

feseal:


lib/bap/bap.mli
```diff
@@ -8007,4 +8007,435 @@ module Std : sig
...
+          [filter] that is applied to a sequence of events or errors, and
+          returns a sequence of events.  *)
+      val create : (event Or_error.t seq -> event seq) -> t
+    end
+
+    module Move : sig
+      type 'a t = {
+        cell : 'a;
+        data : word;
+      } with bin_io, compare, sexp
```
ivg: can you, guys, please add `fields` to the set of generators for all event types

lib/bap_trace/README.md
```diff
@@ -0,0 +1,31 @@
...
+not to load the whole trace at once.
+
+Events, as well as meta attributes are represented with universal
+values, so that trace can essentially contain any information. Since
+BAP's universal values are serializable, we naturally get a first data
+format for free. And we also propose this format to be used as default
+one, or a as a lingua franca, that can be used to translate between
+other formats.
+
+# Roadmap
```
ivg: you can remove the Roadmap section, as it is already implemented

lib/bap_trace/bap_trace.ml
```diff
@@ -0,0 +1,229 @@
...
+  let meta = reader.Reader.meta in 
+  let next = wrap_next reader.Reader.next in
+  let events = make_stream next () monitor in
+  {id; meta; tool; events; proto = Some proto;}
+
+let map_result f = function
+  | Ok v -> Ok (f v)
+  | Error err -> make_error err
+  
+let load ?(monitor=`Fail) uri =
```
ivg: FYI, our usual convention for monadic code is to put lambda's on the same line as binding operator,
```ocaml
find_proto uri >>= fun proto ->
find_proto writers proto >>= fun write ->
write uri t |>
map_result ident
```

In that case each new operation starts at the beginning of a line, thus making the code more readable.

lib_test/bap_trace/test_trace.ml
```diff
@@ -0,0 +1,281 @@
+open OUnit2
+open Core_kernel.Std
+open Or_error
+
+open Bap_types.Std
```
ivg: no,no,no, don't do this! All tests depend only on `Bap.Std`, and are external. 

P.S., you've made the same error on the image test, that always breaks the compilation with inconsistent assumptions.

PR #309, opened 2015-10-27
Add PFSM assisted shingled disassembler, lifter and unit tests

KennethAdamMiller:
This commit adds a shingled disassembler and lifter
and makes it available under
Bap.Std.Disasm_expert.{Shingled, Shingled_lifter}.
It works in a Continuation Passing Style as much as
realistically possible, and utilizes a single pass
analysis to traverse possible code sequences down to
known data, wherein such can be sheered off. Additionally,
the skeleton implementation of a training corpus facility
to build a probabilistic finite state machine for future
higher accuracy selections of possible code sequences, and
heurism facilitation is provided. Lastly, a lifter module
grants users an easy ability to drive the shingled
disassembler as either a graph sheering capability or a
simple conservative pass. Execution speed is expected to
be all but linear save for the usage of memmap, which forces
theoretical considerations to judge O(n*log(n)). Memory is
 expected to be linear, O(n). 

It is duly noted that precise conformance to the original paper's
original graph sheering techniques may result in at maximum
two edge cases possibly missed; that is, sheering occurs
only on one half of the expected potential cases that can be
affected. At the current time, supposing a sequence of static
successors lead to a known data, their reverse transitive closure
is traversed and subsequently marked data. However, 1) in the case
that a known data proceeds a body of instructions, those
instructions are missed. Additionally, 2) there is at least a
scenario wherein a sequence of data bytes that are static
successors to expected valid code are sheered off by means of
reverse static successor traversal. However, testing and data
driven knowledge will inform future recourse, and additionally
various future improvements through use will also advise 
any potential performance improvements.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -268,6 +268,8 @@ Library disasm
                  Bap_disasm_x86_types,
                  Bap_disasm_x86_utils,
                  Bap_disasm_basic,
+                 Bap_disasm_shingled,
+                 Bap_disasm_shingled_lifter,
                  Bap_disasm_block,
                  Bap_disasm_block_intf,
                  Bap_disasm_insn,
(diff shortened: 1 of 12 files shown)


ivg:
Thanks! This is a great contribution!

PR #311, opened 2015-10-30
Add ansi color tagging, fix few bugs.

ivg:
This is a small technical update. Other than few bug fixes this and
there it brings a new feature: ansi color tags.

Ansi colors
===========

We now have two new tags: `foreground` and `background`. They can be
used to add attributes to IR terms. The pretty printing function of
this values will print corresponding color in ANSI escape codes. So,
the only thing that you need is to mark you term with some color, e.g.,

```ocaml
Term.set_attr t foreground `green
```

or

```ocaml
Term.set_attr t background `red
```

And then if you enable attribute printing, you will see colors, if your
terminal supports this. Just in case if you forgot how to enable
printing attributes a gentle reminder: use `--emit-attr` option, like
this,

```
bap --emit-attr=foreground
```

to get a colored foreground, or,

```
bap --emit-attr=background
```

for background. Use shell globbing (if you're using shell), to get both:

```
bap --emit-attr={fore,back}ground
```

To enable colors in your emacs compilation buffer, add the following to
your .emacs

```lisp
(require 'ansi-color)
(defun colorize-compilation-buffer ()
  (toggle-read-only)
  (ansi-color-apply-on-region (point-min) (point-max))
  (toggle-read-only))
(add-hook 'compilation-filter-hook 'colorize-compilation-buffer)
```

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -3017,6 +3017,16 @@ module Std : sig
   (** Color something with a color  *)
   val color : color tag
 
+
+  (** print marked entity with the specified color.  (the same
+      as color, but pretty printing function will output ascii escape
+      sequence of corresponding color.  *)
+  val foreground : color tag
+
+
+  (** print marked entity with specified color. See [foreground].  *)
+  val background : color tag
+
   (** A human readable comment *)
   val comment : string tag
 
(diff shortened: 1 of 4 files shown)


rvantonder:
This is nice :art:

PR #333, opened 2016-01-21
Enable caching of built deps

maurer:
This caches the contents of the `~/.opam` folder which should speed up the build by skipping rebuild of dependencies when unnecessary. Compare:
[No Cache](https://travis-ci.org/BinaryAnalysisPlatform/bap/builds/103706830) / [Cache](https://travis-ci.org/maurer/bap/builds/103887970)

Oddly, no speedup is observed here, but I am tempted to chalk that up to the immense variability of the time used to download components. It might also be possible with better understanding of the test script you're using to cause the fetched software to be cached as well.


ivg:
Maybe the speed up would be on consecutive builds? 

Anyway, thanks!

PR #337, opened 2016-02-02
Bap_bytes module

gitoleg:
This commit intends to switch core_kernel version to >= "112.35.0" and < "113.00.0"

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -608,6 +608,8 @@ module Std : sig
     val pkg_version : string
   end
 
+  type bytes = Bytes.t 
+
   type 'a reader
   type 'a writer
 
--- a/lib/bap_types/bap_bytes.ml
+++ b/lib/bap_types/bap_bytes.ml
@@ -1,0 +1,83 @@
+
+module Std_bytes = Bytes
+
+open Core_kernel.Std
+
+module Binable = Bin_prot.Utils.Make_binable(struct
+    module Binable = String
+    type t = Std_bytes.t
+    let to_binable = Std_bytes.unsafe_to_string
+    let of_binable = Std_bytes.of_string 
+  end)
+
+module Stringable = struct
+  type t = Std_bytes.t
+  let of_string = Std_bytes.of_string
+  let to_string = Std_bytes.unsafe_to_string
+end
+
+module Sexpable = Sexpable.Of_stringable(Stringable)
+
+module T = struct
+  include Std_bytes
+  include Binable
+  include Sexpable
+end
+
+module Elt = struct
+  type t = char
+  let equal t t' = Char.compare t t' = 0
+  let of_bool b = if b then 'a' else 'b'
+end
+
+module Seq = struct
+  type elt = Elt.t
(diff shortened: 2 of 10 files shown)


lib/bap/bap.mli
```diff
@@ -608,6 +608,8 @@ module Std : sig
     val pkg_version : string
   end
 
+  type bytes = Bytes.t
```
ivg: add `with bin_io, compare, sexp`

lib/bap_types/bap_bytes.ml
```diff
@@ -0,0 +1,83 @@
+
+module Std_bytes = Bytes
```
ivg: this should be `module Bytes = Bap_bytes`

lib/bap_types/bap_bytes.ml
```diff
@@ -0,0 +1,83 @@
...
+end
+
+module Blit = Blit.Make(Elt)(Seq)
+
+module Identifiable = Identifiable.Make(struct
+    type t = T.t with bin_io, compare, sexp
+    let of_string = Stringable.of_string
+    let to_string = Stringable.to_string
+    let hash = Hashtbl.hash
+    let module_name = "Bap bytes"
```
ivg: it should be `Bap.Std.Bytes`. The module name is used to register printers in a toplevel. See `Printable` and `Install_printers` modules.

lib/bap_types/bap_bytes.mli
```diff
@@ -0,0 +1,13 @@
+
+
+module Std_bytes = Bytes
```
ivg: why do you need this module in the interface?

lib/bap_types/bap_bytes.mli
```diff
@@ -0,0 +1,13 @@
...
+
+open Core_kernel.Std
+
+type t = Std_bytes.t with bin_io, compare, sexp
```
ivg: should be just `type t = bytes with bin_io, compare, sexp`

lib/bap_types/bap_bytes.mli
```diff
@@ -0,0 +1,13 @@
...
+
+include module type of Std_bytes with type t := t
```
ivg: `include module type of Bytes with type t := t`

opam
```diff
@@ -41,7 +41,7 @@ depends: [
     "bitstring"
     "camlzip"
     "cmdliner" {>= "0.9.6"}
-    "core_kernel" {>= "112.17.0" & < "112.35.0"}
+    "core_kernel" {>= "112.35.0" & < "113.00.0"}
```
ivg: can you, please, split this in two different PRs?
One that adds `Bytes` module, and the other, that actually bumps the version up.

lib/bap/bap.mli
ivg: and where is the module `Bytes`? at all?

ivg:
1. Split into two different PRs
2. Expose `Bytes` module in BAP interface
3. Add tests

PR #343, opened 2016-02-25
bap future library with tests

gitoleg:


lib/bap_future/bap_future.ml
```diff
@@ -0,0 +1,503 @@
...
+            drop ();
+            process ()
+          | _ -> () in
+        process ();
+        let len,l1,l2 = Queue.(length q, length q1, length q2) in
+        if Queue.capacity q = len then wait src;
+        if l1 > capacity && l2 > capacity && ratio l1 l2 > 2 &&
+           (len > l1 || len > l2) then  wait src in
+      link s1 s (step s1 q1);
+      link s2 s (step s1 q2);
```
ivg: shouldn't his be `link s2 s (step s2 q2)`?

gitoleg: yes, I think it should

PR #350, opened 2016-03-07
split each library into a separate oasis section

ivg:
* wrote automation that glues everything together
* fixed bap-fsi-benchmark
* fixed dependencies for uninstall target in Makefile, that broke uninstallation

Now, every feature that is provided by the bap repository is described
in a separate file in the oasis folder. The configure scripts is
responsible to glue them together based on the `--enable-<feature>`
options. There is a `--enable-everything` flag, that is useful for
testing and development. The default way of configuring from the source
repository would be:
```
  ./configure --enable-everything --prefix `opam config var prefix`
```


KennethAdamMiller:
Awesome, now builds can be extremely fast by targeting only what has changed. Example of a command to build a specific subproject, file or entire bap?

ocamlbuild bap_future.cmxa?

ivg:
No, the example would be:

```
./configure --enable-future
make 
```

And yes, the build is now slightly faster, due to the switch to `ppx`. But not as fast, as I expected.

PR #355, opened 2016-03-07
added lost module Bap_trace_id

gitoleg:
Bap_trace_id module wasn't mentioned in oasis file, and this caused an undefined module error while linking with bap-traces library.


ivg:
I wonder why this didn't fail on travis test...

PR #359, opened 2016-03-08
Imported header plugin from bap-plugins

ivg:
The new name for the header is api. The pass utilizes
the new `autoload` feature, and applies known APIs on
the program. APIs are bundled with the plugin itself.
New API's can be added by providing a C header file.

### DIFF
--- a/oasis/api
+++ b/oasis/api
@@ -1,0 +1,12 @@
+Flag api
+  Description: Build API plugin
+  Default: false
+
+Library api_plugin
+  Build$:           flag(everything) || flag(api)
+  Path:             plugins/api
+  FindlibName:      bap-plugin-api
+  CompiledObject:   best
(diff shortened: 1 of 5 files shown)


rvantonder:
It might be a little hard for someone outside of BAP to figure out what api means, because it's such a general term. Maybe scope it for function or subroutine APIs? But I don't have a better suggestion right now. Otherwise, happy to see this in main BAP :+1:

plugins/api/Makefile
```diff
@@ -0,0 +1,5 @@
+header.plugin : header.ml
```
rvantonder: Doesn't look like there's a `header.ml` file in this commit.

ivg: ah, forget to remove the Makefile :)

ivg:
The name is chosen from the perspective of a user, how it will invoke it: 
```
bap --api-add odbx.h 
```
Other than this, there is no need to call it, it will work out of box.

PR #368, opened 2016-03-15
resolve PLT entries

ivg:
TL;DR
-----

No more IDA. Ok, at least on ARM. This is a hacky support for resolving
PLT entries, that works at least on ARM ELF, i.e., on our most favorite
targets. Support for other targets might be added later, but there are
problems.

Implementation details
----------------------

If an undefined symbol has a value, then use it as an address.
Since the value size is a not a size of a plt entry, but a size of the
real symbol, that would be loaded (at least for elf-arm), a hack is used
to hardcode the size to 8 bytes.

This will not work for x86-64, since the value is usually zeroed by the
elf64-x86-64 backend. In fact, only for elf64-x86-64 backend there are
36 different kinds of relocations, each giving their own meaning to the
entries in the symbol table. And each backend define its own relocations
and policies. So, the resolution of the entries should be performed on
the backend level, before a transformation to an abstract symbol or
relocation.

How `objdump` works
-------------------

You might be interested how `objdump` resolves the plt entries. First of
all it actually performed on a backend side. So for each possible
target, and supported relocation it is different. For `elf64-x86-64` and
`R_X86_64_JUMP_SLOT` relocation, it will rely on a fact, that ith
relocation corresponds to the ith PLT entry. Since BFD knows the size of
PLT entry, it is easy to find out the corresponding PLT entry. Of
course, this is fragile and works only for binaries created by a libbfd,
or someone, who tries to conform to it. We cannot rely on this in the
loader backend. But we can easily implement the same logic in a separate
plugin, that will provide a symbolizer for `elf64-x86-64`.

Current status
--------------

Works for elf32-arm. Shouldn't break anything on other targets. Support
for other targets can be added as separate plugins, on per target
basis. Nothing magic here, just lots of reverse engineering.

### DIFF
--- a/lib/bap_image/bap_image.ml
+++ b/lib/bap_image/bap_image.ml
@@ -331,41 +331,42 @@ let memory_of_symbol {memory_of_symbol = lazy f} = f
 let symbols_of_segment {symbols_of_segment = lazy f} = f
 let segment_of_symbol {segment_of_symbol = lazy f} = f
 
-let%test_module _ = (module struct
+let%test_module "image" =
+  (module struct
+    let expect ?(print=false) ~errors data_size ss =
+      let width = 32 in
+      let segment (addr, off, size) = {
+        Backend.Segment.name = "test-segment";
+        Backend.Segment.perm = R;
+        Backend.Segment.off;
+        Backend.Segment.location = {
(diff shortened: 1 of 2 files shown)


rvantonder:
Whaaat. Great :100:

PR #377, opened 2016-03-17
Introduce microx library

ivg:
This is a microexecution framework. It can execute any code,
without user input or anything else. The library provides so far two
classes:
- concretizer that implements different concretization policies for
  undefined values;

- conqueror that will execute programs notwithstanding errors,
  undefined values or infeasibility. Conqueror stands for Concrete
  evaluation with Errors.

Note this library is released from quarantine.

This PR also fixes a small issue noticed by @gitoleg. When we try to
store to undefined memory we must return bot, instead of freshly create
new empty memory.


ivg:
> This PR also fixes a small issue noticed by @gitoleg. When we try to
store to undefined memory we must return bot, instead of freshly create
new empty memory.

Correction: when we try to store to undefined *address*

PR #382, opened 2016-03-18
Add "use file" option to api plugin

rvantonder:
For discussion. Api currently adds the header file to the database (zip)
without checking that it parses correctly first. So if you give it a file that
doesn't parse, it will try to use this broken file every time. Since there is
no easy way to remove the header (that I found), I just add an extra option
which will let you use a header file without using the add feature.

Also adds camlzip to .merlin

### DIFF
--- a/plugins/api/api_main.ml
+++ b/plugins/api/api_main.ml
@@ -22,6 +22,9 @@ type arg = {
 
 type fn_proto = arg list
 
+type options = {add_headers : string list;
+                use_header : string option}
+
 exception Failed_to_parse of string
 
 include struct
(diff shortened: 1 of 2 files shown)


ivg:
I like the idea, but I have a suggestion. Let's do the following, if there is a `--api-add` option, then we just add new api to the bundle and exit with 0, e.g.,
```
bap --api-add odbc.h
```

When there is a `--api-use` option, then it will apply this file without adding it to a bundle. In fact, I think, that `--api-file` will sound better:

```
bap --api-file odbc.h
```

What concerning file deletion from a bundle, then we should also have this option. To remove a file with a `name`, just use 
```ocaml
   Bundle.update bundle ~f:(fun file -> if file = name then `Drop else `Copy)
```

In fact it would be a good idea to add function `remove` to the `Bundle` interface, that will do this. As this update is powerful, but not very obvious.

dbrumley:
This sounds like a good idea to me as an outside.  One question: should bap respect CPATH and similar environment variables (https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html) when you specify a filename?

rvantonder:
@ivg sure, I can update this. But also, thinking more about it, we should still check if the file parses before adding files that don't parse to the bundle. Something like this works (but `ignore` is ugly, any better way to achieve this?):

```ocaml
let insert_headers bundle file =
  let name = "api/" ^ Filename.basename file in
  try
    args_of_file file |> ignore;
    Bundle.insert_file ~name bundle (Uri.of_string file)
  with
  | Parsing.Parse_error → failwith "Could not parse header file."
```

ivg:
That's an interesting idea, to include CPATH. It will not be hard to do this, but the problem is that right now we use `FrontC` as C parser, that is very fragile and is unable to pass most of not all standard headers (mostly due to the fact they they use lots of `#pragmas` and gnu extensions). Looking into CPATH may elude people into thinking that the can just say `--use-api stdio.h`. Unfortunately that's not true.

ivg:
> but ignore is ugly, any better way to achieve this?

It's ok to bail out here, just throw an exception or exit with non-zero status, outputting error information. Here we're in a frontend, and should react to user's actions. If a use said `bap --add-api odbc.h`, the you should either add it or reply with an error message as soon as possible.

dbrumley:
> but the problem is that right now we use `FrontC` as C parser, that is very fragile
Shoot. Well, probably not worth it then.

plugins/api/api_main.ml
```diff
@@ -186,33 +189,55 @@ let fill_args arch fns program =
...
+  | Parsing.Parse_error -> failwith "Could not add header file: parse error."
 
 let () =
   let bundle = main_bundle () in
-  List.iter (Cmdline.parse argv) ~f:(insert_headers bundle);
-  Project.register_pass ~autorun:true (main bundle)
+  let options = Cmdline.parse argv in
+  match options.add_headers with
+  | [] ->  Project.register_pass ~autorun:true (main bundle options)
+  | headers -> List.iter headers ~f:(add_headers bundle);
```
ivg: That's a little different behavior. It should be
```
$ bap --api-add odbc.h
```
But you will get is
```
$ bap --api-add odbc.h
bap: required argument FILE is missing
Usage: bap [OPTION]... FILE
Try `bap --help' for more information.
```
Just add `exit 0`  after this line.

rvantonder: Also needed something like this on line 237

ivg:
I can't merge it, because `This branch is out-of-date with the base branch`
Can you update it?

P.S. github is boring....

rvantonder:
If only github could make 'update branch' rebase the PR onto master instead of merging. ¯\\(ツ)\/¯

ivg:
yeah ツ

PR #405, opened 2016-04-12
Missing microx in .merlin

rvantonder:


.merlin
```diff
@@ -20,6 +20,7 @@ B _build/lib/bap_plugins
 B _build/lib/text_tags
 B _build/lib/regular
 B _build/lib/graphlib
+B _build/lib/microx
```
ivg: The root `.merlin` file should only contain core libraries, like bap itself and its ancestors. 

This is the correct contents of the `B` section of `.merlin` in mine repository (not yet merged):

```
B _build
B _build/lib/bap
B _build/lib/regular
B _build/lib/graphlib
B _build/lib/bap_future
B _build/lib/bap_plugins
B _build/lib/bap_bundle
B _build/lib/bap_config
```

A code that uses some library, should create its own local `.merlin` that refers to it. If we will put everything into the root file, it will work very slowly.

rvantonder: Ah. Then it should go in `propagate_taint` .merlin, since that is where I wasn't resolving it. https://github.com/BinaryAnalysisPlatform/bap/blob/master/plugins/propagate_taint/.merlin

rvantonder: Hmm.. it has this line https://github.com/BinaryAnalysisPlatform/bap/blob/master/plugins/propagate_taint/.merlin#L3 but for some reason didn't work. Maybe I didn't reinstall with the microx package.

ivg: this is a bad line... It will refer to a library visible via the findlib, not to the library compiled in `_build`. So consider the following: you're updating a library, and didn't reinstall it yet, because, plugin, depending on the library doesn't compile. In that case, Merlin will see the old (installed one) library, not the updated, where the compiler will see that library that is in the `_build`. Bad things may happen. 

I'm actually, currently updating merlin files, and trying to impose some order on them. So it is better not to touch them. Otherwise we will have lots of conflicts :)

ivg:
Should be fixed by #406

PR #415, opened 2016-04-27
Adds objdump parser as a symbolizer

dbrumley:
objdump output will be parsed to identify function starts.  The path to
objdump is specified either via a command line option, or the plugin
will search your path for `objdump` (default in general) or
x86_64-elf-objdump i386-elf-objdump (default in macports)

### DIFF
--- a/plugins/objdump/README.md
+++ b/plugins/objdump/README.md
@@ -1,0 +1,52 @@
+# Overview
+
+This plugin uses `objdump` to extract symbols from a binary. It does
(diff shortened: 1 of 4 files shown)


plugins/objdump/README.md
```diff
@@ -0,0 +1,52 @@
+# Overview
```
ivg: instead of `Readme.md` we put all the information into the `man` section of the plugin. Rationale - the readme is not available to a user, and is not installed. Having more than one source of information is bad.

Issue #414, opened 2016-04-25
bap may fail on macho binaries


dbrumley:
BAP is failing on my OSX El Capitan install:
```
$ bap /bin/ls 
LLVM ERROR: getSymbolValue unimplemented in MachOObjectFile
$ bap --version
0.9.9
$ ls /opt/local/bin/llvm-config-mp-3.*
/opt/local/bin/llvm-config-mp-3.4 /opt/local/bin/llvm-config-mp-3.5 /opt/local/bin/llvm-config-mp-3.7
```

KennethAdamMiller:
I am on the latest mac and have had some trouble with this. My fix was to run llvm-config to select a particular llvm instance to default to. The latest llvm port package installs manage multiple versions with reconfigurable symlink trees like they should have ages ago. I think the command I used to fix this is:

sudo port select llvm mp-llvm-3.4

The other thing I've noticed about el capitan and bap is that locally built packages are failing with a no llvm backend error. But the bap-mc installed under opam resolves correctly... not sure what's up with that.

ivg:
> The other thing I've noticed about el capitan and bap is that locally built packages are failing with a no llvm backend error. But the bap-mc installed under opam resolves correctly... not sure what's up with that.

I don't know what is "locally built package" for you, but this error suggests that `llvm` plugin is not loaded. To load `llvm` plugin you need to compile, install and load it. You can compile and install with opam, and load with `Plugins.load`.

KennethAdamMiller:
These locally built binaries are those that used to work using the Bap.Std
interface but not a plugin. Thanks for the tip, I'll add the code to be
peeled away later.

ivg:
fixed by #418, confirmed manually.

PR #418, opened 2016-05-03
Fix PE32 support and few bug fixes.

ivg:
There were few bugs in the LLVM, all in the upstream. This
PR fixes them on our side.

* Address of the segment was incorrect
* Address of symbol was incorrect
* Size of symbol was incorrect

The first issues was due to the use of the relative virtual address,
instead of virtual. They are fixed by adding the image base address.

The latter issue was quite funny. The algorithm that calculated address
in llvm was = "subtract symbol address from section size" (sic). As an
excuse the author wrote a comment "FIXME: Return the correct size."
Indeed, it was quite hard to get the correct size, as the correct
algorithm is quadratic.

But now we can discover symbols from PE32, even without usage of
external tools like IDA or objdump.

Two other bugs were:

1. objdump plugin didn't work for 32-bit binaries, as addresses was
always 64 bit.

2. internal rooters and symbolizers didn't work due to a logic error in
the DNA of the author.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-install: wget https://raw.githubusercontent.com/talex5/ocaml-travisci-skeleton/containers/.travis-opam.sh
 script: bash -ex .travis-opam.sh
 sudo: false
 addons:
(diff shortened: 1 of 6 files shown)


ivg:
and this also fix #414, forget to notice this.

PR #422, opened 2016-05-06
detect binutils installation at configuration time

ivg:
Alike as IDA, objdump now expects a path to objdump to be passed
at configuration time. A corresponding `conf-binutils` package will
be added to opam and the dependency between them will be established.

As a consequence, `bap-objdump` package will not be installable, if
there is no objdump program, moreover, a corresponding external
dependency will be added and thus installed via depext.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ _oasis
 /man/bap.1
 /lib/bap_config/bap_config.ml
 /lib/bap_ida/bap_ida_config.ml
+/plugins/objdump/objdump_config.ml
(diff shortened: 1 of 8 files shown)


dbrumley:
I do not understand this comment.  The implication seems to be the plugin is not installed if `conf-binutils` cannot find `objdump`.  But then this would negate to a large extent the need for a path at all. However, having a path to objdump seems nice on OSX.  

You mentioned something about `gobjdump`, but that does not correctly identify the file format, e.g., at least on my osx system it will not open elf, while `/opt/local/bin/i386-elf-objdump` will for i386.

I wonder if trying to auto-find/configure for objdump is really a good idea, compared to searching a few default options like previously else requiring an explicit path.

dbrumley:
To be more explicit, on OSX I need both `gobjdump` and the elf-objdump variants.  Trying to autoconfigure here seems likely to cause problems where the plugin doesn't install but could really be used.

ivg:
I've taken care of this. The conf-binutils script will find all installations of objdump, and report it to the plugin. The main idea is not to prohibit from installation, but instead provide a mechanism to install the dependencies in a crossplatform way, e.g., now `opam depext bap` will install proper objdump dependencies on both, macosx and linux, that will work with all supported targets. 

Currently there is still a small issue, as I do not take gobjdump into account, as on mine previous port version, there was a normal objdump available. But this is easy to fix anyway.

PR #429, opened 2016-05-17
Improve `--emit-ida-script`

jaybosamiya:
Rewrote most of the `--emit-ida-script` plugin.

The plugin now allows for arbitrary attributes to be inserted into IDA. It stores the attributes in an s-expression based key-value dictionary of the form `(BAP (key_1 value_1) (key_2 value_2) ...)`. By not clobbering any comments which are outside the `(BAP ...)` structure, the user can manually annotate further and still continue to use BAP. Also, the dict is updated, only by clobbering values of the same key (i.e. if one script updates `key_1`, and another updates `key_2`, then they won't affect each other).

A special attribute `color` is used to specify the color with which to colorize the respective address of the attribute.

Help is available through `--emit-ida-script-help`.

Here's a demonstration: 
Running the script: https://asciinema.org/a/5shmd7lnnxorxatcyphv3astg
Screenshots: https://gist.github.com/jaybosamiya/7013ddf22a30103e7f635a1a7591ab7e (Read the README there first)

Thanks a lot @ivg for helping me through my first plugin for BAP :)


ivg:
This is a great work, thanks!

PR #431, opened 2016-05-20
IDA Python Plugin to propagate taint

jaybosamiya:
This PR adds an IDA Python Plugin to be able to select any arbitrary line in Graph/Text view in IDA, and be able to taint that line and propagate taint information and see it visually.

### Keybindings
- Shift-A : Equivalent to `--taint-reg` in BAP
- Ctrl-Shift-A : Equivalent to `--taint-ptr` in BAP

The plugin is loaded automatically into IDA for usage, after installation.

### Testing
```
./configure --prefix=`opam config var prefix` --enable-ida-plugin
make && make reinstall
```

Then, just select some line by placing your cursor over it in IDA, and press Shift-A or Ctrl-Shift-A.

### Color Scheme
- "Nasty" Yellow : Taint source
- "Affected" Red : Lines that got tainted
- "Ignored" Gray : Lines that were not visited by propagate-taint
- "Normal" White : Lines that were visited, but didn't get tainted

#### Taint Source, and Tainted Locations

screenshot

#### Unvisited Regions

gray

PR #433, opened 2016-05-24
Hex-Rays decompiler/pseudocode Taint Plugin

jaybosamiya:
This PR adds an IDA Python Plugin to be able to propagate and see taint information in the decompiler/pseudocode view.

The plugin depends on previously added [`BAP_Taint` plugin](https://github.com/BinaryAnalysisPlatform/bap/pull/431), and extends it further.

As with the previous plugin, this plugin too is loaded automatically into IDA for usage, after installation.

### Changes

+ All files `plugins/ida/python/*.py` act as IDA Plugins henceforth, and are installed automatically to the IDA Plugins user folder
+ `BAP_Taint` (i.e. `bap_propagate_taint` plugin) now accepts callbacks which can be installed so as to make it more extensible
+ Add Hex-Rays plugin to be able to see the taint analysis propagated from the Text/Graph View, whenever the pseudocode view is loaded/reloaded
+ Install callbacks into the `BAP_Taint` plugin, so that the Hex-Rays plugin is called whenever re-analysing taint, thereby updating taint information in the pseudocode view (and also allowing for taint to be propagated _from_ pseudocode view)

### Testing

Same as `BAP_Taint` plugin:

```
./configure --prefix=`opam config var prefix` --enable-ida-plugin
make && make reinstall
```

Then, just select some line by placing your cursor over it in IDA, and press Shift-A or Ctrl-Shift-A.

### Color Scheme

Same as `BAP_Taint` plugin:

+ "Nasty" Yellow : Taint source
+ "Affected" Red : Lines that got tainted
+ "Ignored" Gray : Lines that were not visited by propagate-taint
+ "Normal" White : Lines that were visited, but didn't get tainted

#### Pseudocode View
screenshot-1

#### Corresponding Text View
screenshot-2


plugins/ida/python/propagate_taint_bap_hexrays.py
```diff
@@ -0,0 +1,121 @@
...
+                'red',
+                'yellow',
+            ]
+        ]
+
+        BGR_MASK = 0xffffff
+
+        ea_color = idaapi.get_item_color(ea)
+
+        assert(ea_color & BGR_MASK in coloring_order)
```
ivg: I'm a little bit worried, what if the would be some coloring here made by IDA or some other ida plugin here, that will have a color that is not bapish. Will it provoke this assert? And if it is, then do we really want it?

jaybosamiya: Hmm.. Good catch.. The `assert` would be triggered and we don't want it like that. Will make a modification and push that commit here.

ivg:
Great work, thanks!

PR #435, opened 2016-05-24
IDA symbolizer fix

jaybosamiya:
This PR takes care of a couple of things:

+ There were some unnecessary files (`lib/bap_ida/bap_ida_scripts.ml`, `lib/bap_ida/bap_ida_scripts.mli`) that were unused by the current code. This removes them.

+ The IDA symbolizer sometimes (i.e. at least for gcc compiled x86 and x86-64 binaries) returned _slightly_ modified names (i.e. usually with a preceding `.` or `_`). This PR fixes that.

### Evidence of issue

Before the PR, run the following commands

```bash
cat > test.c <<EOF
#include <stdlib.h>

int main() {
    malloc(0);
}
EOF

gcc test.c -o test # on an x86 or x86-64 machine, alternatively use gcc-multiarch and compile to x86
bap --symbolizer=ida test -d | grep malloc
```

You will see output similar to

```
0000002d: sub .malloc()
00000149: call @.malloc with return %0000014a
```

### Demonstration of Fix

However, after the proposed changes are made, the output of the same is
```
00000146: call @malloc with return %00000147
00000150: sub malloc(malloc_size, malloc_result)
0000019c: malloc_size :: in u64 = RDI
0000019d: malloc_result :: out u64 = RAX
```
which is obviously more useful.

**Note:** You might need to run `bap --cache-clean` to clear cache between runs, to see the updated result.

### Additional Info

Running `make test` in the [saluki plugin directory](https://github.com/BinaryAnalysisPlatform/bap-plugins/tree/master/saluki) gives the expected output:

```
tests/test0.c                                               ok
tests/test1.c                                               ok
tests/test10.c                                              ok
tests/test11.c                                              ok
tests/test12.c                                              ok
tests/test13.c                                              ok
tests/test14.c                                              ok
tests/test2.c                                               ok
tests/test3.c                                               ok
tests/test4.c                                               ok
tests/test5.c                                               ok
tests/test6.c                                               ok
tests/test7.c                                               ok
tests/test8.c                                               ok
tests/test9.c                                               ok
all ok
```

Also, with the proposed changes, saluki now works directly on these same `.c` files even on x86 and x86-64 compiled ELF files.


jaybosamiya:
Well, I'm not sure why Travis Check would fail due to the latest change (seeing that the previous change passed, and this change brings about a minor change (of exactly one line, manually verified).

The error is in https://travis-ci.org/BinaryAnalysisPlatform/bap/builds/132693058#L1746-L1747 which I'm not sure how my change should have affected.

PR #437, opened 2016-05-25
Fix HexRays Taint Plugin for IDA Demo

jaybosamiya:
If one were to install IDA Python into IDA Demo using the instructions [here](https://github.com/idapython/src), then we can use BAP's taint plugin in IDA. But, the new plugin (i.e. for decompilation view) causes a nasty looking error (which basically just means that there is no decompiler support). This PR fixes that. Also, this PR moves `import tempfile` into a smaller scope.


ivg:
hmm... the rebase button doesn't work for me, can you update the PR to the upstream/master?

jaybosamiya:
Merged `upstream/master`.

Issue #117, opened 2015-03-05
`mov pc, lr` is unlifted


ivg:
ARM lifter doesn't handle this 
```asm
0e f0 a0 e1    mov pc, lr          ; MOVPCLR(0xe,Nil)
```

There is nothing special in it, so it can be added fairly easily. Volunteers are welcome!

jaybosamiya:
This issue can be closed now, right?

ivg:
sure!

PR #439, opened 2016-05-25
Add MOVPCLR to the ARM Lifter

jaybosamiya:
I noticed issue #117 which said that the ARM lifter doesn't handle `mov pc, lr`, so I decided to look into the ARM lifter. Here's a PR that adds MOVPCLR to the lifter.

---

Since I've never written code for a lifter before, please let me know if I made any mistakes (or any style errors etc), and I will correct them.

---

### Testing it

Running the following code:

```bash
echo '0e f0 a0 e1' | bap-mc --arch=arm --show-insn=pretty --show-bil=pretty --show-bir=pretty
```

We get the output as

```
MOVPCLR(0xe,Nil)
{
  jmp LR
}
00000004: 
00000005: return LR
```


ivg:
nice!

PR #441, opened 2016-05-26
Restructure IDA Plugins

jaybosamiya:
This PR restructures IDA Plugins in BAP to move the rooter, symbolizer and reconstructor into `bap-ida`, and keep the rest of the plugins in `bap-ida-plugins` to remove the `bap-ida` dependency. A further PR that I'll be making in the [BAP opam repository](https://github.com/BinaryAnalysisPlatform/opam-repository) will fix up the build process in opam.

### Advantages

+ Users can now use BAP along with IDA Demo (after [installing IDA Python onto it](https://gist.github.com/jaybosamiya/951193ae774dfc127b513a93795a46c7))
+ `bap-ida` now means "all integrations for BAP with IDA (that need IDA Pro)"
+ `bap-ida-plugins` now means "BAP/IDA plugins which work with both IDA Pro and IDA Demo"

### DIFF
--- a/oasis/ida
+++ b/oasis/ida
@@ -13,3 +13,12 @@ Library bap_ida
   InternalModules:  Bap_ida_config
   BuildDepends:     fileutils, re.posix
   XMETADescription: make calls into IDA
+
+Library bap_ida_rsr
+  Build$:           flag(everything) || flag(ida)
+  Path:             plugins/ida
+  FindlibName:      bap-plugin-ida_rsr
+  CompiledObject:   best
+  BuildDepends:     bap, bap-ida, cmdliner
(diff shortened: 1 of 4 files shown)


oasis/ida
```diff
@@ -13,3 +13,12 @@ Library bap_ida
   InternalModules:  Bap_ida_config
   BuildDepends:     fileutils, re.posix
   XMETADescription: make calls into IDA
+
+Library bap_ida_rsr
+  Build$:           flag(everything) || flag(ida)
+  Path:             plugins/ida
+  FindlibName:      bap-plugin-ida_rsr
```
ivg: As I told you, the name should be preserved :(

PR #442, opened 2016-05-26
Add aspcud to Travis build

jaybosamiya:
In Travis builds, to prevent issues with opam messing up with the constraints (as we noticed recently), we can use the recommended external dependency solver `aspcud`. This PR adds it.


ivg:
yay! Thanks :)

PR #446, opened 2016-05-31
Plugin for Hex-Rays Decompiler (BAP comments + abstract base plugin + refactor)

jaybosamiya:
This PR does quite a few changes:

+ Add a `bap_utils` library for usage in IDAPython
    + This adds functions like `sexp2list`, `list2sexp`, `cfunc_from_ea` etc, which are extremely useful when writing plugins for IDA, especially for standardizing to using s-expressions.  
    + This also adds a `DoNothing` plugin which can be used to build further libraries, or for making abstract plugins (used later)
+ Add the `SimpleLine_Modifier_Hexrays` abstract base class plugin which is useful  for implementing HexRays (decompiler) plugins by abstracting away a large part of the implementation details
+ Refactor the `BAP_Taint_Pseudocode` plugin to use the abstract base class `SimpleLine_Modifier_Hexrays`.
+ Add a Hex-Rays plugin to propagate BAP comments from Text/Graph View to pseudocode view
+ Ensure that the `BAP_Taint` plugin does not unnecessarily re-initialize itself upon opening a new binary in the same window.

### Test/Screenshots

If we were to add BAP comments (i.e. using the s-expression style) in the text view (here, it is done manually, for testing, but otherwise, the `--emit-ida-script-attr` is a useful way to add BIR attribute information), then we propagate the information to all the related lines in decompilation view.

screenshot1
screenshot2

### DIFF
--- a/plugins/ida/python/bap_utils.py
+++ b/plugins/ida/python/bap_utils.py
@@ -7,19 +7,22 @@ def sexp2list(s):
     """Convert S-Expression to List."""
     sexp = [[]]
     word = ''
+    in_str = False
     for c in s:
-        if c == '(':
+        if c == '(' and not in_str:
             sexp.append([])
-        elif c == ')':
+        elif c == ')' and not in_str:
             if word:
                 sexp[-1].append(word)
                 word = ''
             temp = sexp.pop()
             sexp[-1].append(temp)
-        elif c == ' ':
+        elif c in (' ', '\n', '\t') and not in_str:
             if word:
                 sexp[-1].append(word)
             word = ''
+        elif c == '\"':
+            in_str = not in_str
         else:
             word += c
     return sexp[0]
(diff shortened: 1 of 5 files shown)


plugins/ida/python/bap_utils.py
```diff
@@ -0,0 +1,145 @@
...
+def get_bap_comment(comm):
+    """
+    Get '(BAP )' style comment from given string.
+
+    Returns tuple (BAP_dict, start_loc, end_loc)
+        BAP_dict: The '(BAP )' style comment
+        start_loc: comm[:start_loc] was before the BAP comment
+        end_loc: comm[end_loc:] was after the BAP comment
+    """
+    if '(BAP ' in comm:
```
ivg: Am I right, that this algorithm assumes that value inside the sexp will be a well-formed sexp? E.g., something like this `(BAP (message "hello, world))"))` is not allowed?

ivg: it looks like, that there should be som `in_str`-like logic here.

jaybosamiya: Ah, while I added the `in_str` above, I didn't realize that it'd affect here as well. I'll push a new commit modifying this area as well.

PR #447, opened 2016-05-31
Make emit-ida-script plugin more flexible

jaybosamiya:
This PR allows for quoting using `"` in s-expressions, and also allows for a value to have spaces. This ensures that attributes are overwritten correctly whenever necessary.


jaybosamiya:
Since a large part of `emit-ida-script` is duplicated in `bap_utils`, I will make `emit-ida-script` use `bap_utils` once #446 (which introduces `bap_utils`) is merged. Closing this PR since the code after that modification will no longer use the changes here.

PR #449, opened 2016-06-01
Use latest symbols when calling BAP from IDA + Refactoring using bap_utils

jaybosamiya:
Using the `bap_utils.py` library that I added in PR #446, in this PR, I have refactored the `emit-ida-script` plugin, IDA Loader, Symbolizer+Rooter+Reconstructor (by refactoring `get_symbols`)

Also, we now dump and use the latest symbols when calling BAP from IDA.

Also, by moving BAP execution into `bap_utils`, we now have an easy to use function `run_bap_with(argument_string)` which automatically calls bap with the right input file and also gets the latest symbols from IDA into BAP


ivg:
Yeah, this is much nicer now!

PR #451, opened 2016-06-01
extensible api/abi framework

ivg:
Intro
=====

This PR adds a bunch of libraries and passes, that forms a framework for
applying an API information, written in some high level language to a
low-level intermediate representation of a program, i.e., BIR.

At a high level the framework is very generic, as it should allow to
transform a specification written in any language into IR attributes or
terms. The design allows to specify API not only in C or C like
languages, but also in some logic languages, like Frama-C ACSL. Of
course, the latter is not implemented.

An accurate transformation of a high-level specification to the low
level IR usually requires a knowledge of ABI, that can be seen as a
summary of implementation decisions made by compiler developers. ABI are
usually language, architecture and compiler specific. The framework
provides a library for building ABI specifications for C language.

Other than the framework, this PR actually implements an API
transformation for C language and 9 different ABI. The main contribution
here is a complete support of the ARM ABI, that actually required the
implementation of the whole framework. Due to the lack of time, x86 abi
are still incomplete, but they now can be easily fixed.

It is now also possible to choose a specific ABI via the command line,
using `--x86-abi` and `--x86-64-abi` switches, that allows our
api-dependent analyses (e.g., saluki, wur) to work correctly on windows
and QNX files.

New features
============

- partial support for 8 x86/64 ABI:
  * SysV
  * cdecl
  * stdcall
  * ms32
  * ms64
  * fastcall
  * watcom-stack
  * watcom-regs

- nearly complete support for ARM gnueabi
  (only hard floats are not supported)

- reading and apply API specifications
  written in any language

- abstraction of C language, including type system,
  data model

- an extension mechanism for adding support for new
  abi and api specification languages

- an extension mechanism for adding semantic actions to
  GNU attribute

- propagate all C information, including types and data models into the
  program representation

- redesigned demangling system

- a C parser based on FrontC


rvantonder:
Will it have tests?

ivg:
@rvantonder, if someone will write them, then yes :) So far only `saluki` tests are sort of small guarantee, that it works correctly. Another thing, is that our testing infrastructure has reached its limit, so we actually need to invent something new.

PR #454, opened 2016-06-02
Allow disabling certain plugins/features when enabling everything

jaybosamiya:
This PR allows us to specifically disable certain plugins/features when we enable everything, for example, if we run `./configure --enable-everything --disable-ida`, then it will enable every plugin/feature except for `ida`

### DIFF
--- a/tools/oasis_sections.ml
+++ b/tools/oasis_sections.ml
@@ -5,13 +5,11 @@ let header = "common"
 let has_extension name =
   String.contains name '.'
 
-let everything =
-  Sys.readdir "oasis" |> Array.to_list |>
-  List.filter (fun sec -> sec <> header && not (has_extension sec))
+let enabled = "--enable-"
+let disabled = "--disable-"
 
-let enable_feature arg =
+let feature enable arg =
   let length = String.length arg in
-  let enable = "--enable-" in
   let prefix = String.length enable in
   try
     if String.sub arg 0 prefix = enable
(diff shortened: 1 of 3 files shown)


tools/oasis_sections.ml
```diff
@@ -5,31 +5,34 @@ let header = "common"
...
+let selected enable args =
   args |> List.fold_left (fun fs arg ->
-      match enable_feature arg with
+      match feature enable arg with
       | Some f when Sys.file_exists ("oasis" / f) -> f :: fs
       | _ -> fs) []
 
-
 let features args =
+  let disabled_args = (selected disabled args) in
```
ivg: the parentheses here are not needed

jaybosamiya:
The reason Travis fails here (just for documentation purposes):
```
$ ./configure --enable-everything --disable-objdump --disable-ida 2>&1 | grep ida
  --enable-ida-plugin            Build IDA plugin [default: disabled]
  --disable-ida-plugin           Build IDA plugin [default: disabled]
  --ida-plugin-path str          A directory for IDA plugins

$ ./configure --enable-everything --disable-objdump 2>&1 | grep ida 
  --enable-ida                   Build IDA intergration library [default: disabled]
  --disable-ida                  Build IDA intergration library [default: disabled]
  --enable-ida-plugin            Build IDA plugin [default: disabled]
  --disable-ida-plugin           Build IDA plugin [default: disabled]
  --ida-headless str             Run IDA in a headless mode
  --ida-path str                 A directory with IDA Pro
  --ida-plugin-path str          A directory for IDA plugins
```

Basically, since `oasis/ida` is not loaded, it doesn't see the `Flag ida` line in it, due to which it doesn't recognize `--disable-ida` as a known option

However, without such changes, we don't really have the ability to disable things due to lines of the type `flag(everything) || flag(ida)` in `oasis/ida` for example.

Hence, we will have to come up with a different solution. However, I will leave this PR open and will build upon the solution here :)

jaybosamiya:
With the changes in the latest commit, we filter out the flags when passing to `setup.ml` thereby ensuring that things do not face the problem described above. This works without problems on my system (i.e. the `./configure` analogous to the one that Travis runs, works perfectly on my system, but this breaks Travis for some reason). This will require further analysis.

jaybosamiya:
After turning off the `-quiet` and testing on separate branch (with commit history [here](https://github.com/jaybosamiya/bap/compare/de063e57f5903f0a...31c3585ab2bd4b)), I was able to isolate the issue, and it was due to a
```
`which clang++`
```
which was not being expanded due to some reason. Also, `fsi-benchmark` which depends on `ida` was being built even when `ida` was disabled

With the newly added commits, both above issues are fixed.

However, please do NOT merge this until I am done testing, with ida enabled on my system, as well, and I give the green signal. Thanks :smiley:

jaybosamiya:
As I suspected, we get issues due to the `--objdump-targets=[...; ...]` that will run on our systems (but not on Travis). We should merge only after coming up with a solution to this (and for the future, think about how (if possible) we can test better on Travis)

ivg:
So, we're postponing this PR until better times.  I will close it so far, to prevent accidental merging :)

PR #467, opened 2016-06-10
Use BinaryAnalysisPlatform/bap-ida-python for IDA integration

jaybosamiya:
\<EDIT\>

Note: Due to #468, this PR no longer needs to implement runtime configuration by itself, and hence that portion has been removed from this. The following description is left untouched for history reasons. However, `~/.bap/config` is no longer used. Also, `conf-ida` is still depended on. The only changes that are in this PR now are:

+ Removal of all IDA Python plugin code
+ Targeting the extremely short snippet codes to the new paths due to the https://github.com/BinaryAnalysisPlatform/bap-ida-python/pull/1 PR.

\</EDIT\>

---

This PR basically removes almost all of the the Python code for IDA integration from this repo, since it is moved to BinaryAnalysisPlatform/bap-ida-python via [this PR](https://github.com/BinaryAnalysisPlatform/bap-ida-python/pull/1).

Also, this PR uses the configuration file `~/.bap/config` to know where IDA is at runtime, rather than at compilation time. This ensures that we are able to stay robust to updating IDA, weird IDA install paths, etc. Thereby, we no longer need to depend on `conf-ida` either, and this PR removes that dependency as well. Even if IDA is not on the system, and our ida plugin is installed, it is no longer an issue since the configuration file takes care of this at runtime.

Testing has been done on an Ubuntu 14.04 with IDA-6.6 installed, with IDA-6.9 installed, and on VMs with Ubuntu 14.04 with IDA-Demo-6.9 installed, and with no IDA. All work as expected.


jaybosamiya:
**Note:** The latest commit is due to the namespace changes in https://github.com/jaybosamiya/bap-ida-python/commit/b3bd5bd3a99054b853575022eb248244dce0e756

PR #468, opened 2016-06-14
Allow runtime configuration of plugins

jaybosamiya:
\<EDIT\>

**NOTE:** Due to a lot of discussion that can be seen below, the following description of the PR does not reflect all the changes, but is kept here for history purposes. Please read the documentation in `bap.mli` [here](https://github.com/BinaryAnalysisPlatform/bap/pull/468/files) to see how to use the `Config`.

\</EDIT\>

---

### Background

Previously, one could access compilation time configuration information in plugins through `Config` (for example, `Config.version` used commonly in many plugins). However, this would not allow for runtime configuration.

### Current PR

This PR adds the ability for plugins to
+ `Config.get name` -- get the value assigned to `name`
+ `Config.set name data` -- assign `name` to value `data`
+ `Config.options ()` -- get a list of all options that are currently set for the plugin

Each plugin stores its options in its own subdirectory within a configuration directory that is localized to that specific installation of bap. This allows users to have multiple installations of bap with many different plugins using same/similar names, and still not have conflicts.

Also, the configuration values can be overridden for some runs using environment variables. For example, consider a plugin `P`. If the configuration file for `P` has a variable `K` set to `V1`, but the user executes bap using `BAP_P_K=V2 bap`, then `Config.get "K"` returns `V2`. This allows users to modify configuration for single runs if they wish to.

### Design decisions

+ The configuration format is chosen to be a key-value dictionary with both key and value being strings. This allows users to come up with their own serialization format if they want/need to.
+ The configuration storage format is abstracted away from the user (and the developer of the plugin)
    + This allows us to change it later if we wish to, without upsetting users
+ Plugin developer can get a list of _all_ set values for configuration as a `(string*string) list`, which can be used for caching etc (i.e. if configuration changes, usually it means that the cached value should not be used).
+ Each plugin stores its configuration separately
    + This means that no plugin unwittingly clobbers the configuration of another
    + It also allows users to targettedly configure plugins
+ Configuration variable names are case insensitive. This is due to the logical reason configuration variable names should be descriptive, and not confusing due to different capitalizations having different meaning.
+ Internally, the configuration (currently) is stored as a bunch of files on the `confdir` directory. Each plugin stores in a flat text file `PLUGINNAME/config`, and each line corresponds to one key-value pair (stored in the format `KEY=VALUE`).
+ Environment variable overriding - Using the format `BAP_PLUGINNAME_CONFIGVARNAME=CONFIGVALUE`, one can override the corresponding config variable.


maurer:
What happens when variables contain an _ (probably a common case)?

What about an =? (probably less important)

Might it make sense to be able to `Config.get_global` (but not `Config.set_global`) to access global bap configuration? (For example, if your plugin wants to get the location of the bap cache directory to store its own cache, it might want to be able to do this.)

Why are we using VAR=DATA, one per line, as opposed to `.ini`, `.yaml` `.toml` or similar?
This is mostly a question not because VAR=DATA is fundamentally wrong, but ignores questions of quoting, data representation, namespacing, etc. that people have already explored through various other configuration formats.

ivg:
> What happens when variables contain an _ (probably a common case)?

Don't see any problems with the underscore. We don't split on it. 

> What about an =? (probably less important)

It will not work, yeah.

> Might it make sense to be able to Config.get_global (but not Config.set_global) to access global bap configuration?

It is already available. Since global configuration is constant, it is revealed as a module level value, that is constant by definition, e.g., `Config.prefix`, etc. What concerning the cache dir, then it is hidden, by a cache providing service. If someone needs to use cache, he just needs to say `Cache.store` and it will be automatically stored in the cache directory/server/whatever the services uses to store caches.

> Why are we using VAR=DATA, one per line, as opposed to .ini, .yaml .toml or similar?

All libraries that we found in opam didn't satisfy our requirements. Either they provide to much, like json and xml based storage formats, that imply their own hierarchies (we don't need one, as we use filesystem hierarchy), or they just sucks, or introduce unwanted dependencies like `camlp4`. Also, using simple `VAR=DATA` format makes it easy to automate initial parameters via the `configure` script.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+
+    exception Improper_format of string
+
+    let get_env_options () =
+      let prefix = "BAP_" ^ String.uppercase plugin_name ^ "_" in
+      let prefix_chop_key (k, v) =
+        match String.chop_prefix ~prefix k with
+        | Some k -> Some (k, v)
+        | None -> None
+      in
```
ivg: coding convention: don't put `in` on a separate line.

maurer:
@ivg The _ problem happens with environment variables if the plugin name and variable name both have _s, since it is split on there.

My main worry with VAR=DATA is that we're going to run into quoting/encoding/escaping problems (e.g. what happens when a string is recorded with a newline, a space, a unicode character, etc.) and using existing file drivers is one way to solve that.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+    let plugin_name = name
+    include Bap_config
+
+    let (/) = Filename.concat
+    let confdir = confdir / plugin_name
+    let conf_filename = confdir / "config"
+
+    exception Improper_format of string
+
+    let get_env_options () =
```
ivg: There is `Sys.getenv` function that is already doing this

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+                                        |> split_filter)
+        with Sys_error _ -> []
+      in
+      get_env_options () |>
+      List.fold ~init:conf_file_options
+        ~f:(fun o (k, v) -> List.Assoc.add o k v)
+
+    let get name =
+      let name = String.uppercase name in
+      let search_for = options () |> List.Assoc.find in
```
ivg: it can be done even shorter :)
```ocaml
let get =  options () |> List.Assoc.find ~equal:String.Caseless.equal 
```

jaybosamiya:
@maurer I guess we can resolve this issue (as we discussed offline) by deciding on a convention that plugin names would NOT have underscores in them. I am not sure if anything other than an update to our `bapbundle`/`bapbuild` tools would be needed for this (along with necessary documentation).

With respect to VAR=DATA way of storing: this is exactly why I have abstracted the storage mechanism from the config itself. If we find it necessary to store using that, all we have to do is update our implementation to the new format. For now, we have kept it using `VAR=DATA` due to its simplicity, in addition to the points that @ivg mentioned

lib/bap/bap_self.ml
jaybosamiya: `Sys.getenv` does not get *all* the ones we want. It only returns value for a specific option.
In case we want to get the list of all config values passed to a plugin, we cannot just use `Sys.getenv`

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+      search_for name
+
+    let set ~name ~data =
+      let name = String.uppercase name in
```
ivg: Since, I suspect that I will still look sometimes at the config files directly, and I'm scared of caps locks, so can you actually stick to the lowercase instead?

lib/bap/bap_self.ml
jaybosamiya: Ah! The eta reduction at play :)

Also, I didn't know about `Caseless`. I will update with this

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+      let old_conf = options () in
+      let new_conf = List.Assoc.add old_conf name data in
```
ivg: `List.Assoc.add` will prepend a new binding to the list of bindings. Suppose someone changed a binding `N` times, that will result in `N` repetitions of the same binding in a file. So we have sort of every growing configuration file. To prevent this, you should `List.Assoc.remove` it before adding.

lib/bap/bap_self.ml
jaybosamiya: Ah! But, if all our comparisons are made caseless, then it doesn't matter how we store it. I will switch to lowercase there :)

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+      let conf_lines = List.map new_conf ~f:(fun (k, v) -> k ^ "=" ^ v) in
+      let write_lines () = Out_channel.write_lines
+          conf_filename conf_lines in
+      try
+        write_lines()
```
ivg: coding convention: `f ()` instead of `f()`

lib/bap/bap_self.ml
jaybosamiya: Well, in my tests, I did not come across this. In fact, it overwrites the value. For example, running:
```
# Config.set "asd" "qwe";;
- : unit = ()
# Config.set "zxc" "123";;
- : unit = ()
# Config.set "asd" "789";;
- : unit = ()
# Config.set "zxc" "bnm";;
- : unit = ()
```
gives the config file as
```
ZXC=bnm
ASD=789
```
as expected.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,73 @@ module Create() = struct
...
+      with Sys_error _ ->
+        let dir_permissions = 0o755 in
+        let makedir_and_write_lines () =
+          Unix.mkdir confdir dir_permissions; write_lines() in
+        try
+          makedir_and_write_lines()
```
ivg: I don't see a big win in creating a one line function, and calling them inside the try/catch block, why not just put them body in it. Moreover, I don't think that using exceptions here for control flow is adding much to readability, given that you're making an assumption that a failed write to a file implies the absence of a parent directory. I think the following is more straightforward:
```ocaml
if not (Sys.file_exists conf_dir) 
then Unix.mkdir conf_dir conf_perm;
if not (Sys.is_directory conf_dir)
then invalid_argf "%s is not a directory";
Out_channel.write_lines filename config
```

This example also demonstrates a slightly different naming convention, that is used in OCaml. We use shorter names, as the scope of the name is usually much more limited, and referential transparency, strong typing and immutability reduces a possible impact of name clashes.

jaybosamiya: Ah well, I was following the principle of "It is faster to ask for forgiveness, than permission", which is used very very often in Python. I will switch to this style though.

ivg:
> by deciding on a convention that plugin names would NOT have underscores in them

Sorry, we can't decide such convention. Plugin names have underscores. But I still don't see any problem with that, we don't split on underscore. When we lookup for a variable XXX_YYY in plugin AAA_BBB, we just concatenate them and look in the environment for the `AAA_BBB_XXX_YYY`. What I am missing?

lib/bap/bap.mli
```diff
@@ -589,6 +589,32 @@ module Std : sig
     val info    : ('a,Format.formatter,unit) format -> 'a
     val warning : ('a,Format.formatter,unit) format -> 'a
     val error   : ('a,Format.formatter,unit) format -> 'a
+
```
ivg: I don't see the removal of the `Config` module from the toplevel namespace?

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,66 @@ module Create() = struct
...
 
+  module Config = struct
+    let plugin_name = name
+    include Bap_config
+
+    let (/) = Filename.concat
+    let confdir = confdir / plugin_name
+    let conf_filename = confdir / "config"
+
+    let get_env_options () =
```
ivg: Again, there is a `Sys.getenv` function. There shouldn't be any `get_env_options` function at all.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,66 @@ module Create() = struct
...
+          conf_filename ~f:(fun ch -> In_channel.input_lines ch
+                                      |> split_filter)
+      with Sys_error _ -> []
+
+    let options () =
+      get_env_options () |>
+      List.fold ~init:(conf_file_options ())
+        ~f:(fun o (k, v) -> List.Assoc.add o k v)
+
+    let get = options () |> List.Assoc.find ~equal:String.Caseless.equal
```
ivg: I think, that get function should be implemented as:
```ocaml
let get name = 
  try Some (Sys.getenv name) with
  Not_found -> options () |> List.Assoc.find ~equal:String.Caseless.equal
```

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,66 @@ module Create() = struct
...
+        | k :: vs -> k, String.concat ~sep:"=" vs
+        | [] -> invalid_arg "empty string" in
+      let split_filter = List.map ~f:string_splitter in
+      try
+        In_channel.with_file
+          conf_filename ~f:(fun ch -> In_channel.input_lines ch
+                                      |> split_filter)
+      with Sys_error _ -> []
+
+    let options () =
```
ivg: First of all, as you just trying to concatenate, then it is better to write this function as:
```ocaml
let options () = get_env_options () @ conf_file_options ()
```
Second, we don't need this function at all. See mine comments below.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,66 @@ module Create() = struct
...
+      List.fold ~init:(conf_file_options ())
+        ~f:(fun o (k, v) -> List.Assoc.add o k v)
+
+    let get = options () |> List.Assoc.find ~equal:String.Caseless.equal
+
+    let set ~name ~data =
+      let name = String.lowercase name in
+      let old_conf = conf_file_options () in
+      let remaining_conf = List.Assoc.remove old_conf name in
+      let new_conf = List.Assoc.add remaining_conf name data in
```
ivg: The OCaml style for this would be:
```ocaml
let conf = conf_file_options () in
let conf = List.Assoc.remove old_conf name in
let conf = List.Assoc.add remaining_conf name data in
```

You shouldn't introduce new names, instead you should kill outdated versions of data as soon is possible. There are many reasons for this:
1. Having all these names in the scope, makes code harder to understand:
1.1. reader may think that you will later use all versions of the variable, including old one, so that's why you gave    different names;
1.2. reader need to start with the initial hypothesis that each name is associated with values of different genera, e.g., different types, intents and so on, otherwise why you gave them different names. 
2. It is error prone, as it so easy to access an outdated version of the data. 
3. It is inefficient.
3.1. You extend a lifetime of a variable more than needed
3.2. You prohibit an optimizations, where a compiler can notice that the old value is not used anywhere and just update it destructively.

lib/bap/bap_self.ml
```diff
@@ -60,4 +60,66 @@ module Create() = struct
...
+      let conf_lines = List.map new_conf ~f:(fun (k, v) -> k ^ "=" ^ v) in
+      let write_lines () = Out_channel.write_lines
+          conf_filename conf_lines in
+      try
```
ivg: I still think that using exceptions for control flow here is unjustified. Use `Sys.is_directory` and `Sys.file_exists` to ensure the precondition.

lib/bap/bap_self.mli
```diff
@@ -10,4 +10,31 @@ module Create() : sig
   val info    : ('a,formatter,unit) format -> 'a
   val warning : ('a,formatter,unit) format -> 'a
   val error   : ('a,formatter,unit) format -> 'a
+
```
ivg: I usually try not to duplicate the comments. As they will start to diverge very soon :)

lib/bap/bap_self.ml
jaybosamiya: This is _not_ just a concatenation. `(fun o (k, v) -> List.Assoc.add o k v)` does an over-writing add. Basically, it ensures that the environment configuration overriding happens. Concatenation would not do this (and would create duplicates, and I could not find any documentation on which order `List.Assoc` would `find` things when there are duplicates)

lib/bap/bap_self.ml
jaybosamiya: In such a scenario (i.e. by implementing it like this), `options ()` cannot return all the bindings.

Also, this _is_ cross platform since, as far as I could tell from reading online, both Unix-style OSs (Linux and Mac) and Windows would return it as a list of strings, with the strings having `=` as the separator.

lib/bap/bap_self.ml
jaybosamiya: This function does _not_ work like `Sys.getenv`. It is more similar to `options ()` but reads from env. Please see my comments below.

lib/bap/bap.mli
jaybosamiya: I'm not sure I understand correctly.

jaybosamiya:
Just summarizing the offline discussion b/w @ivg and me:

We can have a `Param` submodule in `Self` which could be used to create parameters which will be able to automatically read and use (in decreasing order of precedence)
+ Command line parameters (to be implemented later)
+ Environment Variables
+ Configuration File
+ Default Values specified by the plugin

I will be pushing changes soon based upon this design decision.

lib/bap/bap.mli
jaybosamiya: This comment does not apply anymore after the latest changes.

lib/bap/bap_self.mli
jaybosamiya: This has been taken care of now.

jaybosamiya:
BTW, command line handling is now implemented. Also the config file can handle configs like
```
# This line should be ignored as it is a comment
# This line should be ignored too, even though AAA=BBB here
ZXC=bnm
ASD=false
x=123
        # No error thrown for this comment here
    z=222
# No error for z declaration

# Above (empty) line should be ignored (i.e. no error)
this should throw an error though as it doesn't define a key-value pair
```

Some changes are yet to be made, and they have clearly been marked with TODO in the commits

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,94 @@ module Create() = struct
...
+        let (/) = Filename.concat in
+        Bap_config.confdir / plugin_name / "config" in
+      let string_splitter str =
+        let str = String.strip str in
+        match String.split str ~on:'=' with
+        | k :: _ when String.prefix k 1 = "#" -> None
+        | [""] | [] -> None
+        | [k] -> invalid_argf
+                   "Maybe comment out \"%s\" using # in config file?" k ()
+        | k :: vs -> Some (k, String.concat ~sep:"=" vs) in
```
ivg: `sep` should be `""`, or you can just omit it, as by default it is an empty string. Also, I think it is a good idea to `strip` both key and data.

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,94 @@ module Create() = struct
...
+      let value = default in
+      let str = get_from_conf_file name in
+      let str = match get_from_env name with
+        | Some _ as v -> v
+        | None -> str in
+      let parse str =
+        let parse, _ = converter in
+        match parse str with
+        | `Error err
+          -> invalid_argf
```
ivg: do not start  a line with the `->`. The preferred formatting is:
```ocaml
| `Error err ->
    invalid_argf "Couldn't parse %S for parameter %S: %s" 
       str name (Error.to_string err) ()
```
Note the usage of `%S`, that will put comments over a string, and output of an error.

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,94 @@ module Create() = struct
...
+               "Could not parse \"%s\" for name \"%s\"" str name ()
+        | `Ok v -> v in
+      let value = match str with
+        | Some v -> parse v
+        | None -> value in
+      ref value
+
+    let create converter ~default ?(docv="VAL") ~doc ~name : 'a t =
```
ivg: The name should be `param` or `parameter` to be consistent with the `flag`

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,94 @@ module Create() = struct
...
+      main := Term.(const (fun x () -> param := x) $ t $ (!main));
+      param
+
+    let flag ?docv ~doc ~name : bool t = assert false
+    (* TODO Implement this *)
+
+    let term_info = ref (Term.info ~doc plugin_name)
+
+    type manpage_block = [
+        `I of string * string |
```
ivg: The pipes should be at the start of the line.

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,94 @@ module Create() = struct
...
+
+    let extract () : 'a t -> 'a =
+      match Term.eval (!main, !term_info) with
+      | `Error _ -> exit 1
+      | `Ok _ -> (fun p -> !p)
+      | `Version | `Help -> exit 0
+
+    let bool = Arg.bool
+    let int = Arg.int
+    let string = Arg.string
```
ivg: and all others, please

lib/bap/bap_self.ml
jaybosamiya: Hmm. I agree it is a good idea to `strip` both key and data.

The reason for using the `~sep:"="` in the `String.concat` is that we might have (for example base64 encoded) data where the `=` are significant in the value. Since we split using the `=`, and then join all except the first using the `=`, this keeps the value exactly intact

For example:
```
# let x = String.split "a==b==c==" ~on:'=';;
val x : Core_kernel.Std.String.t list = [a; ; b; ; c; ; ]
# String.concat ~sep:"=" x;;
- : Core_kernel.Std.String.t = a==b==c==
```
Here, `x` is equivalent to `vs`

lib/bap/bap_self.ml
jaybosamiya: I am going to rename all the functions to their proper names once we decide on them. All the current names can be considered placeholders, just to not get confused (as offline, we were switching between too many names, so I thought it'd be better to just keep one standard and then once we are done deciding, we can change in one go) :)

lib/bap/bap_self.ml
ivg: yeah, you're right, the `sep` should be `=`, my mistake.

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,121 @@ module Create() = struct
...
+        let (/) = Filename.concat in
+        Bap_config.confdir / "config" in
+      let string_splitter str =
+        let str = String.strip str in
+        match String.split str ~on:'=' with
+        | k :: _ when String.prefix k 1 = "#" -> None
+        | [""] | [] -> None
+        | [k] -> invalid_argf
+                   "Maybe comment out %S using # in config file?" k ()
+        | k :: vs -> Some (k, String.concat ~sep:"=" vs) in
```
ivg: Hey, where is the strip? ;)

jaybosamiya: Damn, I forgot :P Will fix soon.

lib/bap/bap_self.ml
```diff
@@ -60,4 +61,121 @@ module Create() = struct
...
+        | `Error err ->
+          invalid_argf "Could not parse %S for parameter %S: %s"
+            str name err ()
+        | `Ok v -> v in
+      let value = match str with
+        | Some v -> parse v
+        | None -> value in
+      ref value
+
+    let param converter ~default ?(docv="VAL") ?(doc="UNDOCUMENTED") ~name =
```
ivg: can you lowercase `"UNDOCUMENT"`. We uppercased `VAL` as it is a reference to a variable, that is according to the man page guidelines is uppercased.

PR #471, opened 2016-06-17
BAP to the Future | Config update

jaybosamiya:
I just _had_ to make a reference to that iconic movie here. :zap: :metal: :zap:
This PR is one from a series of PRs that will transition our `Config` interface into being able to work in a much better way. The point is to simplify the interface that plugin developers use for command line parsing. Over time, we aim to consolidate all the information from different plugins and produce a much nicer and more coherent command line parsing structure (compared to the hacky method we had earlier with Cmdliner being used in all plugins).

This PR does the following:
+ Updates the type of `reader` to a record with a `get` function
    + Turns out, the implementation from #468 guaranteed a truly polymorphic `get` function, but the interface did not guarantee it at the time and thus OCaml made it weakly polymorphic. This fixes it.
+ Updates the interface of `Config` to instead use a callback function
    + This is in preparation for further PRs to come which will update the implementation for the transitions outlined above
+ Uses `Bap_future.future` in implementation of `param`
    + This is because actually, parameters are only set once during execution, and thereafter are not updated. Hence, they have a timeline that spans from undefined to defined, and are hence more better understood as `future`s rather than as `ref`s
+ Updated documentation to for newer interface usage

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -589,11 +589,11 @@ module Std : sig
 
           (* ... *)
 
-          let on_config_ready {Config.get=(!)} =
-            do_stuff !path !debug (* ... *)
-
           let main () =
-            Config.request on_config_ready
+            Config.when_ready
+              (fun {Config.get=(!)} ->
+                 do_stuff !path !debug (* ... *)
+              )
         ]}
     *)
     module Config : sig
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -589,11 +589,11 @@ module Std : sig
 
           (* ... *)
 
-          let configured {Config.get=(!)} =
+          let on_config_ready {Config.get=(!)} =
             do_stuff !path !debug (* ... *)
 
           let main () =
-            Config.parse configured
+            Config.request on_config_ready
         ]}
     *)
     module Config : sig
@@ -637,9 +637,9 @@ module Std : sig
       (** A witness that can read configured params *)
       type reader = {get : 'a. 'a param -> 'a}
 
-      (** [parse configured] parses configuration and command line
-          arguments and calls [configured (reader:Config.reader)] *)
-      val parse : (reader -> unit) -> unit
+      (** [request on_config_ready] parses configuration and command line
+          arguments and calls [on_config_ready (reader:Config.reader)] *)
+      val request : (reader -> unit) -> unit
 
       (** The type for a block of man page text.
 
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -640,9 +640,11 @@ module Std : sig
       (** A witness that can read configured params *)
       type reader = {get : 'a. 'a param -> 'a}
 
-      (** [request on_config_ready] parses configuration and command line
-          arguments and calls [on_config_ready (reader:Config.reader)] *)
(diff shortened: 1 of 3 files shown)


lib/bap/bap.mli
```diff
@@ -632,11 +634,12 @@ module Std : sig
...
-      (** Reads a value from a parameter *)
-      type 'a reader = 'a param -> 'a
+      (** A witness that can read configured params *)
+      type reader = {get : 'a. 'a param -> 'a}
 
-      (** Parse command line arguments and return a param reader *)
-      val parse : unit -> 'a reader
+      (** [parse configured] parses configuration and command line
+          arguments and calls [configured (reader:Config.reader)] *)
+      val parse : (reader -> unit) -> unit
```
ivg: just for the record, ideas for the name `Config.{require,request}`

ivg: And the comment should look something like this:
```ocaml
(** [request f] request the system to call function [f] once configuration parameters are established and stabilized.  An access function will be passed to the function [f], that can be used to safely dereference parameters. An example usage is:
     {[
          let output = Config.(param file ~default:"output.txt" "output")
          let value = Config.(param string ~default:"hello, world" "value")

          let main {Config.get=(!)} = 
            Out_channel.write_all !output ~data:!value 

          let () = Config.request main 
     ]}
*)
val request : (reader -> unit) -> unit
```

lib/bap/bap_self.ml
```diff
@@ -151,11 +156,11 @@ module Create() = struct
     let manpage (man:manpage_block list) : unit =
       term_info := Term.info ~doc ~man plugin_name
 
-    type 'a reader = 'a param -> 'a
-    let parse () : 'a reader =
+    type reader = {get : 'a. 'a param -> 'a}
+    let parse configured : unit =
```
ivg: We should call provided function only after all top level expressions are evaluated, i.e., _after_ the plugin is loaded. 
What we can try to achieve this is to monitor the plugins subsystem events:
```ocaml
let request on_config_ready = 
  Stream.watch Plugins.events (fun subscription -> function
    | `Loaded p when Plugin.name p = plugin_name ->
         on_config_ready {get = (fun p -> Future.peek_exn p)};
         Stream.unsubscribe subscription
    | `Errored name when plugin_name = name -> 
         Stream.unsubscribe subscription)
```

This is not the ideal solution, as it is a little bit fragile, and not very clean, but as an intermediate solution, before we will switch to the global parser, may work.

lib/bap/bap.mli
```diff
@@ -632,11 +634,12 @@ module Std : sig
       val flag :
         ?docv:string -> ?doc:string -> name:string -> bool param
 
-      (** Reads a value from a parameter *)
-      type 'a reader = 'a param -> 'a
+      (** A witness that can read configured params *)
```
ivg: can you also add function:
```
val determined : 'a param -> 'a future
```

This will allow a user to build more complex computations based on parameters, if he likes monadic interface better.

ivg:
So, overall I'm very happy with this PR, there are only two minor concerns.

1. `Config.parse` -- rename and delay the invocation of the callback
2. Add a mapping to the `future` to unleash the power of the future.

PR #472, opened 2016-06-17
Update opam to support brew

rvantonder:
This is what I believe to be the right settings for building bap with brew. tl;dr, `brew install homebrew/versions/llvm34` worked for me. The other packages are available through brew as is. Didn't test on a fresh install though.


ivg:
This changes should be propagated to the [opam-repository](https://github.com/BinaryAnalysisPlatform/opam-repository), as `opam/opam` is used mostly only for travis and debugging.

ivg:
and don't forget to update the branch, otherwise I will be unable to merge it

PR #473, opened 2016-06-17
Update emit-ida-script to use Config instead of Cmdliner + Minor changes in Config

jaybosamiya:
This PR does the following
+ Updates `emit-ida-script` to use the `Config` interface
+ Updates the `Config` interface to
    + Allow `param_all` for taking lists through repetition at command line
    + Adds the standard `some` converter
    + Makes `name` a non named-argument in `param`, `flag` to prevent dangling optional argument problem
+ Update `Config` implementation to
    + Account for above changes
    + Ensure that the filtered arguments are passed to a plugin (Note: In future PRs, as we slowly combine information from different plugin config, we will instead be calling `Term.eval` only once globally, rather than once every plugin.)

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -629,6 +629,10 @@ module Std : sig
         'a converter -> default:'a ->
         ?docv:string -> ?doc:string -> name:string -> 'a param
 
+      val param_all :
+        'a converter -> default:'a list ->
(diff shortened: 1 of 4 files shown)


jaybosamiya:
Closing this PR since the next PR does all of the above and more :)

PR #474, opened 2016-06-17
Update multiple plugins to use the Config interface instead of Cmdliner

jaybosamiya:
This PR does the following
+ Updates multiple plugins to use the `Config` interface
    + `cache`
    + `callsites`
    + `dump_symbols`
    + `emit_ida_script`
    + `ida`
    + `objdump`
    + `propagate_taint`
    + `read_symbols`
    + `taint`
    + `trace`
    + `warn_unused`
+ Updates the `Config` interface to
    + Allow `param_all` for taking lists through repetition at command line
    + Adds the standard `some` converter
    + Makes `name` a non named-argument in `param`, `flag` to prevent dangling optional argument problem
+ Update `Config` implementation to
    + Account for above changes
    + Ensure that the filtered arguments are passed to a plugin (Note: In future PRs, as we slowly combine information from different plugin config, we will instead be calling `Term.eval` only once globally, rather than once every plugin.)

### DIFF
--- a/plugins/ida/ida_main.ml
+++ b/plugins/ida/ida_main.ml
@@ -135,17 +135,29 @@ let main () =
   register_source (module Reconstructor);
   Project.Input.register_loader name loader
 
-let () =
-  let () = Config.manpage [
-      `S "DESCRIPTION";
-      `P "This plugin provides rooter, symbolizer and reconstuctor services.";
-      `P "If IDA instance is found on the machine, or specified by a
-        user, it will be queried for the specified information."
-    ] in
-  let path = Config.(param (some string) "path" ~default:None
-                       ~doc:"Use IDA to extract symbols from file. \
-                             You can optionally provide path to IDA \
-                             executable, or executable name.") in
-  Config.when_ready (fun {Config.get=(!)} ->
-      (* TODO: Use [path] when calling IDA. *)
-      main () )
+module Main = struct
+  open Cmdliner
+  let man = [
+    `S "DESCRIPTION";
+    `P "This plugin provides rooter, symbolizer and reconstuctor services.";
+    `P "If IDA instance is found on the machine, or specified by a
+  user, it will be queried for the specified information."
+
+  ]
+
+
+  let path : string option Term.t =
+    let doc = "Use IDA to extract symbols from file. \
+               You can optionally provide path to IDA executable,\
+               or executable name." in
+    Arg.(value & opt (some string) None & info ["path"] ~doc)
+
+  let info = Term.info name ~version ~doc ~man
+
+  let () =
+    let run = Term.(const main $ const ()) in
+    match Term.eval ~argv ~catch:false (run,info) with
+    | `Ok () -> ()
+    | `Help | `Version -> exit 0
+    | `Error _ -> exit 1
+end
--- a/plugins/ida/ida_main.ml
+++ b/plugins/ida/ida_main.ml
@@ -135,29 +135,17 @@ let main () =
   register_source (module Reconstructor);
   Project.Input.register_loader name loader
 
-module Main = struct
-  open Cmdliner
-  let man = [
-    `S "DESCRIPTION";
-    `P "This plugin provides rooter, symbolizer and reconstuctor services.";
-    `P "If IDA instance is found on the machine, or specified by a
-  user, it will be queried for the specified information."
-
-  ]
-
-
-  let path : string option Term.t =
-    let doc = "Use IDA to extract symbols from file. \
-               You can optionally provide path to IDA executable,\
-               or executable name." in
-    Arg.(value & opt (some string) None & info ["path"] ~doc)
-
-  let info = Term.info name ~version ~doc ~man
-
-  let () =
-    let run = Term.(const main $ const ()) in
-    match Term.eval ~argv ~catch:false (run,info) with
-    | `Ok () -> ()
-    | `Help | `Version -> exit 0
-    | `Error _ -> exit 1
-end
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "This plugin provides rooter, symbolizer and reconstuctor services.";
+      `P "If IDA instance is found on the machine, or specified by a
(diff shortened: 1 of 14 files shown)


jaybosamiya:
**Note:** Some plugins still need to be updated from `Cmdliner` to `Config` and this will be done in a future PR

Just for reference, these plugins require changes in
```
plugins/byteweight/byteweight_main.ml
plugins/demangle/demangle_main.ml
plugins/llvm/llvm_main.ml
plugins/map_terms/map_terms_main.ml
plugins/phoenix/phoenix_main.ml
plugins/print/print_main.ml
plugins/x86/x86_main.ml
```

plugins/ida/ida_main.ml
```diff
@@ -135,29 +135,17 @@ let main () =
...
+      `P "This plugin provides rooter, symbolizer and reconstuctor services.";
+      `P "If IDA instance is found on the machine, or specified by a
+        user, it will be queried for the specified information."
+    ] in
+  let path = Config.(param (some string) "path" ~default:None
+                       ~doc:"Use IDA to extract symbols from file. \
+                             You can optionally provide path to IDA \
+                             executable, or executable name.") in
+  Config.when_ready (fun {Config.get=(!)} ->
+      (* TODO: Use [path] when calling IDA. *)
```
ivg: Such code shouldn't be in the master branch. We shouldn't merge a code the definitely doesn't work. It is better to make a separate PR where path is implemented.

plugins/objdump/objdump_main.ml
```diff
@@ -77,8 +76,8 @@ let main () =
   Stream.merge Project.Info.arch Project.Info.file ~f:run_objdump |>
   Symbolizer.Factory.register name
 
-let info =
-  let man = [
+let () =
+  Config.manpage [
```
ivg: as I can see we forgot to move `manpage` to the global namespace

plugins/trace/trace_main.ml
```diff
@@ -55,55 +69,34 @@ module Cmdline = struct
...
+            exitf "Protocol error: %a" Error.pp err
+          | `System_error err ->
+            exitf "System error: %s" @@ Unix.error_message err
+          | `No_provider ->
+            exitf "No provider for the given URI"
+          | `Ambiguous_uri ->
+            exitf "More than one provider for a given URI"
+          | exception Incompatibe_args ->
+            exitf "Incompatible arguments, see usage SYNOPSIS" )
+
```
ivg: line 108 in the original version cannot be ignored. We should evaluate arguments only after all plugins are loaded.

plugins/warn_unused/warn_unused_main.ml
```diff
@@ -118,31 +118,33 @@ module Cmdline = struct
...
   let passes = [name; "--taint"; "--mark"; "--print"]
   let not_a_pass s = not (List.mem passes s)
 
   let pass name =
     let doc = sprintf "run $mname-%s pass" name in
-    Arg.(value & flag & info [name] ~doc)
-  let taint = pass "taint"
-  let print = pass "print"
-  let mark  = pass "mark"
+    Config.(flag name ~doc)
```
ivg: Do not rename passes, and do not resolve them. Just substitute `ignore_args` with `ignore` and it will work out of box.

plugins/warn_unused/warn_unused_main.ml
```diff
@@ -118,31 +118,33 @@ module Cmdline = struct
...
-    | _ -> match Array.find argv ~f:not_a_pass with
+  let () =
+    let check_passes () = match Array.find argv ~f:not_a_pass with
       | None -> ()
-      | Some pass -> eprintf "Unknown pass: %s" pass; exit 1
-end
+      | Some pass -> eprintf "Unknown pass: %s" pass; exit 1 in
+    Config.manpage man;
+    Config.when_ready (fun {Config.get=(!)} ->
+        check_passes ();
```
ivg: I think `check_passes` is no longer needed.

plugins/warn_unused/warn_unused_main.ml
jaybosamiya: By not renaming them, it overshadows the `taint`, `print`, `mark` needed for `register_pass` etc just after this.

Did the `ignore` substitution though.

jaybosamiya:
`plugins/ida/ida_main.ml` has been moved out of this PR.

plugins/objdump/objdump_main.ml
jaybosamiya: This was still under discussion at the time, and we hadn't decided in either direction. I have not changed it in this PR yet, since we are still uncertain about it (as per our offline discussion)

plugins/warn_unused/warn_unused_main.ml
jaybosamiya: It was the artifact of a previous implementation that got carried over to this. I have removed it.

plugins/trace/trace_main.ml
```diff
@@ -83,20 +83,21 @@ module Cmdline = struct
...
-            exitf "Protocol error: %a" Error.pp err
-          | `System_error err ->
-            exitf "System error: %s" @@ Unix.error_message err
-          | `No_provider ->
-            exitf "No provider for the given URI"
-          | `Ambiguous_uri ->
-            exitf "More than one provider for a given URI"
-          | exception Incompatibe_args ->
-            exitf "Incompatible arguments, see usage SYNOPSIS" )
+    Future.upon Plugins.loaded (fun () ->
```
ivg: This will not work in general... `Plugins.loaded` event will happen after all plugins are loaded, and at this time the commandline parsing may already occur, so it is too late to ask for config. Basically, we should enforce, that at the load time, each config must request the configuration (and tell us what it is looking for). 

So, the order here should be the opposite `Config.when_ready (fun {Config.get=(!)} -> Future.upon Plugins.loaded...`

Also, this example shows, that we need to expose `Config.ready : reader future` interface, for those who would like to use the future interface for building more complex evaluation strategies.

PR #485, opened 2016-06-22
Allow IDA path to be specified at runtime

jaybosamiya:
The PR does the following:
+ Modified the IDA library interface to allow for optionally overriding compile time defaults for `ida_path` and `is_headless`.
+ Allow user to specify `--ida-path` or `--ida-is-headless` on command line to override defaults
+ Allow for `BAP_IDA_PATH` and `BAP_IDA_IS_HEADLESS` environment variables (through the new `Config` interface).

**Note:** For `is_headless`, instead of taking it as a flag, it is taken as an optional boolean. This is because the user might want to override the compile time default in either of the two directions, whereas a flag would only allow "`false` changed to `true`" direction of overriding.


plugins/ida/ida_main.ml
```diff
@@ -129,35 +129,26 @@ let loader path =
...
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "This plugin provides rooter, symbolizer and reconstuctor services.";
+      `P "If IDA instance is found on the machine, or specified by a
+        user, it will be queried for the specified information."
+    ] in
+  let path =
+    let doc = "Path to IDA directory." in
+    Config.(param (some string) "path" ~default:None ~doc) in
```
ivg: it should default to `Ida_config.path`

jaybosamiya: `None` causes it to default to `Ida_config.path` anyways. Basically, if we do not keep compile time constants baked into the library, then it causes a large amount of complication for plugin authors who might want to use just the library, and not the `ida` plugin. I have implemented this on a separate branch if you want to take a look: https://github.com/BinaryAnalysisPlatform/bap/compare/master...jaybosamiya:ida-path

jaybosamiya:
Just an update: this PR should be kept on hold until more changes are made, based upon an offline discussion by @ivg and I.

jaybosamiya:
The latest changes do the following:

+ Use IDA plugin to provide service, not the library
+ Library accepts the provided service, and lets it be used by all plugins
+ Update the `Command` interface to
    + Rename `process` to `parser`
    + Allow functions to get fields from the command

Basically, all of the above added up means this:
+ Specifying `--ida-path=SOMEPLACE` would mean that all plugins get IDA connectivity from `SOMEPLACE`
+ Config file for IDA is now localized to one spot
+ If in the future, we decide to provide for IDA differently (for example, over the network), then the library remains the same, we just add a new provider for the service that implements `create`, `exec`, `close`.

jaybosamiya:
The latest commit makes the check for the path for IDA much earlier (just after configuration ready, before trying to register the service). What this means is that the service will be registered and usable only if the path is fine.

Additionally, this commit also checks that the user has `bap-ida-python` installed in the IDA he/she is trying to point to.

plugins/ida/bap_ida_service.ml
```diff
@@ -0,0 +1,154 @@
+open Core_kernel.Std
```
ivg: did you use `git mv` to migrate this code from `bap_ida.ml` to `bap_ida_service.ml`?

jaybosamiya: Initially, yes, but after that, so many changes were made, that overall, git does not recognize it as a movement of file anymore (basically, it's heuristic is not working anymore). See here: https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F

ivg:
Ok, now I'm really happy with this. Merging!

PR #486, opened 2016-06-22
Support for builtin integer types and proper tags

ivg:
Builtin integers
----------------

This PR adds a support for builtin integral types. This types must be
added on the parser layer, since C is not a context-free language (as
well as many other languages). The following types are automatically
added to the typing context (given that they are supported by the chosen
ABI):

  - [u]int8_t
  - [u]int16_t
  - [u]int32_t
  - [u]int64_t

Implementation notes -- since concrete type representation is defined by
particular ABI we can't specify them in some header file. (Otherwise
this header will no be portable). We also don't want to accidentally
capture this type definitions from the host system, so we can't use
preprocessing (as the preprocessor will fill in this type from its
compiler).

Limitations: currently we do not provide `size_t` or `ssize_t` and some
other types, that are also implementation specific. It looks like, that
we need to extend our ABI interface for this. We're currently working on
the subject.

Proper tag namespaces
---------------------

Since new structure, union or enum definition creates a new namespace
for it fields, we will now extend a field representation from a string
to a pair of strings. A pair is enough since this namespaces cannot be
nested.

Limitation: the anonymous names are not supported to a full extent. The
will just have an empty tag name (the namespace). That will make two
different fields nominally equal if they were defined in the same
scope. A proper way would be to generate unique tags for the anonymous
tags.

The PR also fixes enum representation, now it has a list of fields,
instead of just a number, that abstracts the number of fields.


lib/bap_ida/bap_ida.mli
```diff
@@ -26,7 +26,7 @@ module Std : sig
 
     (** [create target] create an IDA instance that will work with
         [target] executable. *)
-    val create : string -> t
```
jaybosamiya: Seems like you committed this file's changes by mistake. We were using this file for prototyping earlier.

ivg: Bgg... I was sure that I reverted it! Anyway, I the PR is still very weak, made it for the discussion purpose. There are still issues that we must resolve. Btw I was unable to introduce builtin types on a lexer level, so some work still should be done on the ida side.

PR #487, opened 2016-06-23
Update FSI Benchmark

jaybosamiya:
This PR updates FSI Benchmark to use `rooter`s rather than call IDA directly through `Bap_ida`. This now would allow us to do more comparisons as well, as and when we add more rooters.


src/fsi_benchmark_main.ml
```diff
@@ -28,8 +29,8 @@ let bin : string Term.t =
 let tool : string Term.t =
   let doc = "The tool of the function start result that we are going to
   evaluate. If user wants to evaluate bap-byteweight, one can use
-  \"bap-byteweight\". If user want to evaluate IDA, one can use
-  \"idaq\", \"idal\", \"idaq64\" or the specific path of IDA." in
+  \"byteweight\". If user want to evaluate IDA, one can use
+  \"ida\"." in
```
ivg: The default argument is `bap-byteweight`, which is wrong.

src/fsi_benchmark_main.ml
```diff
@@ -77,35 +78,42 @@ let string_of_metric = function
...
-     {false_positive;false_negative;true_positive;prec;recl;f_05} in
-   print std_formatter tool_name result print_metrics) |> function
-  | Ok _ -> ()
-  | Error err ->
-    printf "Function start is not recognized properly due to
-  the following error:\n %a" Error.pp err
+  let rooters = Rooter.Factory.list () in
+  let valid_tool = List.mem rooters tool_name in
+  if not valid_tool
+  then printf "Error: Cannot use tool %S.\n" tool_name
```
ivg: A couple of issues here:

1. It is better to output diagnostic information into `stderr`, not `stdout`; 
2. Even better is to use `error|warning|...` functions for the diagnostics;
3. If it is possible, the user interaction aspect should be decoupled with the business logic aspect, 
    that can be achieved by throwing an exception on the check failure, and capturing in a place that 
    is closer to a user and that is responsible for the interacting with the user.
4. Even better is to check inputs at the place, where they are introduced to the system, and write functions, that has preconditions that shouldn't bother themselves with checking the precondition.

jaybosamiya:
Issues mentioned in comments are taken care of. Also, now the tool allows all possible rooters to be used.

jaybosamiya:
Travis is failing right now due to an outage in launchpad which gives us ocaml. When that is back, re-running travis should give us the green signal :)

Issue #490, opened 2016-06-27
Rooter issues


jaybosamiya:
Without `--rooter=ida`, due to internal rooter on a `.ko` file, we get

```
Failed to create a project: expected structural binary, got raw
```
when we run `bap --loader=ida --symbolizer=ida usbcore.ko -d`

ivg:
So, it is easier to fix it by just ignoring (instead of raising an error) on one hand, on another hand, the fix would be inappropriate. If we are called with an explicit option to use `internal` symbolizer, an it fails, then this shouldn't be ignored.

PR #492, opened 2016-06-27
internal symbolizer shouldn't fail

ivg:
When a raw binary is passed, or when a light loader is used internal
rooter and symbolizer are unable to provide any information. This is an
expected behavior for them and should not be treated as an error.


ivg:
fixes #490

PR #493, opened 2016-06-27
Config defaults

jaybosamiya:
Make the defaults a bit more easier to use. Now, defaults are optional for most converters (which actually hold the defaults now).

For example, one can just say `Config.(param string "x")` instead of the older `Config.(param string "x" ~default:"")`


jaybosamiya:
There are issues with this that are not yet rectified. Closing this PR. Will open a new one without those issues. Sorry for the noise.

PR #494, opened 2016-06-27
Config defaults

jaybosamiya:
Make the defaults a bit more easier to use. Now, defaults are optional for most converters (which actually hold the defaults now).

For example, one can just say `Config.(param string "x")` instead of the older `Config.(param string "x" ~default:"")`

Fixed up issues that might have come up in #493

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -527,6 +527,9 @@ module Std : sig
       access to permission information.  *)
 
 
+  (** Converts an 'a to a string *)
+  type 'a printer = Format.formatter -> 'a -> unit
+
 
   (** {1:api BAP API}  *)
 
@@ -617,9 +620,6 @@ module Std : sig
 
       (** Parse a string to an 'a *)
       type 'a parser = string -> [ `Ok of 'a | `Error of string ]
-
-      (** Converts an 'a to a string *)
-      type 'a printer = Format.formatter -> 'a -> unit
 
       (** Type for converting [string] <-> ['a]. Also defines a default
           value for the ['a] type. *)
@@ -776,8 +776,6 @@ module Std : sig
     end
 
   end
-
-  type 'a printer = Format.formatter -> 'a -> unit
 
   (** Signature for integral type.  *)
   module type Integer = sig
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -621,8 +621,11 @@ module Std : sig
       (** Converts an 'a to a string *)
       type 'a printer = Format.formatter -> 'a -> unit
 
-      (** Interconversion between string and 'a type, along with a default *)
-      type 'a converter = 'a parser * 'a printer * 'a
+      (** Type for converting [string] <-> ['a]. Also defines a default
+          value for the ['a] type. *)
+      type 'a converter
+
+      val converter : 'a parser -> 'a printer -> 'a -> 'a converter
 
       (** [param conv ~default ~docv ~doc name] creates a parameter
           which is referred to on the command line, environment
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -701,7 +701,7 @@ module Std : sig
       (** [string] converts values with the identity function. *)
       val string : string converter
 
-      (** [enum l p] converts values such that unambiguous prefixes of
+      (** [enum l] converts values such that unambiguous prefixes of
           string names in [l] map to the corresponding value of type ['a].
 
           {b Warning.} The type ['a] must be comparable with
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -621,12 +621,12 @@ module Std : sig
       (** Converts an 'a to a string *)
       type 'a printer = Format.formatter -> 'a -> unit
 
-      (** Interconversion between string and 'a type *)
-      type 'a converter = 'a parser * 'a printer
+      (** Interconversion between string and 'a type, along with a default *)
+      type 'a converter = 'a parser * 'a printer * 'a
 
       (** Create a parameter *)
       val param :
-        'a converter -> default:'a ->
+        'a converter -> ?default:'a ->
         ?docv:string -> ?doc:string -> string -> 'a param
 
       (** Create a parameter which accepts a list at command line by
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -624,7 +624,19 @@ module Std : sig
       (** Interconversion between string and 'a type, along with a default *)
       type 'a converter = 'a parser * 'a printer * 'a
 
-      (** Create a parameter *)
+      (** [param conv ~default ~docv ~doc name] creates a parameter
+          which is referred to on the command line, environment
+          variable, and config file using the value of [name], with
+          the type defined by [conv], using the [default] value if
+          unspecified by user.
+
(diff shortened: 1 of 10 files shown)


lib/bap/bap.mli
```diff
@@ -621,12 +621,12 @@ module Std : sig
       (** Converts an 'a to a string *)
       type 'a printer = Format.formatter -> 'a -> unit
 
-      (** Interconversion between string and 'a type *)
-      type 'a converter = 'a parser * 'a printer
+      (** Interconversion between string and 'a type, along with a default *)
+      type 'a converter = 'a parser * 'a printer * 'a
 
       (** Create a parameter *)
```
ivg: Shouldn't we actually add more documentation here? At least we should note what is the `docv` and how it can be referenced in the `doc`. And that if the default value is not specified, then a converter's default will be used.

lib/bap/bap.mli
```diff
@@ -621,12 +621,12 @@ module Std : sig
       (** Converts an 'a to a string *)
       type 'a printer = Format.formatter -> 'a -> unit
 
-      (** Interconversion between string and 'a type *)
-      type 'a converter = 'a parser * 'a printer
+      (** Interconversion between string and 'a type, along with a default *)
+      type 'a converter = 'a parser * 'a printer * 'a
```
ivg: Usually we're trying not to use unlabeled tuples of arity greater than two, especially in the interface or on a module scope, preferring to use records instead. The latter also provides a tighter typing discipline, as `(parser,printer,default)` belongs to a vast family of `'a * 'b * 'c` triples, where `{parser;printer;default}` belongs to a specific type `'a converter` that establishes explicit constraints on its fields. For the completeness sake there is another way to establish such constraint, by defining a type (not type alias): `type 'a converter = Converter of 'a parser * 'a printer * 'a`. With this definition `Converter (parser,printer,default)` also defines a value of type converter. 

Also, originally we were trying to make our converters type compatible with Cmdliner's converters, but since we start to diverge it is better to make this type abstract, and provide constructors.

lib/bap/bap.mli
```diff
@@ -621,12 +621,12 @@ module Std : sig
       (** Converts an 'a to a string *)
       type 'a printer = Format.formatter -> 'a -> unit
```
ivg: In the `bap.mli` there is the same type that is defined just after this module. If you will move it above, then this definition can be removed.

lib/bap/bap.mli
```diff
@@ -701,7 +701,7 @@ module Std : sig
       (** [string] converts values with the identity function. *)
       val string : string converter
 
-      (** [enum l p] converts values such that unambiguous prefixes of
+      (** [enum l] converts values such that unambiguous prefixes of
           string names in [l] map to the corresponding value of type ['a].
 
           {b Warning.} The type ['a] must be comparable with
```
ivg: One of the additional bonuses of making `'a converter` abstract is that now we can easily add new field, `compare` that will default to pervasives compare.

lib/bap/bap_self.ml
```diff
@@ -108,7 +108,13 @@ module Create() = struct
         Some (Sys.getenv name)
       with Not_found -> None
 
-    let get_param ~(converter:'a converter) ~default ~name =
+    let get_conv : 'a converter -> 'a Arg.converter =
```
ivg: We have a convention that if a function has a verb in its name, then it has some side effect, e.g., `fetch_data` vs. `data`. So, for functions that just projects a field `x` we use a noun `x`. This is to emphasize a declarative nature of the specification, and that an operation actually doesn't have any const.

lib/bap/bap_self.ml
```diff
@@ -126,18 +132,25 @@ module Create() = struct
         | None -> value in
       value
 
-    let param converter ~default ?(docv="VAL") ?(doc="Undocumented") name =
+    let param converter ?default ?(docv="VAL") ?(doc="Undocumented") name =
       let future, promise = Future.create () in
+      let default =
+        let _, _, converter_default = converter in
```
ivg: `converter_default` actually rings that we miss some opportunities, provided by a language. `converter.default` would be much better.

lib/bap/bap_self.ml
```diff
@@ -184,26 +197,40 @@ module Create() = struct
...
-    let pair = Arg.pair
-    let t2 = Arg.t2
-    let t3 = Arg.t3
-    let t4 = Arg.t4
-    let some = Arg.some
+    let converter : 'a Arg.converter -> 'a -> 'a converter =
+      fun (parser, printer) default -> parser, printer, default
+
+    let bool = converter Arg.bool false
+    let char = converter Arg.char '\x00'
```
ivg: it is better to use something printable, e.g., `x` for the default. This will also discourage a user from comparing this arbitrary sentinel value to see if it is default or not.

lib/bap/bap_self.ml
```diff
@@ -184,26 +197,40 @@ module Create() = struct
...
+    let int = converter Arg.int 0
+    let nativeint = converter Arg.nativeint Nativeint.zero
+    let int32 = converter Arg.int32 Int32.zero
+    let int64 = converter Arg.int64 Int64.zero
+    let float = converter Arg.float 0.
+    let string = converter Arg.string ""
```
ivg: maybe `"<undefined>"` would be better.

jaybosamiya: Well, the whole point of the defaults is to be able to use it in a simple way, and in my opinion, the most sane default for a string is an empty string `""`.

lib/bap/bap_self.ml
jaybosamiya: I do not think that NUL `'\x00` is an arbitrary sentinel.
The whole point is that there is no clear printable that is "obvious" as a default character; but to most users, the NUL is an obvious default.

ivg: It is `x` that is arbitrary, since there is no specific logic why we chose them. If someone will see that it is zero, then he will try to use:

```ocaml
let var = Config.(param char "var")
...
let x = if !var = '\x00' then create_fresh () else var
...
```

That I would like to discourage in general, as one should use at least option for this. There is another concern, that if this value will slip into the man page, will it be rendered correctly. As it is non-printable special character.

lib/bap/bap_self.ml
ivg: see my comment about char above, it partially applies to the string parameter.

lib/bap/bap_self.ml
jaybosamiya: I do not think that a user would use such a verbose way of doing things, especially when he can just say `~default:(create_fresh ())`. Also, manpage does not have issues with this (I have tested). I think it is better if we discuss this face-to-face.

lib/bap/bap_self.ml
jaybosamiya: For the string parameter, I cannot accept that `"<undefined>"` is better than `""`. Your argument there does not apply here whatsoever. Also, the person can always say `(some string)` if they want to, right? I do not see users taking the insane way to do things, and actually, having `""` is a very good default. Again, maybe it is better we discuss this offline.

lib/bap/bap.mli
jaybosamiya: We decided not to do this (just documenting here), since we will be switching to the new enum thing soon as well.

PR #502, opened 2016-06-29
Allow disabling certain plugins/features when enabling everything

jaybosamiya:
This PR allows us to specifically disable certain plugins/features when we enable everything, for example, if we run `./configure --enable-everything --disable-ida`, then it will enable every plugin/feature except for `ida`

All of the issues that were there in the previous PR that tried to do this (i.e. #454) have been handled here. Expansions, quotes, empty strings, strings with spaces -- all are handled now.

### Tests

All of the following have passed (i.e. work as expected).

#### On personal machine

+ [x] Without parameters
```
./configure
```

+ [X] With lots of paramenters (including `opam config var conf-binutils:targets` which has spaces)
```
./configure --prefix=`opam config var prefix` --with-cxx=`which clang++` --mandir=`opam config var man` --enable-everything --enable-tests --enable-ida --ida-path=`opam config var conf-ida:path` --enable-objdump --objdump-path=`opam config var conf-binutils:objdump` --objdump-targets="`opam config var conf-binutils:targets`"
```

+ [X] Disabling some things (specifically, `objdump`)
```
./configure --prefix=`opam config var prefix` --with-cxx=`which clang++` --mandir=`opam config var man` --enable-everything --enable-tests --enable-ida --ida-path=`opam config var conf-ida:path` --disable-objdump
```

+ [X] Help!
```
./configure --help
```

+ [X] Passing weird parameters (to check if error checking works as expected)
```
./configure --trollol
```

+ [X] Passing weird parameters (to check if error checking works as expected)
```
./configure "--this is SPARTA"
```

#### On Travis CI

+ [X] Disabling `ida`, `fsi-benchmark`, `objdump`
```
./configure "--prefix=/home/travis/.opam/system" "--with-cxx=`which clang++`" "--mandir=/home/travis/.opam/system/man" "--enable-everything" "--enable-tests" "--disable-ida" "--disable-fsi-benchmark" "--disable-objdump"
```

### DIFF
--- a/configure
+++ b/configure
@@ -32,13 +32,15 @@ SETUPS=`ocaml tools/collect.ml setup.ml $SECTIONS`
 AB=`ocaml tools/collect.ml files.ab $SECTIONS`
 
 # Correct for "--disable" type of arguments
-ARGS=`ocaml tools/oasis_sections.ml --args "$@"`
-set --
-while read -r arg; do
-    set -- "$@" "$arg"
-done <<EOF
+if ! $FST; then
+    ARGS=`ocaml tools/oasis_sections.ml --args "$@"`
+    set --
+    while read -r arg; do
+        set -- "$@" "$arg"
+    done <<EOF
     $ARGS
 EOF
+fi
 
 ocaml tools/cat.ml '"\n# $name\n"' -- $SECTIONS $AB _oasis
 ocaml tools/cat.ml '"\n# $name\n"' -- $TAGS _tags.in _tags
--- a/configure
+++ b/configure
@@ -25,11 +25,18 @@ for i in "$@"; do
   esac
 done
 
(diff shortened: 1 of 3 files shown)


jaybosamiya:
Just documenting here for if/when someone wants to figure out what the black magic in the `./configure` script is:

The first part of the script (pre-existing `--a=b` -> `--a b` splitter thing) tells us if it found any parameters whatsoever through the `$FST`. If it did, then we filter through the arguments using the ocaml script into a newline delimited `$ARGS` which is then set into the `$@` through the loop (since that is the only shell variable that acts like an array, rather than a space delimitted string when being used through `"$@"`). Then the `configure` process occurs as usual.

ivg:
damn you're shell guru :bow:

PR #504, opened 2016-06-30
Config plugins v2

jaybosamiya:
This PR updates the remaining plugins to use the `Config` interface.

### DIFF
--- a/plugins/phoenix/phoenix_main.ml
+++ b/plugins/phoenix/phoenix_main.ml
@@ -82,11 +82,13 @@ let () =
   Config.manpage man;
   Config.when_ready (fun {Config.get=(!)} ->
       let cfg_format =
-        let self = !labels_with in
-        let self = if !with_name then `with_name :: self else self in
-        let self = if !with_asm  then `with_asm  :: self else self in
-        let self = if !with_bil  then `with_bil  :: self else self in
-        self in
+        let deprecated_options =
+          [`with_name, !with_name;
+           `with_asm,  !with_asm;
+           `with_bil,  !with_bil] |>
+          List.filter ~f:snd |>
+          List.map ~f:fst in
+        deprecated_options @ !labels_with in
       let options = create
           !output_folder
           cfg_format
--- a/plugins/phoenix/phoenix_main.ml
+++ b/plugins/phoenix/phoenix_main.ml
@@ -16,14 +16,30 @@ let run options project =
   let dest = Phoenix.store () in
   printf "Stored data to %s@." dest
 
-let label_with x y =
+(* Deprecated since, by design choice, this no longer maintains order
+   of input. However, the [--label-with] will maintain order of
+   input. *)
+let dep_label_with x y =
   let name = sprintf "labels-with-%s" x in
   let doc = sprintf "Put %s on graph labels" y in
-  Config.(flag name ~doc)
+  Config.(flag ~deprecated:"Use --label-with=.. instead." name ~doc)
 
-let with_name = label_with "name" "block name"
-let with_asm  = label_with "asm"  "assembler instructions"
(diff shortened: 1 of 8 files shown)


jaybosamiya:
Minor issues:
+ The api plugin's help will now show `--rem` and `--remove` as separate options, since we do not have alias abilities in our `Config` interface. The reason for not allowing aliases was to have a single point of truth at the command line for the plugins. However, the reason for aliasing here is to ensure that people's scripts don't break.
+ The phoenix plugin's `cfg_format` code feels a little weird to me. Might want to look into improving it later on.

plugins/phoenix/phoenix_main.ml
```diff
@@ -17,56 +16,43 @@ let run options project =
...
+  let doc = sprintf "Put %s on graph labels" y in
+  Config.(flag name ~doc)
 
-let output_folder : _ Term.t =
-  let doc = "Output data into the specified folder" in
-  let default = "phoenix" in
-  Arg.(value & opt string default & info ["output-folder"] ~doc)
+let with_name = label_with "name" "block name"
+let with_asm  = label_with "asm"  "assembler instructions"
+let with_bil  = label_with "bil"  "bil instructions"
```
ivg: there is a semantic difference here, in the base the order in which options were specified was preserved, and this allowed us to specify format `name, asm, bil` or `name, bil, asm` etc.

plugins/phoenix/phoenix_main.ml
```diff
@@ -76,10 +62,23 @@ let man = [
...
+  Config.manpage man;
+  Config.when_ready (fun {Config.get=(!)} ->
+      let cfg_format =
+        let self = [] in
+        let self = if !with_name then `with_name :: self else self in
+        let self = if !with_asm  then `with_asm  :: self else self in
+        let self = if !with_bil  then `with_bil  :: self else self in
+        match self with
+        | [] -> [`with_name]
+        | _ as x -> x in
```
ivg: it is not the choice of one from three, it should work as a checkbox basically. Something like this:
```ocaml
let cfg_format = List.filter [`with_name,!with_name; `with_asm, !with_asm; `with_bil, !with_bil] ~f:snd |> List.map ~f:fst
```

plugins/phoenix/phoenix_main.ml
jaybosamiya: Well, with the current `Config` interface then we cannot preserve this, and will have to extend our interface.

jaybosamiya:
As I mentioned in my comment,
> The reason for not allowing aliases was to have a single point of truth at the command line for the plugins

ivg:
> As I mentioned in my comment

yeah, I noticed it, but we still have synonyms. On the other hand, cmdliner accepts any unambiguous prefix of an option, so that means that if you have `--api-remove` then `--api-r` and `--api-re` and `--api-rem` are all implicit synonyms of `--api-remove`. So in this PR you don't need to specify option twice, just removing `rem` variant will not break things.

But, I still think that there is nothing wrong with synonyms, so we can add an optional `string list` flag for it. A user will be discouraged from using it as he needs to write `~synonyms:["a-better-name"]`. So he will use it only if it really have a reason to do this.

Also, back to our offline discussion, I think that the better type for the deprecated option would string. So that a user can write:
```ocaml
Option.(param string "output.txt" ~deprecated:"use our new database interface instead")
```

jaybosamiya:
The following changes have been made now:

+ Merge master (to be able to use features from #505)
+ Use synonyms for `--api-remove` to also accept `--api-rem`
+ Start showing deprecation warnings for `--labels-with-*` and favour `--labels-with=`
+ Cleaned up the code that is used in the deprecation part.

PR #505, opened 2016-07-01
Extending Config

jaybosamiya:
This PR does the following to extend `Config`:

+ Allows plugins to specify that a `config` `param` is deprecated. The reason for adding this is so that if/when developers decide that a certain option is no longer going to be used, then they can mark it as deprecated for a few releases, along with instructions to the alternative that was added, and then once the users have updated their plugins, these options can be phased out smoothly.
    + This addition prints a helpful warning message anytime the user uses this parameter
    + Plugin developers can also customize the message to point towards a better/alternative usage if they want
    + Help page (i.e. the `--help` manpage) clearly shows the parameter as deprecated

+ Allow plugins to specify synonyms for parameters. This, while discouraged, is necessary under some very specific scenarios, which is why we are adding it.

+ Allow plugins to say `as_flag` for a param, which allows usage of the param like a flag as well, specifying defaults.
    + For example, consider `Config.(param (some int) ~as_flag:(Some 10) "x")`. This results in 3 possible command line invocations:
        1. No `--x` - Results in `default` value (specifically here, `None`).
        2. Only `--x` - This causes it to have the value `as_flag` (specifically here, `Some 10`).
        3. `--x=20` - This causes it to have the value from the command line (specifically here, `Some 20`).

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -650,23 +650,31 @@ module Std : sig
           usable as normal, but will emit a warning to the user if they
           try to use it.
           Example usage: [Config.(param ~deprecated int "--old")].
+
+          Additionally, [synonyms] can be added to allow multiple
+          arguments referring to the same parameters. However, this is
+          usually discouraged, and considered proper usage only in rare
+          scenarios.
       *)
       val param :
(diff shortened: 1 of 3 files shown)


ivg:
ha, `as_flag` is a really good catch!

PR #506, opened 2016-07-01
llvm-config-3.8 --system-libs

parkre:
Adds support for `llvm-config-3.5+` flags (namely `--system-libs`)


ivg:
hit the "update branch" button

PR #507, opened 2016-07-01
LLVM 3.4

parkre:
• Support for `llvm-config-3.5+` flags
• Binary smart pointer fix in `plugins/llvm/llvm_binary.hpp`

### DIFF
--- a/plugins/llvm/llvm_binary.hpp
+++ b/plugins/llvm/llvm_binary.hpp
@@ -504,7 +504,8 @@ image* create(const char* data, std::size_t size) {
     OwningPtr<object::Binary> binary;
     if (error_code ec = createBinary(buff, binary))
         llvm_binary_fail(ec);
-    return create_image(std::move(binary));
+	std::unique_ptr<object::Binary> b(std::move(binary));
+    return create_image(binary));
 }
 
 } //namespace img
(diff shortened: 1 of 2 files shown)


plugins/llvm/llvm_binary.hpp
```diff
@@ -18,93 +18,93 @@
...
 namespace llvm { namespace object {
 
-template <typename T>
-content_iterator<T>& operator++(content_iterator<T>& a) {
-    error_code ec;
-    a.increment(ec);
-    if(ec) llvm_binary_fail(ec);
-    return a;
-}
+    template <typename T>
```
ivg: We do not indent inside namespaces. And use 4 whitespaces for indentation. Add the following to yours emacs config, if you want your emacs to produce correct indentation:
```lisp
(setq-default c-basic-offset 4)
(add-hook 'c++-mode-hook
          '(lambda ()
             (require 'auto-complete-c-headers)
             (add-to-list 'ac-sources 'ac-source-c-headers)
             (c-set-style "stroustrup")
             (c-toggle-hungry-state)
             (c-set-offset 'innamespace 0)))
```

ivg:
The first commit is already merged, so make a new PR without it. Also fix the indentation, so I can review only the code changes, not the whitespaces.

plugins/llvm/llvm_binary.hpp
parkre: thanks, i was fixing that now

PR #509, opened 2016-07-04
fix possible use-after-free in llvm_binary plugin

parkre:
*** This pull request is independent of LLVM 3.8 support.

• In `plugins/llvm/llvm_binary.hpp`, there is a `Binary` smart pointer that is improperly handled due to the fact that the pointer to the binary has a shorter lifetime than the image.

• To resolve this issue, we can make the binary smart pointer a part of `image`.

### DIFF
--- a/plugins/llvm/llvm_binary.hpp
+++ b/plugins/llvm/llvm_binary.hpp
@@ -63,8 +63,8 @@ std::vector<MachOObjectFile::LoadCommandInfo> load_commands(const MachOObjectFil
         info = obj->getNextLoadCommandInfo(info);
     }
     return cmds;
-}
-
+}   
+    
 } // namespace utils
 
 namespace seg {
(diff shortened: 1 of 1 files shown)


dbrumley:
As background, this pull request is intended to produce a bap that works with both llvm-3.4 and llvm-3.8, right?

ivg:
The PR message should contain a description of a problem that this PR is solving, why it is introduced and so on. See other PRs for the reference.

plugins/llvm/llvm_binary.hpp
```diff
@@ -430,12 +430,13 @@ uint64_t image_entry(const COFFObjectFile* obj) {
 
 template <typename T>
 struct objectfile_image : image {
-    explicit objectfile_image(const T* obj)
+    explicit objectfile_image(const T* obj, std::unique_ptr<object::Binary> binary)
```
ivg: Both parameters are referring to the same object under different types. The should be only one parameter of the most specific type, passing the same information twice is a bad practice. For example, here the prototype of the constructor should be `explicit objectfile_image(std::unique_ptr<T> binary)`

plugins/llvm/llvm_binary.hpp
```diff
@@ -430,13 +439,14 @@ uint64_t image_entry(const COFFObjectFile* obj) {
...
 struct objectfile_image : image {
-    explicit objectfile_image(const T* obj)
-        : arch_(image_arch(obj))
-        , entry_(image_entry(obj))
-        , segments_(seg::read(obj))
-        , symbols_(sym::read(obj))
-        , sections_(sec::read(obj))
-        {}
+    explicit objectfile_image(std::unique_ptr<object::Binary> binary)
+        : arch_(image_arch(dyn_cast<T>(binary.get())))
```
ivg: This is a trade of that I can't understand: why to call `dyn_cast` five times in the constructor, instead of calling it once at the point where you call the constructor?

Also, calling `get` function on a smart pointer must be avoided if possible. It is possible here.

plugins/llvm/llvm_binary.hpp
```diff
@@ -458,28 +458,29 @@ struct objectfile_image : image {
     std::vector<seg::segment> segments_;
     std::vector<sym::symbol> symbols_;
     std::vector<sec::section> sections_;
-    std::unique_ptr<object::Binary> binary_;
+    std::unique_ptr<T, std::default_delete<object::Binary>> binary_;
```
ivg: There is no  need to specify the default deleter, it is an optional argument, just `std::unique_ptr<T>` will be sufficient

PR #510, opened 2016-07-05
Onward to a single grammar config

jaybosamiya:
This PR does the following things:

+ Makes sure full config names are used at command line. 
This basically ensures that the complete name of argument shows up
everywhere, rather than stripped with plugin name (for example,
`--taint-ptr` shows up now, rather than `--ptr` in the manpage).
+ Allows plugins to register their grammars into the global grammar, along with code to fix up the manpage issues (basically, the previous change would have prevented `--<plugin>-help` from working, but after the changes, it will work even in the single grammar system).
+ Adds a `front_end` method which will be exposed to front ends (with a more clear interface, which abstracts `Cmdliner` away) in order to be able to set up the front end information (i.e. the default `--help` manpage etc).
+ A temporary workaround is added in order to have the plugins working correctly even while front ends are not moved to the new grammar. However, once the front ends are moved to the new grammar (in a future PR), this workaround will be removed.

### DIFF
--- a/lib/bap/bap_self.ml
+++ b/lib/bap/bap_self.ml
@@ -70,6 +70,14 @@ end = struct
         | `Ok () -> Promise.fulfill front_end_promise ()
         | `Version | `Help -> exit 0)
 
+  (* Ensures that all plugins can get their options while the front
+     ends are still being updated. _Must_ be removed after plugins are updated. *)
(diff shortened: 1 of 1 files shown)


jaybosamiya:
Superseded by #512

PR #511, opened 2016-07-12
Please ignore this PR.

jaybosamiya:
Please ignore this PR.

Issue #503, opened 2016-06-29
Discussion: Change to command line arguments


jaybosamiya:
In regards to the updating to the new single-grammar based command line parsing, there are some alternatives we need to consider and decide soon.

Note: I use configuration and command line options almost interchangeably in the discussion below.

Since there exist some plugins which might register pass or not depending on what configuration it has got at runtime, we would be unable to get the complete set of command line options (there is a circular dependency, options depend on registering depends on options). Hence the proposal is to do one of the following:

+ **Peek Arguments** - Peek at the arguments that are known at that point of time, so that plugins are able to receive all configuration parameters, and then are able to register passes, which peeks at the arguments to see "oh, was I called? then lemme run"
    + Advantage(s)
        + User interface remains the same
        + plugin authors write plugins as they've been writing until now
    + Disadvantage(s)
        + Since we are only peeking at items, we can no longer say "unknown option XYZ". This leads to problems for the users who will say "but it should have run the pass" when they've made a typo
        + for plugin developers where due to some mistake in their code, they might end up registering a pass which isn't called and then they would not realize that it is only because of some command line problem

+ **Promise** - Each plugin must, as a promise, provide a `bool Config.param` (i.e. a flag) which would be used as a witness to register a pass (i.e. cannot register a pass that is not promised earlier).
    + Advantage(s)
        + User interface remains the same
        + we can see complete list of promised passes at a single point (man page)
    + Disadvantage(s)
        + Plugin authors are now forced to do this promise-witness thing for each pass
        + it is significantly harder to guarantee that the witness is not given later on, after the whole command line is parsed (i.e. plugin developers can cause system to break quite easily)
        + users will sometimes see a pass in the manpage which might not work "no such pass registered"

+ **New `-p` argument** - Passes can now be called via `-p pass1,pass2,pass3,...` (`--pass` and `--passes` would alias to the `-p`)
    + Advantage(s)
        + No extra work for plugin developers
        + Users now see a more clear interface which separates the passes from plugin options (this will especially help new users, since actually speaking, the order of options doesn't matter, but a new user might think that something like the current `--taint-reg=0x2 --taint --somepass --taint-ptr=0x3 --taint --someotherpass` refers to a special ordering; but with the new system, it would be clear that the internal working is just `--taint-reg=0x2 --taint-ptr=0x3 -p taint,somepass`)
    + Disadvantage(s)
        + Users would need to update their scripts to use the new command line options

[...]

PR #512, opened 2016-07-12
Frontend Config + One grammar system + -p command line option

jaybosamiya:
This PR makes a _lot_ of changes.
- Passes are no longer called using `--PASS`, but instead must be called using `-p PASS`. This concludes the discussion #503.
- The command line grammar is now a single grammar that contains all the possible arguments. This means that when plugins add command line options, these options propagate to the main `bap --help`. Note that `--PLUGIN-help` also works, and lists only arguments _specific_ to that plugin, along with its own manpage.
- `Config` is now moved completely out of the `Self()`. In a future PR, `Self()` will be completely phased out and plugin developers will no longer have to do the "magic" `include Self()` anymore.
- Frontends now use a very similar interface to write command line options as `Config`, but since they would use `Frontend.Config`, they are allowed to also set up special commands (similar to the way `status` and `add` are commands for `git`). This allows a lot of versatality for frontends.
- Frontends are allowed to set up positional arguments as well (such as `FILE` used in `bap`).
- Checks are put in so that plugins cannot mess with frontend, and frontend cannot mess with plugin parameters.
- Update all of our frontends to the new interface.
- Allows frontends to set up "post-checks" which basically run _after_ plugins are loaded, but _before_ frontend is made ready. This allows to catch errors for things like `--rooter` etc very early, even though it depends on the plugins finishing off their registrations.
- Does pre-processing on the command line arguments (but hidden away internally) so that plugin loading is no longer needed to be handled specifically by the frontend (such as the `-l`, `-L` etc. options), but is done in advance, and still works perfectly from the grammar.

Note: Some more minor changes and fixes might need to be done, but I am making this PR so that the review process can start off faster. Also, a refactor of the implemntation of `Config` needs to be done (for minor cleanups) and would be done in a future PR.

Sorry for the noise but this PR is same as #511 but re-opened since Travis could not detect that PR at all.


lib/bap/bap_configuration.ml
```diff
@@ -0,0 +1,829 @@
...
+        printf "  %s %-16s %s@." status (Plugin.name p) (Plugin.desc p));
+    exit 0
+
+  let verbose =
+    let doc = "Print verbose output to log file" in
+    Arg.(value @@ flag @@ info ["verbose"] ~doc)
+
+  type action = Keep | Drop
+
+  let filter_option options prefix =
```
ivg: Why do we still need this function?

ivg:
Some observed bugs:
1. `bap --list-formats` doesn't work as expected:
```
$ bap --list-formats
Config.error> Required argument FILE is missing
```

2. printing format choosing doesn't work any more:
```
$ bap bin/arm-linux-gnueabi-echo -dbir
```

3. Man page for the main frontend doesn't contain sections from manpages of the plugins

lib/bap/bap_configuration.ml
jaybosamiya: Currently, the `filter_option` function exists to get rid of `--no-PLUGIN` type of arguments. If we remove it, then we'll need to add in `--no-taint`, `--no-propagate-taint`, `--no-ida`, etc (i.e. for every plugin). While this would be an easy change to make, it'd cause further bloat to the man page. Are we sure we want to do this? If so, then I can make that change, and remove this completely.

jaybosamiya:
As for `--list-formats`, it is one of those "command" type of params. Now, my initial assumption was the frontends would not have "command" type of params.. One of the ways of doing this is to make a `list` plugin, which basically does this instead. In fact, to me, it makes sense as a plugin rather than a front end thing. The change needed for allowing `--*` based commands which override the `required` positional arguments is such a massively intrusive change that it'd involve changing parts of the `Frontend.Config` interface (at least as per my understanding, though you might be able to suggest better).

Printing format choosing - I'm going to have to look into it again. I am unable to reproduce the bug on my system. I'll try getting on a completely clean build soon, to see if that ends up reproducing the bug.

Manpage - didn't we decide _not_ to put it in the main manpage (since the manpage explanations can be seen with `--PLUGIN-help`?) The change to do this will not be to complicated, but would lead to even more bloat on the frontend manpage.

ivg:
> One of the ways of doing this is to make a list plugin, which basically does this instead. 

One of the reasons why it was made on a frontend side is because, this command must be evaluated only when all plugins are loaded. At the time of adding this option, there were no such mechanism, that why I added this option, that naturally belongs to the `print` plugin. I think it would be indeed a good idea to move this option to a plugin. We should, however decide, whether it should be a `print` plugin, e.g., `--print-formats`. Or a `list` plugin, that will gather all list options, e.g., `--list-formats`, `--list-passes`, etc. An argument against the former, is that formats can be added externally, so the listing operation actually exists on a higher level, then the `print` plugin. An argument against the latter, is that I don't see a clear mechanism for extending it (especially if we will stick to `--list-something` format, instead of `--list=something,else`). Another argument against it, is that we cannot make this interface mandatory, so new plugins may forget to use it, and introduce their own mechanisms. 

> Printing format choosing - I'm going to have to look into it again. I am unable to reproduce the bug on my system. 

It might be because, you have a different order of loading... or you forgot to push some fixes ;)

> Manpage - didn't we decide not to put it in the main manpage 

Did we? Maybe this indeed would be a good idea. As we still have all possible options (so we satisfy the greppable requirement). And we can get an extended information by querying specific plugin for the help, e.g., `--<plugin>-help` (that satisfies structural requirement). So, indeed it might be a good idea.

ivg:
@gitoleg, you're the new owner of this PR, so make sure that it is merged, or at least the necessary parts of it are merged. We still have the configuration system broken, and we need it fixed.

ivg:
Let's admit we won't merge it and it anyway bitrotten beyond repair. We will be inspired with this code when we will release BAP 2.0.

PR #514, opened 2016-07-12
Goodbye segmentation faults caused by PE32+ binaries

parkre:


ivg:
I've ran it on a testsuite and can confirm, that there're no more segfaults. I will extend the testsuite with tests that check that all names and addresses of symbols, segments, sections and entry points are correct, and then will merge it.

ivg:
Ok, so the symbols are on their places, but sections are not (as there is no intersections between sections and symbols). The sections issue is not a regression, so we will fix it in a separate PR (I will create an issue, that describes the problem)

PR #517, opened 2016-07-13
LLVM Plugin Refactor

parkre:
#fix 516

### DIFF
--- a/plugins/llvm/llvm_binary.hpp
+++ b/plugins/llvm/llvm_binary.hpp
@@ -69,8 +69,8 @@ std::vector<MachOObjectFile::LoadCommandInfo> load_commands(const MachOObjectFil
 
 const pe32plus_header* getPE32PlusHeader(const COFFObjectFile& obj) {
     uint64_t cur_ptr = 0;
-    const char * buf = (obj.getData()).data();
-    const uint8_t * start = reinterpret_cast<const uint8_t *>(buf);
+    const char *buf = (obj.getData()).data();
+    const uint8_t *start = reinterpret_cast<const uint8_t *>(buf);
     uint8_t b0 = start[0];
     uint8_t b1 = start[1];
     if (b0 == 0x4d && b1 == 0x5a) { // check if this is a PE/COFF file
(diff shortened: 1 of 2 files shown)


plugins/llvm/llvm_binary.hpp
```diff
@@ -108,52 +108,24 @@ using namespace llvm;
...
-        , offset_(static_cast<uint32_t>(s.PointerToRawData))
-        , addr_(static_cast<uint32_t>(s.VirtualAddress + hdr.ImageBase))
-        , size_(static_cast<uint32_t>(s.SizeOfRawData))
-        , is_readable_(static_cast<uint32_t>(s.Characteristics) &
-                       COFF::IMAGE_SCN_MEM_READ)
-        , is_writable_(static_cast<uint32_t>(s.Characteristics) &
-		       COFF::IMAGE_SCN_MEM_WRITE)
-	, is_executable_(static_cast<uint32_t>(s.Characteristics) &
-			 COFF::IMAGE_SCN_MEM_EXECUTE)
+    segment(std::string name, uint64_t offset, uint64_t addr,
```
ivg: What is the purpose of this constructor? It just passes its arguments to the fields. It is no better than struct initializer. Just make all fields public (and remove trailing underscores from the names), and remove field accessors. The `segment` is just a tuple, there is no functional dependency between fields, so we can't build any abstraction here.

plugins/llvm/llvm_binary.hpp
```diff
@@ -108,52 +108,24 @@ using namespace llvm;
...
-	, addr_(static_cast<uint64_t>(s.VirtualAddress + hdr.ImageBase))
-        , size_(static_cast<uint64_t>(s.SizeOfRawData))
-        , is_readable_(static_cast<uint64_t>(s.Characteristics) &
-                       COFF::IMAGE_SCN_MEM_READ)
-        , is_writable_(static_cast<uint64_t>(s.Characteristics) &
-                       COFF::IMAGE_SCN_MEM_WRITE)
-        , is_executable_(static_cast<uint64_t>(s.Characteristics) &
-                         COFF::IMAGE_SCN_MEM_EXECUTE)
-    {}
+    template <typename S>
```
ivg: Remove this code from the constructor. This is a code that is specific to a particular format. Nothing format specific should be in the `segment` class.

plugins/llvm/llvm_binary.hpp
```diff
@@ -214,47 +182,46 @@ std::vector<segment> read(const MachOObjectFile& obj) {
...
-	const coff_section *s = obj.getCOFFSection(it);
-	uint32_t c = static_cast<uint32_t>(s->Characteristics);
-	if ( c & COFF::IMAGE_SCN_CNT_CODE ||
-	     c & COFF::IMAGE_SCN_CNT_INITIALIZED_DATA ||
-	     c & COFF::IMAGE_SCN_CNT_UNINITIALIZED_DATA )
-	    segments.push_back(segment(*pe32, *s));
-    }
-    return segments;
+template <typename T>
+segment make_segment(T image_base, const coff_section &s) {
```
ivg: you should pass `image_base` by a reference, otherwise it maybe sliced.

plugins/llvm/llvm_binary.hpp
```diff
@@ -214,47 +182,46 @@ std::vector<segment> read(const MachOObjectFile& obj) {
...
+    return segment(s.Name,
```
ivg: Use `base` for `image_base` otherwise it sounds like a base class of image.

plugins/llvm/llvm_binary.hpp
parkre: Done.

plugins/llvm/llvm_binary.hpp
parkre: Done.

ivg:
Build is failing, to build on you local machine with clang use configure option `--with-cxx=clang++` (of course you need to install `clang++`)

ivg:
There are also lots of warnings on your branch. Please, use emacs to compile, this will minimize the chance of a missing warning.

plugins/llvm/llvm_binary.hpp
```diff
@@ -392,17 +295,23 @@ std::vector<symbol> readPE32(const COFFObjectFile& obj) {
...
+        auto addr = sec->VirtualAddress + image_base + sym->Value;
         symbols.push_back(symbol(*it,addr,size));
     }
     return symbols;
 }
 
 std::vector<symbol> read(const COFFObjectFile& obj) {
     if (obj.getBytesInAddress() == 4) {
-	return readPE32(obj);
+	const pe32_header *pe32;
```
ivg: this is a copy-paste of https://github.com/BinaryAnalysisPlatform/bap/pull/517/files#diff-fe06f65ca8c62da7846dcdba4b819235R196, it definitely needs some abstraction.

ivg:
Also, can you do the same trick with `symbol` structure, i.e., to remove all this abstraction burden away. 

P.S. I'm specially retarding the merge, as I want to integrate the testsuite before, so that we can be sure, that everything is fine :)

parkre:
Please see latest pull request, which includes #517

PR #520, opened 2016-07-14
Integrate testsuite

### LINKED ISSUE #375: Add regression tests for disassembler
Per #372, we should test more robustly for regression bugs in the disassembler.

ivg:
This PR will integrate the platform testsuite with Travis CI. Every time a new build is spawned,
the testuite will be checked out and run, after the unit testing. 

This will finally resolve #375

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "testsuite"]
+	path = testsuite
+	url = https://github.com/BinaryAnalysisPlatform/bap-testsuite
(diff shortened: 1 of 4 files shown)


ivg:
I've also created a branch https://github.com/BinaryAnalysisPlatform/bap-testsuite/tree/pr512 that works with the new syntax. 

Ideally, we should first introduce new syntax, then update our tests, only then remove old syntax.

PR #525, opened 2016-07-15
LLVM Plugin Refactor + Corrected COFF Section Addresses

### LINKED ISSUE #516: Sections in PE32 and PE32+
In LLVM-3.4 COFF Binary Object produces incorrect addresses for symbols, segments and sections. The first two were fixed in #418, but sections were forgotten. So, they still have relative addresses, that doesn't intersect with address space of symbols, and as a result we have empty output when we try to print a disassembly of a program, e.g., 
```
bap -dasm bin/i686-w64-mingw32-echo
```
produces empty output.

parkre:
This PR resolve #516 and finish #517

### DIFF
--- a/plugins/llvm/llvm_binary.hpp
+++ b/plugins/llvm/llvm_binary.hpp
@@ -69,8 +69,8 @@ std::vector<MachOObjectFile::LoadCommandInfo> load_commands(const MachOObjectFil
 
 const pe32plus_header* getPE32PlusHeader(const COFFObjectFile& obj) {
     uint64_t cur_ptr = 0;
-    const char * buf = (obj.getData()).data();
-    const uint8_t * start = reinterpret_cast<const uint8_t *>(buf);
+    const char *buf = (obj.getData()).data();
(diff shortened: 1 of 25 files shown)


plugins/llvm/llvm_binary.hpp
```diff
@@ -434,39 +345,67 @@ using namespace llvm;
...
+section make_section(const coff_section &s, const uint64_t image_base) {
+    return section{s.Name, s.VirtualAddress + image_base, s.SizeOfRawData};
+}
+
+template <typename T>
+std::vector<section> readPE(const COFFObjectFile &obj, const T image_base) {
+    std::vector<section> sections;
+    for (auto it = obj.begin_sections();
+         it != obj.end_sections(); ++it) {
+        const coff_section *s = obj.getCOFFSection(it);
```
ivg: shouldn't we check that the `coff_section` is not `null`? In the rest part of the code we do the check...

ivg:
Ivan is happy now :)

PR #526, opened 2016-07-15
updated testsuite

ivg:
The testsuite started to produce spurious fails, that are totally false positives. 
I've rewrote it in a different way, so that these errors will stop (I hope), 
here is the excerpt from a corresponding commit to the bap-testsuite repo:

## Rewrote log parsing

Previously it was nice `spawn cat $logfile` with `expect` afterwards,
that must work perfectly fine, and worked on all machines, including
Travis, for a long time. But for some reason it started to spuriously
trigger failures, although the log file indeed contained the expected
strings. I have no explanation for this behavior, and finally falled
back to rewritting the log parsing using ugly foreach cycles.

Perhaps, something is wrong with Travis+caching+something else. This
service is very unreliable. Maybe it would be better for us to switch
from Travis to our custom test automation. *angry*


ivg:
damn, what's wrong with github and travis 😫

PR #531, opened 2016-07-28
Introduce -p, --pass, --passes

jaybosamiya:
This PR introduces `-p`, `--pass`, `--passes` as command line options for specifying passes. It does **not** remove `--PASS` though.

The reason that this is done as a quick temporary fix to the code (in a separate PR from #512) is so that the test suite can be updated to work with master (after this is merged) and the transition is smoothened, and we are more confident of the changes.

### DIFF
--- a/src/bap_main.ml
+++ b/src/bap_main.ml
@@ -200,7 +200,10 @@ let program source =
                                a b c d e f g i j k []), passopt in
   let open Bap_cmdline_terms in
   let passopt : string list Term.t =
-    let doc = "Runs passes (comma separated)" in
+    let doc =
+      "Runs passes (comma separated). This option replaces the \
+       previously existing $(b,--)$(i,PASS) options which are now \
+       deprecated and will soon be removed." in
     Arg.(value & opt (list string) [] &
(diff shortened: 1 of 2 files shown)


src/bap_main.ml
```diff
@@ -195,10 +195,16 @@ let program_info =
...
 let program source =
   let create
-      a b c d e f g i j k = Bap_options.Fields.create
-      a b c d e f g i j k [] in
+      passopt
+      a b c d e f g i j k = (Bap_options.Fields.create
+                               a b c d e f g i j k []), passopt in
   let open Bap_cmdline_terms in
+  let passopt : string list Term.t =
+    let doc = "Runs passes (comma separated)" in
```
ivg: Since it is possible that this temporal hack will live for a quite a long time, a month or so, can you add a little bit better doc strings. Or, at least, we can redirect a reader to the `--<pass>` option description, by saying in `doc`, that `--passes=p1,p2,p3 is an alias to --p1 --p2 --p3. Please, use the former, as the latter is deprecated and soon will be removed"

Issue #440, opened 2016-05-25
IDA loader option doesn't create segments?


rvantonder:
`bap /bin/true --loader=ida -lmain --main | less` is empty.
`bap /bin/true -lmain --main | less` has segments.

```ocaml
open Core_kernel.Std
open Bap.Std
include Self()
open Format

let print_segments p =
  Project.memory p |> Memmap.to_sequence |> Seq.iter ~f:(fun (mem,x) ->
      Option.iter (Value.get Image.segment x) ~f:(fun segment ->
          printf "Segment: %s@.%a@." (Image.Segment.name (segment)) Memory.pp mem))

let () = Project.register_pass' (fun proj ->
    print_segments proj)
```

ivg:
yep, it doesn't, because IDA doesn't have segments in the ELF's sense. IDA's segments maps to ELF's sections :) As a workaround we can create a fake segment for each section (1-to-1 mapping). A more elaborated and (correct?) way, would be to merge contigious sections back into segments, based on whether they are code or data. But the latter doesn't guaranteeee a correct mapping to the real segments.

dbrumley:
AFAIK, IDA plugins often merge in sections into one segment.  At least one plugin in the SDK does that (the source to many IDA loaders is in the SDK in case that wasn't known).

ivg:
We can recreate segments by merging consequent sections with the same permissions. Putting this work into low priority queue.

ivg:
No clear resolution, feel free to reopen, and propose a particular solution.

PR #550, opened 2016-09-12
Add IDA brancher and kernel symbol recovery

rvantonder:
This introduces:
- IDA brancher. Specify `--brancher=ida` to use IDA's information of
  branch points / jump edges.
- More IDA loader functionality. The `--loader=ida` option will now also
  add IDA's "synthetic" sections to the binary image, such as `extern`.
- Linux kernel module symbols. Kernel symbols are fixed up at runtime.
  To get the symbols statically in BAP, we need a) relocation information
  of which symbols are called at which addresses, and b) to fix up the CFG
  at the callsites to kernel symbols. Part (b) is performed by the
  `ida_ko_symbol_mapper` plugin.

### DIFF
--- a/plugins/ida/ida_ko_symbol_mapper.ml
+++ b/plugins/ida/ida_ko_symbol_mapper.ml
@@ -1,0 +1,147 @@
+open Core_kernel.Std
+open Bap.Std
+include Self ()
+open Graphlib.Std
+
+let map_pc_to_dest w relocs =
+  let open Option in
+  List.find_map relocs ~f:(fun (pc,dest) -> some_if (w = pc) (dest))
+  >>= fun dest -> return (Bil.int dest)
+
+class bil_ko_symbol_relocator relocs = object(self)
+  inherit Stmt.mapper as super
+
+  (** If there's a BIL jmp with an address to one of our assembly addresses, in
+      relocs, it currently jumps to itself. We need to update it to jump to the
+      extern functions in the lookup table provided by IDA brancher. We
+      substitute it with [w], our address in the table *)
+  method map_jmp e =
+    let open Option in
+    match e with
+    | Bil.Int w ->
+      map_pc_to_dest w relocs
+      |> Option.value ~default:(Bil.int w)
+      |> super#map_jmp
+    | _ -> super#map_jmp e
+end
+
+module Cfg = struct
+  include Graphs.Cfg
+
+  (** No edge update method? Rolling my own... *)
+  let update_edge cfg e e' = cfg |> Edge.remove e |> Edge.insert e'
+
+  (** use the same label (edge which is `Jump, `Cond, etc) *)
+  let update_dst b cfg e =
+    let src = Edge.src e in
+    let lbl = Edge.label e in
+    let e' = Edge.create src b lbl in
+    update_edge cfg e e'
+
+  let update_src b cfg e =
+    let dst = Edge.dst e in
+    let lbl = Edge.label e in
+    let e' = Edge.create b dst lbl in
+    update_edge cfg e e'
+end
+
+let full_insn_of_mem mem arch =
+  Disasm_expert.Basic.with_disasm ~backend:"llvm" arch
+    ~f:(fun dis ->
+        let open Disasm_expert.Basic in
+        dis |> store_kinds |> store_asm |> fun dis' ->
+        insn_of_mem dis' mem)
+
+let remap_block b memory arch relocs =
+  let bil_remapper = new bil_ko_symbol_relocator relocs in
+  let arch = Arch.to_string arch in
+  (* We need a full_insn type to reconstruct an Insn with an updated
+      bil. So re-disassemble from mem to get the full_insn type *)
+  Block.insns b |> List.map ~f:(fun (mem,insn) ->
+      let bil = Insn.bil insn in
+      let bil' = bil_remapper#run bil in
+      match full_insn_of_mem mem arch with
+      | Ok (_,Some x,_) ->
+        (mem,Insn.of_basic ~bil:bil' x)
+      | _ -> (mem,insn))
+  |> Block.create memory
+
+(** Destructs a disassembly block to instructions, transforms the bil with
+    relocation information in remap_block, and reconstructs / updates edges. *)
+class remap_cfg entry relocs arch = object(self)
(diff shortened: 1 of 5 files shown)


plugins/ida/ida_main.ml
```diff
@@ -4,15 +4,21 @@ open Bap_future.Std
...
+open Ida_info
+open Ida_commands
+
 include Self()
 
-module Symbols = Data.Make(struct
-    type t = (string * int64 * int64) list
-    let version = "0.1"
-  end)
+module Ida_futures = struct
```
rvantonder: After discussion, it's clear that I'm probably abusing the Future library here. But here's why it is the way it is:

These futures hold IDA information, which I populate in the loader if the cache does not already contain the information (caching may or may not be enabled). Thus, any read of img/symbols/brancher will first check if it's in the cache. If it's not in the cache, then it will try to read from these futures.

The correct thing here may be to wrap symbols / brancher / img in streams and then wrap any reads of these streams where I am currently doing `read_ida_future`.

plugins/ida/ida_main.ml
```diff
@@ -166,12 +252,70 @@ let checked ida_path is_headless =
        (ida_path/"plugins"/"plugin_loader_bap.py"))  >>| fun () ->
   ida_path
 
+let run_ko_symbol_mapper_pass ida_futures =
```
rvantonder: This pass changes the Cfg recovered from disasm (BIL), which is then propagated to the IR construction. Therefore this pass must run after collecting branch information, but before other passes like API which manipulate the IR. For instance, if this were a standalone plugin, we would have to run passes like API again after running this pass, which seems redundant. The current solution is to run this pass right after the IDA information is available.

plugins/ida/ida_ko_symbol_mapper.ml
```diff
@@ -0,0 +1,147 @@
...
+  let bil_remapper = new bil_ko_symbol_relocator relocs in
+  let arch = Arch.to_string arch in
+  (* We need a full_insn type to reconstruct an Insn with an updated
+      bil. So re-disassemble from mem to get the full_insn type *)
+  Block.insns b |> List.map ~f:(fun (mem,insn) ->
+      let bil = Insn.bil insn in
+      let bil' = bil_remapper#run bil in
+      match full_insn_of_mem mem arch with
+      | Ok (_,Some x,_) ->
+        (mem,Insn.of_basic ~bil:bil' x)
```
rvantonder: Note that this line changes the semantics of the bil attached to the jmp instruction associated with a call site of a relocation. I.e., before it was a relocatable PC-relative jump to itself, and now it jumps to a kernel symbol at an address in a synthetic segment in memory. There seems to be no better way to go about it (we can't insert BIL statements without an associated instruction, as far as I'm aware, and that's probably a good thing).

plugins/ida/ida_main.ml
```diff
@@ -166,12 +252,70 @@ let checked ida_path is_headless =
...
+        let lookup = match Brancher_info.Cache.load id with
+          | Some lookup -> lookup
+          | None ->
+            info "Ko_symbol_pass: No caching enabled, using futures!";
+            read_ida_future_list (fst ida_futures.brancher) in
+
+        let simpl_relocs lookup =
+          let (!) = Word.of_int64 ~width:32 in
+          List.fold ~init:[] lookup ~f:(fun acc (addr,_,l) ->
+              match List.hd l with
```
rvantonder: `l` contains a list of potential 'else' target jumps. Most of the time, there's only one else branch target. But, IDA can give us more (for instance, for jump tables). There is a separate pass in the works for emitting all the jumps of jump tables, but it is not ready yet. Until then, we roll with the first (and most of the time, only) else branch in the lookup table.

plugins/ida/ida_main.ml
```diff
@@ -22,63 +28,110 @@ module type Target = sig
...
+let handle_other_flow default rest =
+  let (!) = Word.of_int64 ~width:32 in
+  match rest with
+  | [] -> []
+  | l -> List.fold l ~init:[] ~f:(fun acc dest_addr ->
+      match List.find default ~f:(fun (addr,_) -> addr = Some !dest_addr) with
+      (* If IDA dest addr is in default BIL brancher, use same edge info *)
+      | Some (Some addr,kind) -> (Some !dest_addr, kind)::acc
+      (* XXX heuristic: we are assuming `Jump always (when it could be cond) *)
+      | _ -> (Some !dest_addr, `Jump)::acc)
```
rvantonder: So it's possible that this jump may under certain conditions be `Cond and not `Jump. I.e., conditional calls in ARM. I use the BIL brancher to get the kind of jump when available. When not, this is just assuming `Jump (there isn't a nice utility function like `val kind_of_jmp : bil -> kind` in the default brancher code... I tried to extract such a utility from the default brancher code, but didn't manage.

plugins/ida/ida_main.ml
```diff
@@ -166,12 +252,70 @@ let checked ida_path is_headless =
        (ida_path/"plugins"/"plugin_loader_bap.py"))  >>| fun () ->
   ida_path
 
+let run_ko_symbol_mapper_pass ida_futures =
+  Project.register_pass ~autorun:true ~name:"ko_symbol_mapper" (fun proj ->
+      match Project.get proj filename with
+      | Some file when String.is_suffix file ".ko" ->
```
rvantonder: Right now, this pass is triggered on files with a ".ko" extension. Open to other suggestions... (maybe a switch?)

rvantonder:
If/when this gets merged, it will also resolve #440

ivg:
It would be nice, if we can split this PR into two parts:

1. the brancher part
2. everything else.

Or even easier, just remove everything else, and keep the brancher part. The reason is that we now have a full fledged support for kernel modules, so we no longer need ko_symbol_mapper and stuff like that. But having the IDA brancher would be very nice, given its ability to recover switch statements.

@rvantonder, do you have some free time and/or desire to do this in the next couple of weeks or month?

rvantonder:
Soonest I can look at it is around 4 weeks from now... so I may have a chance then.

ivg:
Cool, then I will put it in our work list, and if it will popup sooner, we will do it :)

ivg:
We still need the IDA brancher, but I will close this PR so that it won't hang around our PR queue.

PR #556, opened 2016-09-19
constraint ocurl version to 0.7.7

ivg:
It looks like that ocurl.0.7.8 doesn't compile on Ubuntu Precise,
although it works fine at least on the Trusty.

### DIFF
--- a/opam/opam
+++ b/opam/opam
@@ -151,7 +151,7 @@ depends: [
     "ounit" {build}
     "ocamlgraph"
     "ocamlfind" {>= "1.5.6" & < "2.0"}
-    "ocurl" {>= "0.7.7"}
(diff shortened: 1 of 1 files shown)


hcarty:
This may be worth submitting upstream - several other libcurl features are already wrapped behind version checks.  Something new may have been introduced and be in need of such a check.

ivg:
@hcarty, yep, but I doubt, that I will have enough time to do this in the nearest future. Also, the opam packaging of the ocurl can be done better, with a proper `conf-curl` that will be versioned, so that we can explicitly constraint `ocurl` versions with the `curl` versions.

hcarty:
Do you mind if I submit an issue to the `ocurl` project bringing up the 12.04 compatibility issue, linking to this PR?

PR #573, opened 2016-10-06
added pretty printers for some events in bap-traces

gitoleg:
this PR adds pretty printers for next events in `bap-traces`: `timestamp`, `pc_update` and `context_switch`.


ivg:
Why did you choose to print names that are different from the names of corresponding tags?

gitoleg:
Do you mean underscore sign ? like `pc_update` instead of `pc-update` ?

ivg:
yep

gitoleg:
updated

Issue #571, opened 2016-09-30
segfault on a non-elf file


ivg:
The LLVM backend doesn't perform any sanity check on the input file, and just maps it to the memory structures.  In particular this leads to a segfault if a non-elf file is fed to bap, e.g.,

```
cat > test << EOF
> Hello, world
> EOF
bap test -d
1
Segmentation fault (core dumped)
```

There is also a strange `1` outputted, that shouldn't be there.

PR #578, opened 2016-10-12
fixed segfault in llvm stubs

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/571

Although segfault could happen with every type of files, because `llvm` just can't guess a magic, as it is for example given in the issue.

### DIFF
--- a/plugins/llvm/llvm_binary.hpp
+++ b/plugins/llvm/llvm_binary.hpp
@@ -302,11 +302,11 @@ image* create_image_elf(std::unique_ptr<object::Binary> binary) {
 
 image* create_image_obj(std::unique_ptr<object::Binary> binary) {
     if (binary->isCOFF())
-	return create_image<COFFObjectFile>(move(binary));
+        return create_image<COFFObjectFile>(move(binary));
     if (binary->isELF())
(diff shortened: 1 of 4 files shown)


gitoleg:
If you are talking about error codes, I don't see any sense in printing of them. As you remember, without this changes, errors messages looks like:
```
bap test -d
1
Segmentation fault (core dumped)
```
where `1` is an error code, and it doesn't look like something meaningful. More than, this error code doesn't represent a right error, since a `1` corresponds to `arch_not_found`, and `invalid_file_type` is error code(`1`)  that is generated in example in issue. And indeed, having an error code `ec`, we can write a following:
```
std::cout << ec << "  " << ec.value() << std::endl;
```
And result will be `1 2`, so at all, outputting an error code is not a source of knowledge what is going on.

ivg:
> If you are talking about error codes, I don't see any sense in printing of them. 

There is indeed no sense in printing only error numbers, but a proper diagnostics will help. Consider looking at the `error_code` class [interface][1]

[1]: http://en.cppreference.com/w/cpp/error/error_code

ivg:
please, update `testsuite` to the latest master

Issue #569, opened 2016-09-29
LLVM 3.8 Fails Tests


maurer:
I don't know the full extent of tests it fails, but when built against llvm-3.8 the testsuite does not pass.

Look at [Travis](https://travis-ci.org/maurer/bap/builds/162488244) to see 3.8 fail while 3.4 passes, and my [branch](https://github.com/maurer/bap/tree/llvm-3.8-tests) for preliminary work on fixing some of the failures.

I don't know that this indicates any bugs in the 3.8 bindings themselves, only that the tests are brittle to the LLVM version at a minimum.

Issue #553, opened 2016-09-15
static elf binaries with llvm-3.8


ivg:
Due to a bug in the llvm-3.8 we're getting a segfault on a statically compiled stripped ELF binary. The issue affects only the 3.8 version, (i.e., manifests itself on Ubuntu Xenial) and only on static binaries (i.e., such binaries that do not contain `.dynsym` section). 

Since the bug is on the upstream side, we can only provide some kind of workaround, for example, no to call `getSymbolSizes` function if we detect that the binary is static. Also, since totally static binaries, with a stripped `.dynsym` section are quite rare, I don't think that the issue should block the release. 
# The bug

Both `dynamic_symbol_begin` and `dynamic_symbol_end` from `lib/Object/SymbolSize.cpp` (see the code snippets below) have a null-pointer dereference, as `DotDynSymSec` is `NULL` in static binaries. 

``` c++
template <class ELFT>
elf_symbol_iterator ELFObjectFile<ELFT>::dynamic_symbol_begin() const {
  DataRefImpl Sym = toDRI(DotDynSymSec, 0);
  return symbol_iterator(SymbolRef(Sym, this));
}

template <class ELFT>
elf_symbol_iterator ELFObjectFile<ELFT>::dynamic_symbol_end() const {
  const Elf_Shdr *SymTab = DotDynSymSec;
  DataRefImpl Sym = toDRI(SymTab, SymTab->sh_size / sizeof(Elf_Sym));
  return basic_symbol_iterator(SymbolRef(Sym, this));
}
```

PR #590, opened 2016-12-08
fix llvm-3.8 specific issues

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/569
fix https://github.com/BinaryAnalysisPlatform/bap/issues/553
changes:

1) allowed a soft fail on disassembler. Soft fails usually bound with
    some fails in checking of different conditions - like using a some 
    certain register with certain instruction. It leads to undesired behavior when
    disassembling fails for this instruction. And it is the reason, why
   `main` symbol was invisible for some arm files.

2) added a config for `x86` plugin to distinguish
   llvm versions while registering lifting functions;

3) `x86` plugin was reworked to allow some flexibility and
   extensibility for plugin developers;

4) fixes in `lib_test` library, because of differences
   in llvm 3.4 and 3.8.

5) fixed a segfault with stripped static elf files ( better to say that added a workaround, since a segfault is in llvm 3.8)


lib_test/bap_project/test_project.ml
```diff
@@ -5,20 +5,26 @@ open Format
...
+  cmp : cmp;
 }
 
 let arm = {
   arch = `armv7;
   addr = 16;
   code = "\x00\x20\xA0\xE3";
   bil  = "R2 := 0x0:32";
-  asm  = "mov r2, #0x0";
+  asm  = "mov r2, #(0x0|0)$";
```
ivg: fix the instruction to use registers instead of the numerals. The test is not for verifying that we can print numbers, 
but for verifying that we can annotate project structure.

plugins/x86/x86_backend.ml
```diff
@@ -5,14 +5,19 @@ open X86_tools_types
 module Insn = Disasm_expert.Basic.Insn
 module Table = String.Table
 
-module type Data = sig
-  val table : lifter Table.t
+module type T = sig
```
ivg: Please, use `S` for signatures, not `T`

plugins/x86/x86_mov_offset.ml
```diff
@@ -0,0 +1,189 @@
...
+    let module L = (val insn : T) in
+    let module B = (val back : X86_backend.T) in
+    let module S = (val sema : Semantic) in
+    List.iter L.all (fun op ->        
+        let f = S.lift (L.asm_of_t op) L.semantic in
+        let s = L.sexp_of_t op |> Sexp.to_string in
+        B.register s X86_opcode.Mov f)
+
+  let register () =
+    register' (module Mov_oa_ia32)  (module IA32)  (module Sema32);
```
ivg: it is quite obvious, that `register` is not recursive, so you can just use `register` here without the ugly `'`. Alternatively, you can use another name, e.g., `add`

plugins/x86/x86_opcode.ml
```diff
@@ -2,26 +2,17 @@ open Core_kernel.Std
...
-include X86_opcode_cmps
-include X86_opcode_cmpxchg
-include X86_opcode_ins
-include X86_opcode_lods
-include X86_opcode_mov
-include X86_opcode_movs
-include X86_opcode_outs
-include X86_opcode_scas
-include X86_opcode_stos
+type t =
```
ivg: I don't like it, this is just an attempt to move the problem into another dimension, without solving it. 
We either should get rid of opcode type, or implement it as an abstract or extensible type.

plugins/x86/x86_operands.ml
```diff
@@ -14,10 +14,10 @@ let r ~f mem insn =
   | [| Op.Reg reg |] -> f mem reg
   | _ -> invalid_operands ~here:[%here] insn
 
-
 let i ~f mem insn =
   match Insn.ops insn with
   | [| Op.Imm imm |] -> f mem imm
+  | [| Op.Imm imm; Op.Reg reg |] when Reg.name reg = "Nil" -> f mem imm
```
ivg: I think that this change is to global, why do you allow all instructions to be followed by a bogus Nil register? Is it really true that in x86 all instructions have this property? Probably, you should solve this issue local only for offending instructions.

plugins/x86/x86_tools_prefix.ml
```diff
@@ -3,34 +3,25 @@ open Bap.Std
...
-    | `REPE_PREFIX
-    | `REPNE_PREFIX
-    | `EXCEPTION of int
-    | `NONE
-  ] [@@deriving sexp, compare]
+  type t = X86_prefix.t [@@deriving sexp, compare]
 
-  let parse mem op =
-    let code = Memory.(mem ^! min_addr mem) |>
+  let parse mem op =
```
ivg: I think that the whole code that deals with prefixes is broken. Maybe it follows the spec, but it doesn't follow the real life. Also, since we're trying to make the lifter extensible, this particular place is yet another bottleneck, adding new instruction will eventually require us to change this code.

lib_test/.merlin
```diff
@@ -20,3 +20,4 @@ S bap_sema
 S bap_types
 S bap_trace  
 PKG oUnit
+PKG pcre
```
ivg: I think that it should be removed... although it is minor, make sure that you remove it in the next PR.

plugins/x86/x86_prefix.ml
```diff
@@ -0,0 +1,53 @@
+open Core_kernel.Std
+open Or_error
+open Bap.Std
+
+type t = [
+  | `_F0 (** LOCK                *)
```
ivg: I'm not sure that the underscore is a good and nice looking prefix (I got it that you need to have a prefix, as some prefixes doesn't have a predefined semantics, or their semantics is overloaded, so you just want to have type, that represents a subset of integers). My suggestion is to use `x` as the prefix, so that `0x67` is represented as 
```
`x67
```
That looks quite to nice to my taste, what do you think?

plugins/x86/x86_prefix.mli
```diff
@@ -0,0 +1,22 @@
...
+  | `_3E (** DS segment override *)
+  | `_26 (** ES segment override *)
+  | `_64 (** FS segment override *)
+  | `_65 (** GS segment override *)
+  | `_66 (** operand override    *)
+  | `_67 (** address override    *)
+] [@@deriving bin_io, sexp, compare, enumerate]
+
+(** [get mem] - returns a list of prefixes, if any *)
+val get : mem -> t list Or_error.t
```
ivg: I don't think that there is any reason to return an error here. Since prefixes are totally optional, the absence of a prefix should never be considered as an error. You basically abusing here the error type, and treating it as an option.

plugins/x86/x86_prefix.mli
```diff
@@ -0,0 +1,22 @@
...
+
+(** [get_opt mem] - returns a list of prefixes, if any *)
+val get_opt : mem -> t list option
```
ivg: I really dislike adding any special kind of suffixes and prefixes to the function names. We are using the `_exn` prefix, almost because it is 

1. used in the Core library
2. makes code looks ugly, so it should discourage people from using such functions

But, we still trying to keep away from it. Introducing the `_opt` suffix, doesn't fit into the above justifications. Moreover, I don't see why you should return the `t list option`, but not just to return the `t list`. The empty list is an absolutely correct representation of the absence of the prefix. If there could be one or more entities of something, then the list is a perfect abstraction. What does `'a list option` means to you? It's like it is possible to have no prefixes, or... no prefixes, or ....

So, just expose one `get` function, that returns a list of prefixes found in the provided chunk of memory. If there are no prefixes, then there are no prefixes.

plugins/x86/x86_tools_prefix.ml
```diff
@@ -1,40 +1,14 @@
...
-    | 0xF3, #cmps
-    | 0xF3, #scas -> `REPE_PREFIX
-    | 0xF2, #cmps
-    | 0xF2, #scas -> `REPNE_PREFIX
-    | _ -> `NONE
+  type t = X86_prefix.t [@@deriving sexp, compare]
 
   let lock bil = Bil.special "lock" :: bil
 
+  let exception_ n = [ Bil.cpuexn n ]
```
ivg: do you really need this new definition? It doesn't introduce any abstractions and is even bigger  than its own body. I would call it Mr. Mackey definition :)

plugins/x86/x86_prefix.ml
```diff
@@ -0,0 +1,53 @@
...
+  Seq.unfold_step ~init:(Memory.min_addr mem)
+    ~f:(fun addr ->
+        match Memory.get ~addr mem with
+        | Ok word -> Seq.Step.Yield (word, Word.succ addr)
+        | Error _ -> Seq.Step.Done)
+
+let get mem =
+  let rec loop bytes prefs =
+    match Seq.next bytes with
+    | None -> Error (Error.of_string "Decode prefix: opcode not found")
```
ivg: it's not an error!

plugins/x86/x86_prefix.ml
```diff
@@ -0,0 +1,53 @@
...
+
+let deref: Memory.t -> word Seq.t = fun mem ->
+  Seq.unfold_step ~init:(Memory.min_addr mem)
+    ~f:(fun addr ->
+        match Memory.get ~addr mem with
+        | Ok word -> Seq.Step.Yield (word, Word.succ addr)
+        | Error _ -> Seq.Step.Done)
+
+let get mem =
+  let rec loop bytes prefs =
```
ivg: It can be expressed more clearly using `Seq.fold`, or even better with few maps and `take_until`, e.g.,
```
Seq.(bytes >>| Word.to_int >>| ok_exn >>| to_known_prefix >>| take_until ~f:not_none) 
```

plugins/x86/x86_opcode_mov.ml
```diff
@@ -155,12 +117,10 @@ type mov_ia32 = [
 
 type mov_amd64 = [
   | mov_rr
-  | mov_ri
+  | mov_ri
```
ivg: you should teach your Emacs to remove trailing whitespaces.

plugins/x86/x86_backend.ml
```diff
@@ -1,50 +1,61 @@
...
-        | `EXCEPTION no -> Ok [ Bil.cpuexn no ]
-        | `NONE -> lift mem insn
-      with exn -> Or_error.of_exn exn in
-    String.Table.set
-      table
-      ~key:(X86_opcode.sexp_of_t op |> Sexp.to_string)
-      ~data:lift
+module Make (PR : PR) = struct
+
+  type record = {
```
gitoleg: I would say that approach to register prefix handlers is wrong. because at first we register just one more function with our lifter function - so a question why a lifter can't do a whole work and treat prefixes as needed? and at second, handler `prefix -> bil -> bil` will not suitable for all types of prefixes, e.g. address or data size overriding prefixes, since instruction `bil` code will be dependent from prefix, i.e. will be wrong to generate `bil` code for instruction and prefix separately.

plugins/x86/x86_prefix.mli
ivg: This sound like a general excuse, that totally doesn't suite this case. What kind of reason you're trying to emphasize? Look at the code, the only error condition that you raise is called `"opcode not found"`. How does it differ from the `None`? You do not add any extra semantics. Moreover, even this error condition is subsumed in the `get_opt` function, so you can actually distinguish between the error condition, and incorrect prefix. So no, I'm not taking this excuse. The function type doesn't have any sense.

plugins/x86/x86_backend.ml
ivg: Ok, these points sound good, but they actually stand against the current design. If there are prefixes that modify current instruction semantics in a such way, that you can't fix it later on a bil level, then you current design will not work also. Now, the question is, why we're trying to treat prefixes as a separate function. Maybe they should be handled at each particular instruction case? We have a module that provides convenience functions for prefix handling, so probably, instead of registering any callbacks we should just handle the prefixes straightforwardly in each instruction case.

ivg:
Looks good to me, feel free to merge as soon as Travis allows

Issue #391, opened 2016-03-26
pshufd semantics bug


percontation:
There's a typo affecting the semantics of the pshufd instruction, e.g. `pshufd 0x1, %xmm0, %xmm1`.

The `lsl` should be a `lsr` on https://github.com/BinaryAnalysisPlatform/bap/blob/93ee72c0768a9a3692e48e599f1a72431d129e58/plugins/x86/x86_utils.ml#L39

ivg:
@feseal, can you please also take a look on this issue?

ivg:
fixed in #592.

PR #592, opened 2016-12-15
fix an extract function for x86 pshufd/pshufb insns

gitoleg:
this expression `Bil.(cast low t (e lsl (n * (int_exp t et))))`
means to take `low t` bits from an `e lsl (n * (int_exp t et))`,
where the expression from the right of `lsl` actually
evaluates to a start bit, where to start extaction. so doing a
left shift and taking low bits is wrong approach

those changes influences on instructions: `PSHUFB` and `PSHUFD`


ivg:
resolves #391

gitoleg:
yeap, and necessary to edit changes file.

Issue #381, opened 2016-03-18
Remove plugins from bap without specifying `.plugin` extension


rvantonder:
When we list plugins with `--list-plugins`, there's no extension on the plugin names. But to remove a plugin currently, we always have to add the `.plugin` extension to the name. If we can match `<plugin-name>.plugin` and `<plugin-name>` without the extension, that would be helpful for users.

ivg:
Then for the symmetry we should do the same for the `bapbuild install` command.

PR #593, opened 2016-12-15
use bapbundle with plugins without specifying extension

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/381

allows to install/remove bap plugins without specifying extensions, e.g.
 `bapbundle install mycode` instead of `bapbundle install mycode.plugin`. 
The latter version is still available. The same is for `update` and `show` commands.

### DIFF
--- a/tools/bapbundle.ml
+++ b/tools/bapbundle.ml
@@ -53,12 +53,19 @@ let named_resource r = match String.split ~on:'=' r with
   | [name;file] -> Some name, file
   | _ -> None,r
 
-let open_bundle () = match target.contents with
+let normalized x =
+  let suffix = ".plugin" in
+  let with_suffix name =
+    if Filename.check_suffix name suffix then name
+    else name ^ suffix in
+  let is_plugin x = Sys.file_exists (with_suffix x) in
+  if is_plugin x then with_suffix x else x
+
+let open_bundle () = match normalized (target.contents) with
   | "" -> raise Target_unspecified
   | x when Sys.is_directory x -> raise Target_is_directory
   | x when Sys.file_exists x -> Bundle.of_uri (Uri.of_string x)
   | x -> raise Target_doesn't_exist
-
 
 module Update = struct
   let resources = ref []
@@ -178,6 +185,7 @@ module Install = struct
 
   let main () =
     if target.contents = "" then raise Target_unspecified;
+    target := normalized !target;
     if Sys.file_exists !target
     then if Sys.is_directory !target
       then raise Target_is_directory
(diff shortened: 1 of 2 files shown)


tools/bapbundle.ml
```diff
@@ -233,7 +246,7 @@ let usage_msg = "USAGE
 DESCRIPTION
 
   Manages BAP bundles. If you're looking for a way to
-  create a bundle, then bapbuild is the tool that you're
+  create a bundle, then bapbundle is the tool that you're
```
ivg: haha, that's cool))

tools/bapbundle.ml
```diff
@@ -53,13 +53,25 @@ let named_resource r = match String.split ~on:'=' r with
   | [name;file] -> Some name, file
   | _ -> None,r
 
-let open_bundle () = match target.contents with
+module Plugin_bundle = struct
```
ivg: This is kind of strange choice of naming, as well as a strange module.  You named a module as a `Plugin_bundle` so a person who will later read/modify this code, will assume that the module is responsible for the plugin's bundle implementation. However, it is only responsible for handling filenames of plugins. Moreover, some functionality of the module (the `rename_if_plugin` function) leaked out (that's why you were in need to use local open).
 
If you really think, that we need any modularization here, then I would suggest to name it `Name`.  
The `rename_if_plugin` function should be moved inside, and renamed to `normalized`. Speaking about using the  `rename`  verb in function names: this name sounds like an action, that has a side effect. A casual reader, will think that function `rename` will actually rename files on a filesystem. In general, try not to use verbs in names of pure functions. Stick to nouns, adjectives and participles. If a function performs non-desctructive transformation (aka functional update), use either `with_*` form, or past participle, e.g., `Student.with_name student newname`, or `Student.renamed student newname`.

tools/bapbundle.ml
```diff
@@ -189,7 +202,7 @@ module Remove = struct
   let args = [destdir_arg]
   let main () =
     if target.contents = "" then raise Target_unspecified;
-    let file = Filename.concat !destdir !target in
+    let file = Filename.concat !destdir !target |> rename_if_plugin in
```
ivg: Try not to abuse the pipe operators (including `>>|` and `>>=`). These operators are used to build a pipeline of actions. For example:
```ocaml
let prepare file = 
    file |> move |> rename |> fix |> move ...
```

Notice, that all constituents of the pipeline, except the first one, are actions, i.e., verbs that are expected to have an effect. 

In our case, your usage of the pipeline looks justified, except that, as I explained above, the `rename_if_plugin` is not an action. So, the pipeline here just amplifies the delusion, that rename will actually rename the file. 

In fact, if we will use a proper naming discipline here, it would look like:

```ocaml
let file = path |> normalized 
```

Unless you're French this order is a little bit weird and the readability sucks, as `normalized path` is much better. But what is more significant, the absence of the pipeline operator tells us, that there are no side-effects here. 

If you just want to reduce the amount of parentheses, then I would suggest you either to use a normal application operator:

```ocaml
let file = Name.normalized @@ Filename.concat !destdir !target in
```

or multiple let bindings:

```ocaml
let file = Filename.concat !destdir !target in
let file = Name.normalized file in
...
```

Issue #572, opened 2016-09-30
fix compilation warnings for 4.03


ivg:
We can't fix all of them because some require code changes that are not backward compatible. But many of them should be fixed (and actually look valid). 

Those that can't be fixed (that require switching to a new syntax, that is not available on <4.03), should be silenced with a proper `-w` flag.

PR #596, opened 2016-12-16
fixed some compilation warnings

gitoleg:
fixed https://github.com/BinaryAnalysisPlatform/bap/issues/572

next warnings were silenced with `ocaml.warning` attribute:
```
Warning 3: deprecated: [@@noalloc] should be used instead of "noalloc"
Warning 3: deprecated: Format.pp_close_tbox
Warning 3: deprecated: Format.pp_open_tbox
Warning 3: deprecated: Format.pp_print_tab
Warning 3: deprecated: Format.pp_set_tab
```
next warnings were fixed by code edition
`Warning 57: Ambiguous or-pattern variables under guard`

next warnings weren't fixed and still appears during compilation:
```
Warning 58: no cmx file was found in path for module Clexer, and its interface was not compiled with -opaque
Warning 58: no cmx file was found in path for module Cparser, and its interface was not compiled with -opaque
Warning 58: no cmx file was found in path for module Topdirs, and its interface was not compiled with -opaque
Warning 58: no cmx file was found in path for module Z, and its interface was not compiled with -opaque
Warning: -pp overrides the effect of -syntax partly
```

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 2. PR#590 fixed llvm 3.8 specific issues
 3. PR#592 fixed a bug in lifting x86 PSHUFD/PSHUFB instructions
 4. PR#595 fixed bap exit status
+5. PR#596 fixed most of compilation warnings
 
 ### Features
 1. PR#593 bapbundle: it is no longer needed to specify the .plugin extension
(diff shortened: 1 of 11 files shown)


ivg:
The deprecated tabular boxes should also be fixed.

ivg:
The patch looks fine, except, that the issue with tabluations should also be addressed. Would you like to solve it with this PR? You would prefer to open a separate issue and PR?

gitoleg:
I will solve with this PR

Issue #606, opened 2017-01-03
bap-x86 plugin references undefined configuration parameter


ivg:
The x86 plugin is trying to reference a variable that is defined in another plugin, that actually breaks the build on configuration time when the installation is governed by OPAM. 

Since we can't actually peek into the configuration of another package (there is no such interface), we can either create our own variable for the LLVM version and pray that this variable is consistent with the actual version. Or we can expose this variable in the LLVM plugin. The latter will actually require us to install some part of the plugin as a library. Probably it would be a good idea, as it will also facilitate Matt's work on maurer/libbap#4. 

The proposed interface of the `llvm_disasm` library would be:

```ocaml
val init : ?x86_syntax:[`att | `intel] -> unit Or_error.t
val version: string
```

Yet another alternative, would be to use the plugins interface and query the plugin for the metadata, however, I dislike this, as it will create a ordering requirement on plugins, that we're trying to avoid.

Issue #607, opened 2017-01-03
Docker CI builds the released version


ivg:
Currently, our Docker CI is triggered every merge, but what it builds is the latest release. What we really need, is to build whatever was merged. What I'm suggesting, is to create a new branch in our opam-repository fork, that will build BAP packages directly from the master branch (using the git url, instead of the archive). This will facilitate the release process, as we may be assured that whatever we are realising is actually buildable. 

We also, need to have some feedback from dockerhub, that the build was successful  (currently one need to go and look at the queue, not very CIish).

ivg:
Speaking of the integration, we should be able to get a nice DockerHub build status icon (ala the Travis Build status icon) once the Badges/shields#856 is merged.

PR #613, opened 2017-01-12
got rid of `llvm_version` parameter in x86 plugin

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/606

now `x86 plugin` is independed from `llvm_version` configuraion parameter


ivg:
The real test, would be to check that we can build `x86` and `llvm` indepdently, e.g.,

You can test it manually withing a fresh switch:

```
opam install bap-std
git clone <this-branch>
./configure --enable-llvm ...; make; make install
./configure --enable-x86 ...; make; make install
```

However, a correct way of testing would be to finish #607 first. So far, I would assume that the change correct, but we will not release unit #607 is closed.

Issue #579, opened 2016-10-12
implement proper error handling in llvm-image backend


ivg:
# Introduction

Currently, an error handling in the LLVM image backend (aka loader) is totally broken. It relies on a mechanism, that brings C++ runtime into an undefined state so that any consequent call to C++ may actually lead to undefined behavior. Fortunately, so far we're not doing any more calls and exit, but in the future we may (and should) so this issue will fire up sooner or later.
# Background

LLVM library forbids the use of exceptions and uses plain old error codes for handling error conditions. Our backend developers bypassed this restriction by relying on the `caml_fail` function, that actually reintroduced exceptions. Unfortunately, this approach violates internal invariants of C++ (and C, and any other language, except OCaml) invariants. The `caml_fail` function will just unwind the stack, ignoring all automatic variables, and pass the control back to the OCaml. In particular, that means, that destructors of automatic variables will not be called. It may lead to other consequences, that depends on the particular C++ runtime used. 
# Mitigation strategies
## Fast and dirty (Workaround)

The workaround would be to substitute `caml_fail` function with the call to `abort`. This will guarantee an abnormal termination of a program, without a possibility to come back to a broken runtime. However, this should be considered as a temporary workaround only as the error handling policy for the loader assumes that broken files are normal and expected, and should be handled without program termination. 
## Proper error handling

Latest LLVM started to rely on the monadic style error handling using `ErrorOr<T>` template class. Our backend should use the same methodology to propagate error back to the OCaml connection. 
## Ideal solution

The loader interface (`Backend` module) doesn't allow to propagate error information back to the OCaml, as the loader function just returns a function type. Since we cannot change the interface anymore, we need to find out some other mechanism for error reporting, that should be optional. One the solutions would be to register an OCaml callback function, that can be used by backends to provide diagnostic information, e.g.,

``` c
   void bap_error(category_t cat, level_t level, const char *fmt, ...)
```

PR #620, opened 2017-01-25
implemeted llvm error handling

gitoleg:
fixed https://github.com/BinaryAnalysisPlatform/bap/issues/579

There are not any calls to `caml_fail` anymore. A new error handling approach uses a special class for delivering errors messages from bottom to top without throwing exceptions or other tricks to abort a program. An `error_or` class intends to hold either value either error. Actually, there is a suitable  `ErrorOr` class in `llvm` library, and probably it would be a good idea to use it in future as it's a
native for`llvm`. But there are some differences between `3.4` and `3.8` versions, that makes it rather tricky or even impossible to use `ErrorOr` class outside version-depended modules. (e.g. `3.4` uses 
own `system_error` while `3.8` uses a system one).


lib/bap_llvm/llvm_value_or.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+    void discard(const error_msg &msg) {
+        _msg = msg;
+        _payload = nullptr;
+    }
+
+private:
+
+    value_or(const self &) { /* deprecated. use move constructor instead. */}
+    void assert_me() { assert(_payload != 0); }
```
ivg: I don't see the deleted assignment operator (that should always come in pair with the delete copy-constructor). Moreover, I don't see the move-assignment (that comes in pair with the move constructor), it should be something like this:

```c++
or_error<T> & operator=(or_error<T>&& other) {
    _payload.swap(other);
    return *self;
}
```

lib/bap_llvm/llvm_value_or.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+private:
+    pointer _payload;
+    error_msg _msg;
+};
+
+// grab error from other value_or
+template <typename D, typename S>
+value_or<D> grab_error(const value_or<S> &s) {
```
ivg: grab have a move-like meaning. It doesn't steal the error from another message. So better name it `copy_error`. Also, there is a question, what if the `s` value doesn't have an error?

lib/bap_llvm/llvm_value_or.hpp
```diff
@@ -0,0 +1,91 @@
...
+    return value_or<D>(s.get_error_msg());
+}
+
+template <typename T>
+value_or<T> make_value_or(const T &t) { return value_or<T>(t); }
```
ivg: it is better to rename these constructors to just `make_value`, e.g., `make_value(image)` makes more sense, than `make_value_or(image)` or (once you will proper rename the class name), `make_error_or(image)` Does it really mean, that an error _or_ value will be created? Depending on a flip of a coin? Nope, it will only create a value.

lib/bap_llvm/llvm_binary.hpp
```diff
@@ -128,54 +130,58 @@ typedef SymbolRef::Type kind_type;
...
     auto symbol_sizes = getSymbolSizes(obj);
-    for (auto it : symbol_sizes)
-        symbols.push_back(make_symbol(it.first, it.second));
-    return symbols;
+    if (!symbol_sizes)
+        return copy_message<symbols>(symbol_sizes);
+    symbols syms;
+    for (auto it : *symbol_sizes) {
+        if (auto addr = get_addr(it.first, obj))
+            if (auto sym = make_symbol(it.first, it.second, *addr))
```
ivg: Loosing diagnostics. Should be:

```c++

auto addr = get_addr(it.first, obj);
if (failure error = addr) warnings.push_back(error << "skipping symbol" << it.first);
```
Here the main problem is with the `warnings`, that is a collection of failures that were ignored. You can either pass it as a separate argument or, as we planned it originally, report it directly to OCaml side via `bap_llvm_diagnostic` function. The latter is dirty but easy, the former is clean but verbose. A usual tradeoff.

lib/bap_llvm/llvm_binary.hpp
```diff
@@ -185,150 +182,161 @@ struct section {
...
+        auto segments = seg::read(*ptr);
+        auto symbols  = sym::read(*ptr);
+        auto sections = sec::read(*ptr);
+
+        if (auto er = entry || segments || symbols || sections) {
+            bap_notify_error(er.message());
+            return nullptr;
+        }
+        binary << entry.warnings() << segments.warnings() << symbols.warnings() << sections.warnings();
+        for (auto w : binary.warnings())
```
ivg: you should print warnings only if BAP_DEBUG is set to 1. This is temporary, of course, as later with the new interface a frontend will take care of warnings.

Issue #588, opened 2016-12-02
Do not store API information in the api plugin


ivg:
Plugins should be sealed after they installed. They actually can be installed to a readonly place, and it makes sense. They can still contain data, but it should be immutable. 

The proposed way to store api files is to save them in the filesystem. Here comes the details:

1. add `api-path` _configuration_ time parameter, that will default to the `$datadir/bap-api`;
2. add `api-path` plugin parameter that will be initialized with the a singleton list consisting of one element - the path provided at configuration time.
3. add `BAP_API_PATH` parameter that will be a colon separated list of paths.   
3. The `api-add` option should copy the provided file into the path. The first value should be chosen, if more than one exits (it would be the default configuration parameter). If the first option is not writable, then the first one writable location should be chosen. The `api-add` should refuse to add a file with a name that is already available in the search path.  The option should be marked as DEPRECATED to discourage a user from its usage. We will remove this option at bap 2.0.0.
4. The `api-remove` option should remove all files with the specified name from the paths. The option should be marked as DEPRECATED to discourage a user from its usage. We will remove this option at bap 2.0.0  Note: ideally, we should remove `api-add` and `api-remove`, as a plugin shouldn't change the user system. A user should create its own repositories and do not modify the global one. If he likes, the he should do it manually with tools appropriate to his operating system. 

5. When ran (without options or with any option except `help` or `version`) the plugin should make a consistency check, that verifies that all names are unique. If the check is not passed it should terminate with an non zero exit code. 
6. Add option `list-paths` that will list registered paths and exit with zero code (unless consistency check fails). This option is intended to be machine readable. So the output should list each path on a separate without any additional noise.

7. Add options `show` that will output information about available aps in a human readable form. The outputted message should contain paths, and file names. Ideally, the output should be grouped by the language name, e.g.,

```
  Language: C
        posix : ~/.opam/devel/share/bap-api/c/posix
        pam: ...
        ...
    Language: OCaml
       pervasives: ....
       ...
   Total number of available API: 84     
```

PR #621, opened 2017-02-01
Api files aren't stored in api plugin now

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/588

features:
- added `api-path` configuration parameter
- added `api-path`, `api-show` and `api-list-paths` plugin parameters
- added `BAP_API_PATH` env variable
- `--api-add` and `--api-remove` works with filesystem now


plugins/api/api_main.ml
```diff
@@ -43,37 +103,95 @@ let main proj =
...
-      | _ -> `Copy)
+let sanity_check paths =
+  let is_same api api' =
+    api.file <> api'.file && api.lang = api'.lang &&
+    Filename.basename api.file = Filename.basename api'.file in
+  let check_for_pair api apis =
+    match List.find ~f:(is_same api) apis with
+    | None -> ()
+    | Some api' ->
+      eprintf "sanity check failed for api plugin, apis with same \
```
ivg: Ok, here we're requiring a user to fix the problem manually, so we need to be very verbose. 

```
Sanitization error: there are conflicting specifications for the same API(s):
<api> <path1> <path2>  ...
....

Each API should be specified only once. Please resolve the conflict manually, either by removing
the conflicting files or by renaming them. 
```

plugins/api/api_main.ml
```diff
@@ -43,37 +103,95 @@ let main proj =
...
+  let all = api_of_paths paths in
+  List.iter ~f:(fun api -> check_for_pair api all) all
 
+let check_paths paths =
+  List.filter_map paths
+    ~f:(fun p ->
+        if Api_path.exists p then Some p
+        else
+          let () = eprintf
+              "Api path %s doesn't exists, excluded from paths\n" p in
```
ivg: First of all it is a warning, i.e., an error condition that doesn't impend further actions. Second, I'm not sure that we should actually make noise about it. Consider, how gcc (or any other compiler)( acts when you specify  `-I<dir>` and `<dir>` doesn't exist? So probably, we should just ignore it.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+  let name t = t.desc.name
+
+  let create lang path =
+    let path = Api_path.of_string path in
+    if Api_path.is_apiable_path path then
+      let name = Api_path.name_of_path path in
+      Some {desc = {lang; name}; path;}
+    else None
+
+  let cp dest api =
```
ivg: do not flip parameters ordering, in `cp` destination is always to the right.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+    let dest' = Api_path.with_lang dest (lang api) in
+    ok_if_true (Api_path.exists dest) ~error:`No_dst >>= fun () ->
```
ivg: You should ensure, that the destination path is a directory.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+    try_with
+      (fun () ->
+         let dest' = Api_path.to_string dest' in
+         FileUtil.mkdir dest';
```
ivg: what is the purpose of the check above, that ensured that the path exists if you anyway create it?

ivg: Also, if you're still going to create a path, then you should use `~parent:true` option, to ensure, that you are able to create the path, if the parents do not exist.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+    | Some path ->
+      exit_with @@ eprintf "Can't find api %s:%s at %s\n"
+        desc.lang desc.name (Api_path.to_string path)
+    | None ->
+      exit_with @@ eprintf "Can't find api %s:%s\n"
+        desc.lang desc.name
+
+  let no_path api =
+    exit_with @@ eprintf
+      "There are not any writable paths for %s plugin\n" api.name
```
ivg: Put it simpler - error messages are for users, not for programmers:

`"Can't store an api %s: permission denied" `

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+  | None ->
+    let rec add = function
+      | [] -> Api_error.no_path api.desc
+      | p :: paths' ->
+        match Api.cp p api with
+        | Ok () ->
+          info "wrote %s:%s to %s"
+            api.desc.lang api.desc.name (Api_path.to_string p)
+        | Error `No_dst ->
+          error "path not exists %s" (Api_path.to_string p);
```
ivg: should be 
```
info "Api.cp: skipping a non-writable destination %a" Api.pp p"
```

It is not an error, or even a warning. There can be a plenty of paths that are writable, so there are no reason to panic.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+          add paths'
+        | Error (`Fail exn) ->
+          error "failed to write at %s path: %s"
```
ivg: A call to `error` should precede a program/plugin termination. If you're reporting an error and then ignoring it, then  either it is not an error, or it is an error in the desoxyribonucleic acid.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+    api_to_rem : api_descr list;
+    list_paths : bool;
+    show_apis  : bool;
+    api_paths  : string list;
+  } [@@deriving fields]
+end
+
+let env_paths =
+  try
+    String.split ~on:':' (Sys.getenv "BAP_API_PATH") |>
```
ivg: Use `Path.filenames_of_path` function.

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+  with Not_found -> []
+
+let all_paths o =
+  List.map ~f:Api_path.of_string
+    (env_paths @ Api_config.api_path @ Api_options.api_paths o)
+
+let get_file file = Some file
+
+let files paths lang =
+  let apis = List.concat @@
```
ivg: Use `List.concat_map`

plugins/api/api_main.ml
```diff
@@ -1,39 +1,215 @@
...
+    List.map ~f:(fun path -> Api_path.find_apis ~lang path) paths in
+  List.map ~f:(fun api -> Api_path.to_string api.path) apis
```
ivg: either join two maps, our use piping (e.g., `|>`, `>>|`, or `>>=`). You can even use monads for this, e.g.,

```ocaml
let files path lang = 
   List.(paths >>= Api_path.find_apis ~lang >>| fun api -> Api_path.to_string api.path)
```

PR #635, opened 2017-03-06
cleanup travis env variables

stephengroat:
should make it easier to read; also adding in os x

can't understand why the linux builds are breaking (https://travis-ci.org/BinaryAnalysisPlatform/bap/jobs/208369664#L1516); didn't change any source


ivg:
Thanks for the contribution.

From your PR I learned two new things:
1. cmdliner.1.0.0 is in opam and it breaks bap (#636)
2. os x workers are now available for mere mortals.

We will fix the first one (thanks for noticing it). The second thing is very interesting. But it looks like, that the container branch of the @talex5 fork of ocaml/ocaml-ci-script doesn't support Mac OS X. And why should it, as far as I know, OS X workers can be run in a container, for obvious reasons. 

Probably, the solution would be to use the upstream ocaml/ocaml-ci-script repo in case of Mac OS X. 

Note: I've copied the .travis-ci.sh file from talex5's repo, as at some point of time the container branch didn't work with the upstream yorick and the family scripts due to a renaming. So, the only solution was to fork the fork :)

stephengroat:
i'll work on a private branch and see if i can get something working and re-submit a pr. i didn't want to create too much more churn

PR #638, opened 2017-03-07
Move to current ocaml-ci-scripts

stephengroat:
also adds osx, currently allowing for failures


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: c
+before_install: if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew upgrade && brew install gmp; fi
+install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-opam.sh
 script: bash -ex .travis-opam.sh
-sudo: false
+sudo: required
```
ivg: We specifically are using the container infrastructure. There are several reasons for this:
1. Greater amount of RAM (that enables faster builds)
2. Better availability of the workers (builds start much faster)
3. Caching (previously caching was enabled only for the container infrastructure).

I'm not sure, though, whether these benefits are still actual.

stephengroat: I'll try to address these in order

1. Yup, these builds will be slower, but they'll follow the `ocaml-ci-scripts` master, which should be updated and current
2. I haven't had too much of an issue with build availability. travis does limit the number of parallel builds that `sudo` builds can do. I'm currently working on a little bit of `bash` that I'm trying to add to `ocaml-ci-scripts` that removes the need for `sudo` if you use the travis `apt` extension
3. `cache` is now available on `sudo` builds

stephengroat:
@ivg [`travis` times](https://travis-ci.org/BinaryAnalysisPlatform/bap/builds/208760452) don't seem too bad for `linux sudo: required`

let me know if these are below what you're looking for and i'll close the pr

ivg:
Time is pretty fine, somewhat about five to ten minutes overhead, but still tolerable. 

What concerning Mac OS X, is it possible to set up the matrix such that we will test it only on one compiler version.

And would you be so kind to remove explicit installation of external dependencies, as they should be grabbed automatically by Yorick (I hope)

stephengroat:
Easy to do one `osx` build, just tell me which ocaml version (4.02,4.03,4.04) and do that.

I can remove the dependencies, but explicitly listing them does speed up the `4.02` build by about 10 minutes

ivg:
The issue with beagle should be resolved by #637.

ivg:
> I can remove the dependencies, but explicitly listing them does speed up the 4.02 build by about 10 minutes

Ouch, yet another 10 minutes is too much I think. Let's stick with the current approach, especially, since we are not planning to add any more system deps, in near future.

Issue #645, opened 2017-03-31
regression on mach-o files


ivg:
This was already reported in #414 and fixed in #418. Now we have it back. 

```
$ bap x86_64-darwin-echo 
LLVM ERROR: Dynamic symbols unimplemented in MachOObjectFile
```

It is in the `ivg/bap/microflux` branch, but I suspect, that it should be in the master branch also.

The `x86_64-darwin-echo` file is obtained from the [echo.c][1], on the primus branch of our testsuite. Our final goal would be to add this file to the testsuite (and to pass all the tests) on it. 

The file was compiled with the following compiler:
```
 gcc -v
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Target: x86_64-apple-darwin15.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Obviously, now we can't pass any test. 

P.S. bap fails on any macho file right now, this one is just the simplest example. 

[1]: https://github.com/BinaryAnalysisPlatform/bap-testsuite/blob/primus/src/echo.c

PR #646, opened 2017-04-02
fixed macho dynamic symbols error when compiling with llvm 3.4

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/645

`llvm 3.4` doesn't support dynamic symbols for macho files, as it could be seen from code snippet below. 
```
symbol_iterator MachOObjectFile::begin_dynamic_symbols() const {
  // TODO: implement
  report_fatal_error("Dynamic symbols unimplemented in MachOObjectFile");
}

symbol_iterator MachOObjectFile::end_dynamic_symbols() const {
  // TODO: implement
  report_fatal_error("Dynamic symbols unimplemented in MachOObjectFile");
}
```

### DIFF
--- a/lib/bap_llvm/llvm_binary_34.hpp
+++ b/lib/bap_llvm/llvm_binary_34.hpp
@@ -218,6 +218,13 @@ error_or<symbols_sizes> getSymbolSizes(const ObjectFile &obj) {
     return success(std::move(sizes));
 }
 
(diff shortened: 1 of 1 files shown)


ivg:
Can you actually explain how this change fixes the issue, and how it specific to 3.4, and whether there are any regressions on 3.8?

gitoleg:
I added an overloaded function, that doesn't read dynamic symbols for macho in case of `llvm-3.4`.

Issue #632, opened 2017-02-22
support kernel style relocations 


ivg:
Introduction
---------------

Kernel modules and some binaries compiled for embedded systems are using relocations of type `R_xxx_PCx` to call to external functions. The main feature of this kind of relocations is that a stub function is not created, and the destination address is set to zero (although, theoretically, it can be any other value, specified in the relocation record). 

Our first attempt (#550) was to use a brancher service that will emit fake addresses for each relocatable call, and then proceed as normal. However, the idea is nice, it is kind of broken. First of all, it is very IDA-specific as the fake addresses are created by IDA and are consistent only with IDA. This may break interactions with other tools. Second, and this is the main problem, it is hard to implement the same scheme without IDA, so only a user that has IDA Pro will be able to analyze kernel modules. 

I'm proposing another solution, that is more consistent and should provide a better approximation of a program. 

Symbolizer
--------------

Our symbolizer interface is basically a function of type `addr -> string option`, that gives a symbolic name to an object (usually a subroutine) located at the specified address. Since there is no address of a called function (it is NULL - a value shared across all externs), a symbolizer can't easily provide the symbolic information, even if it has it. 

 We can change a behavior of symbolizer, without actually changing its interface. The new behavior would be a strict superset of the old behavior, so we shouldn't expect any compatibility issues. As you remember the current behavior of a symbolizer is to return a name of an object for the provided address. We will extend the behavior and allow symbolizers to return a name of symbol referenced by an address. For example, if a symbolizer is asked to symbolize a call instruction, then it may return a name of a callee. During the reconstruction, it would be very easy to distinguish between a caller and a callee, so I wouldn't expect any issues here. 

Brancher
-----------

There are, however, few more issues associated with this kind of relocations. Since the disassembler doesn't know that the jump instruction is actually a call, it treats it as an absolute jump and doesn't emit a fall-through destination. So, a brancher needs to know and handle these relocations. It is not hard to implement an IDA based (in fact #550 already implements it), or binutils based brancher. However, we shouldn't rely on a regular availability of these tools. An ideal solution would be to pass the relocation information from the loaders and use it in our default brancher, however, this approach requires us to define a new interface for the loader and will not play nice with a standalone disassembly. It also requires lots of work, to implement support for different relocations. 

[...]

PR #654, opened 2017-04-25
llvm loader, based on ogre library

gitoleg:
this PR introduces new loader, based on Ogre library.
Also solve an issue with supporting for elf-relocatable files:
https://github.com/BinaryAnalysisPlatform/bap/issues/632

important changes and few useful tricks that available now:
1)  Legacy llvm loader is still available in configuration time:
   `./configure --enable-llvm-legacy` (disabled by default)
In this case bap can be used with either of the next loaders:
`bap --loader=legacy-llvm`
`bap --loader=llvm`
and merged loader (without explicit specification).
2) old elf loader (not llvm-based) is disabled by default, but still available in configuration time:
`./configure --enable-elf`
3) Now it's possible to rebase the whole image relative to some other
base address:  `bap /bin/ls --llvm-base=0x0042` 
4) New loader supports executable files, like it was before, but also can load lot's of other types of files: 
   - macho, coff, elf object files
   - elf kernel modules and macho kernel-modules (unzipped ones and for llvm 3.8. only)
   - macho `*.dylib`, elf `*.so` and coff `*.dll` files
(elf `.got/.plt` entries still not suppoted)
5)  It's enough to specify an appropriate brancher to deal with elf relocatable files. This brancher uses information about relocations to form correct list of destinations. So, in example bellow, the call to `printk` was successfully resolved, and synthetic subroutine created:
```
`bap my_module.ko  -d  --brancher=relocatable` : 
....
00000312: LR := 0xC24:32
00000313: call @printk with return %00000314
00000314: 
....

00000331: sub printk()
```


lib/bap_llvm/bap_llvm_binary.ml
```diff
@@ -63,3 +63,6 @@ external section_addr : t -> int -> int64  =
 
 external section_size : t -> int -> int64  =
   "bap_llvm_binary_section_size_stub"
+
+external llvm_load : Bigstring.t -> string =
```
ivg: prefix our external symbols with `bap` we are not the LLVM team, we are the BAP Team. The chances to get a name collision with other `llvm` function are high.

lib/bap_llvm/bap_llvm_loader.h
```diff
@@ -0,0 +1,6 @@
+#ifndef BAP_LLVM_ELF_LOADER_H
+#define BAP_LLVM_ELF_LOADER_H
+
+const char * load_elf(const char* data, size_t size);
```
ivg: either make it static, or prefix with `bap_llvm` to prevent name collisions.

lib/bap_llvm/bap_llvm_loader.ml
```diff
@@ -88,7 +92,45 @@ let of_data data : Backend.Img.t option =
   Some (Backend.Img.Fields.create
           ~arch ~entry ~segments ~symbols ~sections)
 
+module Check = struct
```
ivg: No, please don't use a file specification for passing diagnostics. Provide an explicit interface for passing the diagnostic messages.

lib/bap_llvm/bap_llvm_loader.ml
```diff
@@ -88,7 +92,45 @@ let of_data data : Backend.Img.t option =
...
+
+module Ogre_loader = struct
+  open Result.Syntax
+
+  let from_data data =
+    match llvm_load data with
+    | "" -> Ok None
+    | s ->
+      match Ogre.Doc.from_string s with
+      | Error er -> Or_error.errorf "can't construct doc"
```
ivg: you are loosing the diagnostics here, the `err` value contains the significant piece of information.

ivg: also, we have `Result.failf` function, so that you can easily construct an error message:
```ocaml
Result.failf "llvm loader provided a broken specification - %a" Error.pp er
```

lib/bap_llvm/bap_llvm_loader.ml
```diff
@@ -88,7 +92,45 @@ let of_data data : Backend.Img.t option =
...
+        | Some er -> Or_error.error_string er
+
+  let from_file path = Bap_fileutils.readfile path |> from_data
+
+end
+
 let init () =
-  match Image.register_backend ~name:"llvm" of_data with
+  Image.register_loader ~name:"llvm" (module Ogre_loader);
+  match Image.register_backend ~name:"llvm-legacy" of_data with
```
ivg: Don't forget to remove this in future. I don't want to have them both, however, we can provde the legacy loader as a separate plugin.

lib/bap_llvm/llvm_loader.h
```diff
@@ -0,0 +1,12 @@
+#ifndef LLVM_LOADER_H
+#define LLVM_LOADER_H
+
+struct doc;
+
+struct doc* doc_load(const char*, size_t);
```
ivg: Do you know anything about namespace pollution? Come on...

lib/bap_llvm/llvm_loader_scheme.hpp
```diff
@@ -0,0 +1,200 @@
+#ifndef LLVM_LOADER_SCHEME_HPP
```
ivg: Do we really need to reimplement Ogre.ml in C++? The idea is that you just do

```c++
scheme << "(segment " << addr << " " << size << " " << name << ")";
```

You do a double work, the OCaml side will check that everything is fixed. The C++ will still suck. I think that this is an overengineering.

lib/bap_llvm/llvm_loader_utils.hpp
```diff
@@ -0,0 +1,91 @@
...
+    std::ostringstream s_;
+};
+
+std::string quoted(const std::string &s) {
+    return "\"" + s + "\"";
+}
+
+std::string declare() {
+    return
+"(declare arch (name str))(declare code-start (addr int))(declare entry-point (addr int))\
```
ivg: use 
```
"xxx" 
"yyy" 
"zzz" 
```
instead of
```
"xxx\
yyy\
zzz"
```

lib/bap_llvm/bap_llvm_ogre_loader.ml
```diff
@@ -0,0 +1,67 @@
...
+  let align     = "aligin" %: int
+
+  let declare name scheme f = Ogre.declare ~name scheme f
+
+  let header () = declare "elf-header"
+      (scheme p_off $ p_size $ hdr_type $ hdr_flags $ align)
+      (fun off size typ flags align ->
+         {off; size; data = (typ, flags, align)})
+
+  let physical () = declare "physical-view"
```
ivg: a couple of comments, describing what the attribute is doing would help.

lib/bap_llvm/llvm_loader_scheme.hpp
```diff
@@ -0,0 +1,64 @@
...
+};
+
+std::string nonempty(const std::string &s) {
+    if (s == "") return "\"\"";
+    else return s;
+}
+
+std::string declare() {
+    return
+"(declare arch (name str))(declare code-start (addr int))(declare entry-point (addr int))\
```
ivg: use implicit string literals concatenation, instead of 
```
"aaa\
bbb\
ccc"
```
Use
```
"aaa"
"bbb"
"ccc"
```

lib/bap_llvm/llvm_loader_stubs.c
```diff
@@ -0,0 +1,37 @@
...
+#include <caml/alloc.h>
+#include <caml/custom.h>
+#include <caml/bigarray.h>
+#include <caml/compatibility.h>
+#include <caml/callback.h>
+
+#include "llvm_loader.h"
+#include "llvm_loader_stubs.h"
+
+void failn(int n) {
```
ivg: use `static` to prevent namespace pollution. All C functions that are not really exported (i.e., that are not part of the interface, should be static).

lib/bap_llvm/llvm_loader_utils.hpp
```diff
@@ -0,0 +1,91 @@
+#ifndef LLVM_LOADER_UTILS_HPP
```
ivg: this file looks like a duplicate of `llvm_loader_scheme`

lib/ogre/ogre.ml
```diff
@@ -219,10 +219,14 @@ module Doc = struct
...
     typecheck_entry name packed doc >>| fun () ->
     let packed = {fields = normalize_entry name packed doc} in
-    {
-      doc with entries =
-                 Map.add_multi doc.entries ~key:name ~data:packed
-    }
+    let equal e e' = compare_entry e e' = 0 in
+    let entries = Map.change doc.entries name
+        ~f:(function
+            | None -> Some [packed]
```
ivg: since both branches return `Some`thing, use `Map.update`.

lib/bap_llvm/bap_llvm_ogre_coff.mli
```diff
@@ -0,0 +1,33 @@
...
+  (** coff symbol *)
+  val symbol :
+    (string * int64 * int64, (string -> int64 -> int64 -> 'a) -> 'a)
+      Ogre.attribute
+
+  (** coff symbol is function *)
+  val function_ : (string * int64, (string -> int64 -> 'a) -> 'a) Ogre.attribute
+end
+
+module Make(Fact : Ogre.S) : sig
```
ivg: So, this is a naming/interface issue. What is `image` is totally non-obvious, as the interface is not clearly specified, it is kind of internal and is known only to you. On the other hand, we have `segment`, `section` and `symbols` and `mapped` propositions, that are documented and well understood. The purpose of the `Make(Fact)` functor is to derive these well-known properties from the properties that are specific to a concrete file format. The deep understanding of the `Make(Fact).{segments,symbols,sections}` functions is that they add rules to our logic database. Thus, my suggestion is to use the following interface:

```ocaml
module Rules : sig
  module type S = sig
    type 'a m
    val segments : unit m
    val sections : unit m
    val symbols : unit m
  end 
  module Make(Fact : Ogre.S) : S with type 'a m := 'a Fact.t
  include S with type 'a m = 'a
end
```

(the `Rules.S` module type can be actually reused by different backends)

And the interface is supposed to be used as follows:

```ocaml
module Derive = Rules.Make(Fact)
let image =  Fact.all_ignore [
     Derive.segments;
     Derive.sections;
     Derive.segments;
   ]    
```

lib/bap_llvm/bap_llvm_ogre_elf.ml
```diff
@@ -0,0 +1,86 @@
...
+  open Scheme
+  open Fact.Syntax
+
+  let segments =
+    Fact.foreach Ogre.Query.(begin
+        select (from program_header $ virtual_program_header $ program_header_flags)
+          ~join:[[field name]]
+      end)
+      ~f:(fun hdr (_,addr,vsize) (_,ld,r,w,x) ->
+          hdr, (addr, vsize), (ld,r,w,x)) >>= fun s ->
```
ivg: use the point-less^Wfree notation

lib/bap_llvm/bap_llvm_ogre_elf.ml
```diff
@@ -0,0 +1,86 @@
+open Core_kernel.Std
+open Bap.Std
+open Monads.Std
+
+open Bap_llvm_ogre_types
+
+module Scheme = struct
+  open Ogre.Type
+
+  (** flags that describes an entry behavior *)
```
ivg: there is no need to duplicate comments in the `ml`, that are already in the `mli`.

lib/bap_llvm/bap_llvm_ogre_coff.mli
```diff
@@ -0,0 +1,33 @@
+
+module Scheme : sig
```
ivg: I think that we shouldn't provide `mli` files for the scheme, and schemes should be provided only as `ml` file. The justification, is that the `scheme` is an embedded DSL that describes data in our data-base. It is crucial to know the field names. For example, take the `section_header` value -- it is not usable, as you don't know how to reference the fields, you also don't know how to construct the values from other languages. On the other hand the definition: `Ogre.declare ~name:"section-header" (scheme name $ off $ size)` is self-describing.

So, provide each scheme as a separate `ml` file, that has only attributes and field definitions.

lib/bap_llvm/bap_llvm_ogre_elf.ml
```diff
@@ -0,0 +1,86 @@
...
+    Fact.Seq.iter s
+      ~f:(fun ((name,off,size), (addr, vsize), (ld,r,w,x)) ->
+          if ld then
+              Fact.provide segment addr vsize r w x >>= fun () ->
+              Fact.provide mapped addr size off >>= fun () ->
+              Fact.provide named_region addr vsize name
+          else Fact.return ()) >>= Fact.return
+
+  let sections =
+    Fact.foreach Ogre.Query.(select (from section_header))
```
ivg: Looks like that the `Fact.foreach query ~f:ident` is a recurring pattern, so feel free to add `Fact.collect query` to the `Ogre` library.

lib/bap_llvm/bap_llvm_ogre_loader.ml
```diff
@@ -0,0 +1,72 @@
...
+  let from_data data =
+    try
+      let doc = Bap_llvm_binary.bap_llvm_load data in
+      Ogre.Doc.from_string doc >>= fun doc -> to_image_doc doc
+    with Llvm_loader_fail n -> match n with
+      | 1 -> Ok None
+      | 2 -> Or_error.error_string "file corrupted"
+      | n -> Or_error.errorf "fail with unexpeced error code %d" n
+
+  let from_file path =
```
ivg: We should think about how to change the loader interface, so that if a `from_file` is not provided, then the image loader will use the `from_data`.

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -0,0 +1,177 @@
...
+    std::size_t i = 0;
+    for (auto it = begin; it != end; ++it, ++i) {
+        bool ld = (it->p_type == ELF::PT_LOAD);
+        bool r = static_cast<bool>(it->p_flags & ELF::PF_R);
+        bool w = static_cast<bool>(it->p_flags & ELF::PF_W);
+        bool x = static_cast<bool>(it->p_flags & ELF::PF_X);
+        auto off = it->p_offset;
+        auto filesz = it->p_filesz;
+        auto name = name_of_index(i);
+        s << "(program-header "  << name << " " << off << " " << filesz << ")";
```
ivg: tip: you can ask `data_stream`'s `<<` operator to add spaces around words automatically. This will make you definitions much shorter.

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -0,0 +1,177 @@
...
+        s << "(virtual-program-header " << name << " " << it->p_vaddr << " " << it->p_memsz << ")";
+        s << "(program-header-flags "   << name << " " << ld << " " << r << " " <<  w << " " << x  << ")";
+    }
+}
+
+template <typename T>
+void section_header(const T &hdr, const std::string &name, data_stream &s) {
+    s << "(section-header " << quoted(name) << " " << hdr.sh_addr << " " << hdr.sh_size << ")";
```
ivg: tip: the `data_stream::<<` operator should handle quotes automatically. It is ok if you will over-quote, e.g., if you will output `"main"` instead of `main`, alternatively, you may implement smart quoting, that will quote only if the word contains whitespaces or parenthesis)

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+template <typename T>
+void symbol_entry(const Elf_Sym_Impl<T> &sym, const std::string &name, uint64_t addr,
+                  data_stream &s) {
+    s << "(symbol-entry " << quoted(name) << " " << addr << " " << sym.st_size << ")";
+    if (sym.getType() == ELF::STT_FUNC)
+        s << "(code-entry " << name << " " << addr << ")";
```
ivg: shouldn't you quote the name here? Nevermind, please add quoting to the `data_stream` (also the `data_stream` name is no good, use `ogre` or `ogre::doc` or something like this).

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+template <typename T>
+void section_headers(const ELFObjectFile<T> &obj, data_stream &s) {
+    auto elf = obj.getELFFile();
+    for (auto it = elf->section_begin(); it != elf->section_end(); ++it) {
+        auto name = elf->getSectionName(it);
+        if (name)
+            section_header(*it, name.get().str(), s);
+        else
+            s.fail(name.getError().message());
```
ivg: this is neat!

lib/bap_image/bap_image.ml
```diff
@@ -349,10 +349,14 @@ module Derive = struct
...
-    Fact.require arch >>= fun s -> match Arch.of_string s with
-    | Some s -> Fact.return s
-    | None -> Fact.failf "unknown/unsupported architecture %s" s ()
-
+    Fact.foreach Ogre.Query.(select (from arch)) ~f:ident >>= fun s ->
+    Fact.Seq.reduce ~f:(fun a1 a2 ->
+        if Arch.equal a1 a2 then Fact.return a1
+        else Fact.failf "arch is ambigous" ())
+        (Seq.filter_map ~f:Arch.of_string s) >>= fun a ->
+    match a with
```
gitoleg: At least because I have to feed some function to `Seq.reduce`  and therefore it will not look so briefly (
_a-m-b-i-g-u-o-u-s_ ... it's hard!!

ivg: yeah, sure `Seq.reduce ~f:Arch.equal`

gitoleg: nope again) the signature is `'a -> 'a -> 'a`, but not `'a -> 'a -> bool` .

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+    | None -> Fact.failf "unknown/unsupported architecture" ()
+
+  let calls insns =
+    let insns =
+      Seq.filter insns ~f:(fun (mem, insn) ->
+          Bap_disasm_insn.(may affect_control_flow insn)) in
+    addr_width >>= fun width ->
+    Fact.collect
+      Ogre.Query.(select (from external_reference)) >>= fun syms ->
+    let tab = Seq.fold ~init:Bap_table.empty ~f:(fun tab (mem,_) ->
```
ivg: two points here: first, do not name variables by their type, e.g., a table, a list, etc, unless you're writing a generic algorithm for that particular container. Name variables according to what they represent, not according to how the represent it. The second issue, is that it is much more readable when you put the value that you're folding near the `fold` function, not like 10 lines later. Basically, it should be easily deductible from code, that this code iterates over `insns`.

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+  open Fact.Syntax
+
+  let addr_width =
+    Fact.require arch >>= fun a ->
+    match Arch.of_string a with
+    | Some a -> Fact.return (Arch.addr_size a |> Size.in_bits)
+    | None -> Fact.failf "unknown/unsupported architecture" ()
+
+  let calls insns =
+    let insns =
```
ivg: bad variable naming: the are not instructions, but instructions that may affect control flow. Though it is also not right.

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+
+  method! map_sub s =
+    if Term.(has_attr s synthetic) then
+      let name =
+        List.find ~f:(fun (tid, name) -> Term.tid s = tid) names in
+      match name with
+      | None -> s
+      | Some (tid,name) ->
+        Tid.set_name tid name;
+        Ir_sub.with_name s name
```
ivg: Probably, we should move `Tid.set_name` into the `Ir_sub.with_name` function.

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+  Term.get_attr jmp address >>= fun addr ->
+  Map.find calls addr >>= fun name ->
+  target_tid jmp >>= fun tid ->
+  Some (tid,name)
+
+(** [find_externals_names program insn externals] - returns a list of
+    (tid, name) where every [tid] denotes a sythetic subroutine term
+    that must be renamed to [name]. *)
+let find_externals_names prg insns calls =
+  let jumps = List.rev @@ (new jmp_finder)#run prg [] in
```
ivg: why do you need to reverse this list? Does the order matter here?

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+          let call = Call.with_target call (Direct unq_tid) in
+          Ir_jmp.with_kind jmp (Call call)
+        | _ -> jmp
+      else jmp
+end
+
+(** [reduce prg] removes duplicated synthetic subroutines,
+    that appeared during program lifting and external calls
+    resolving. *)
+let reduce prg =
```
ivg: this function looks like a hack. Why should we introduce duplicated subroutines in the program on the first hand?

lib/bap/bap_synthetic_symbolizer.ml
```diff
@@ -0,0 +1,142 @@
...
+          else String.Map.add unq name tid, rem) in
+  let prg = (new relinker prg unique)#run prg in
+  List.fold duplicates ~init:prg
+    ~f:(fun prg tid ->
+        if is_synthetic_sub prg tid then
+          Term.remove sub_t prg tid
+        else prg)
+
+
+let resolve doc insns prg =
```
ivg: This pass looks unnecessary over-complicated to me and extremely non efficient. It should be rewritten.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -253,12 +253,27 @@ let next dis s =
 
 let stop_on = [`Valid]
 
+let relocate brancher lift mem insn =
```
ivg: This function is incorrect, first of all, second, it should exist at all. Brancher is not a relocator. It shouldn't perform relocations. 

So, first of all, why it is incorrect. You take a BIL program, that may have N destinations and an output from a brancher, that may have M destinations, and arbitrary say that all destinations of the BIL program is the first destination from the brancher output. It doesn't sound right to me. So this function should be removed.

lib/bap_image/bap_image.ml
```diff
@@ -349,10 +356,14 @@ module Derive = struct
...
 
   let arch =
-    Fact.require arch >>= fun s -> match Arch.of_string s with
-    | Some s -> Fact.return s
-    | None -> Fact.failf "unknown/unsupported architecture %s" s ()
-
+    Fact.collect Ogre.Query.(select (from arch)) >>= fun s ->
+    Fact.Seq.reduce ~f:(fun a1 a2 ->
+        if Arch.equal a1 a2 then Fact.return a1
+        else Fact.failf "arch is ambigous" ())
```
ivg: the typo is still unfixed.

lib/bap_llvm/bap_llvm_macho_scheme.ml
```diff
@@ -0,0 +1,22 @@
+open Core_kernel.Std
+open Ogre.Type
+open Bap_llvm_ogre_types.Scheme
+
+(** macho segment command *)
+let segment_cmd () =
+  Ogre.declare ~name:"segment-command" (scheme name $ off $ size)
+    Tuple.T3.create
+
+let segment_cmd_flags () =
```
ivg: missing docs.

lib/bap_llvm/bap_llvm_macho_scheme.ml
```diff
@@ -0,0 +1,22 @@
...
+  Ogre.declare ~name:"segment-command-flags"
+    (scheme name $ readable $ writable $ executable)
+    (fun name r w x -> name, (r,w,x))
+
+let virtual_segment_cmd () =
```
ivg: missing docs

PR #655, opened 2017-04-26
fix ida cmdline parameters

gitoleg:
removed a version parameter, that is considered as senseless, also constrained a `ida-paths` to allow user to specify only one additional path instead if providing a list of them.

### DIFF
--- a/plugins/ida/ida_main.ml
+++ b/plugins/ida/ida_main.ml
@@ -213,7 +213,7 @@ module Cmdline = struct
 
     let path =
       let doc = "Path to IDA directory." in
-      Config.(param (some string) "path" ~synonyms:["path"] ~doc) in
+      Config.(param (some string) "path" ~doc) in
     let headless =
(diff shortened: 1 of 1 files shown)


plugins/ida/ida_main.ml
```diff
@@ -224,23 +211,19 @@ module Cmdline = struct
         `P "$(b,bap-ida)(3), $(b,regular)(3),$(b,bap-plugin-byteweight)(1), $(b,bap-plugin-objdump)(1)"
       ] in
 
-    let paths =
+    let path =
       let doc = "Path to IDA directory." in
-      Config.(param (list string) "path" ~synonyms:["paths"]
-                ~default:[Bap_ida_config.ida_path] ~doc) in
+      Config.(param (some string) "path" ~synonyms:["path"] ~doc) in
```
ivg: a strange synonym :)

PR #663, opened 2017-05-12
added tags to bundles

gitoleg:
Now every bundle has a list of tags associated with it. 
And bap plugins subsequently also could be tagged, e.g. `x86-plugin` has the next list of associated tags: `core, lifter, disassembler`. Also it's possible to filter plugins with some particular tags.


lib/bap_plugins/bap_plugins.mli
```diff
@@ -85,30 +85,33 @@ module Std : sig
       | `Errored of string * Error.t (** failed to load a plugin    *)
     ] [@@deriving sexp_of]
 
-    (** [list ?library ()] scans a list of directories, provided by a
+    (** [list ?query ?library ()] scans a list of directories, provided by a
```
ivg: `query` is a bad name for the parameter. The better names would be `provides`, The idea is to look at the point of invocation: `list ~provides:["disassembler"; "x86"]` is much better than `list ~query:["disassembler"; "x86"]`.

lib/bap_plugins/bap_plugins.mli
```diff
@@ -85,30 +85,33 @@ module Std : sig
...
         [library] parameter, and, finally, [Config.libdir] for files with
-        a [plugin] extension.*)
-    val list : ?library:string list -> unit -> plugin list
+        a [plugin] extension. Returns all plugins that support at
```
ivg: The comment should be:

```ocaml
(** [list ~provides:features ?library () ....
     ....
     .... Returns all plugins that provide a superset of [features]. *)

File: oasis/abi [hunk 3]
```diff
@@ -18,3 +18,4 @@ Library abi_plugin
   BuildDepends:     bap, bap-abi
   InternalModules:  Abi_main
   XMETADescription: apply abi information to a project
+  XMETAExtraLines:  tags="core, pass"
```
ivg: maybe `basic` is a better name, as it provides the `basic` functionality needed by all frontends. However, you may leave the tags as they are, I'm leaving this comment for discussion only. We will rearrange the features names several time before the release. The main idea, is that the downstream users shouldn't notice, that we introduced this filter - thus all the existing plugins should be loaded automatically, thus should be in the same category.

tools/build_plugins.sh
```diff
@@ -10,7 +10,20 @@ build_plugin() {
...
         bapbuild -package bap-plugin-$plugin $plugin.plugin
-        bapbundle update -desc "`ocamlfind query -format "%D" bap-plugin-$plugin`" $plugin.plugin
+        DESC=`ocamlfind query -format "%D" bap-plugin-$plugin`
+        TAGS=`ocamlfind query -format "%(tags)" bap-plugin-$plugin`
+        if [ -z "$TAGS" ]; then
+            bapbundle update -desc "$DESC" $plugin.plugin
+        else
+            bapbundle update -desc "$DESC" -tags "$TAGS" $plugin.plugin
+        fi
+        if [ -f $plugin/resources ]; then
```
ivg: This is already removed from the master branch, check your merge. It is strange, why it adds it back.

ivg: Also, make sure, that David's fix for `mkdir` is not reverted.

lib/bap_plugins/bap_plugins.ml
```diff
@@ -223,26 +224,36 @@ module Plugins = struct
               with exn -> Some (Error (file,Error.of_exn exn))
             else None))
 
-  let list ?library () =
+  let is_selected tags p =
```
ivg: The query should be denoted via the for_all, e.g., 
```ocaml
List.for_all requested ~f:(List.mem provided)
```

oasis/abi
```diff
@@ -18,3 +18,4 @@ Library abi_plugin
   BuildDepends:     bap, bap-abi
   InternalModules:  Abi_main
   XMETADescription: apply abi information to a project
+  XMETAExtraLines:  tags="pass"
```
ivg: Are quotes required in case of a singleton set of tags?

oasis/elf
```diff
@@ -25,3 +25,4 @@ Library elf_loader_plugin
   BuildDepends:     bap, bap-elf, bap-dwarf
   InternalModules:  Elf_loader_main
   XMETADescription: read ELF and DWARF formats in a pure OCaml
+  XMETAExtraLines:  tags="disassembler"
```
ivg: why is it a disassembler?

gitoleg: because it provides a very useful feature named "copy-paste"

oasis/abi
gitoleg: I'm afraid there are. At least I have an error when omit them: 
```
ocamlfind: Cannot parse 'plugins/elf_loader/META': '=' must be followed by a string constant
```

ivg:
Please, update the testsuite to the master branch, so that I can merge it.

PR #668, opened 2017-06-20
Primus Promiscuous Machine

ivg:
This is a big update to Primus, with a severe rewiring of the whole system. 

The Primus Machine monad is now unary, that makes it much easier to work with (no more value restriction, shorter and more understandable types). It also no longer uses objects and row-typing and all context is stored using the universals. The interpreter itself is also no longer a class, and is not derived from Biri. In fact, a new evaluation component is added called Eval, that is the generic base for Expi, Bili, and Biri. The Primus Interpreter doesn't use memory storages from Bili, nor it uses Bil.result anymore, instead the plain word type is used to represent all machine values.

The main change is that the Primus Machine Monad is now a composition of Fail, Continuation, and State monads. The Fail monad enables exceptional control flow, the Continuation monad reifies program control into the first class value, and the State monad reifies the program state. The fork function clones the state and current continuation. The switch one restores previously stored state and continuation. Basically, it means that we have a co-routine style monad with backtracking and everything needed for the non-determistic and probablistic computing. 

On top of a new monad, a new mode of execution is provided, that is called "promiscuous mode". In this mode, every time we see a branching instruction we clone our machine and set up the state of the clone in a such way, that it will take a branch that is not selected by the parent machine. We are doing this on a data level, by setting variables that define the jump destination to the desired value. Of course, we do not propagate this changes back to the initial environment (this will require an SMT solver), so we just put so called conflict statements into the state, under assumption, that they can be later used (either resolved by the SMT, or just present to a user). 

The promiscuous mode, if enabled, will just clone machines, but will keep running the main machine up. An actual switching between concurrent machine clones is performed by a scheduler. We provide several schedulers, implementing different search strategies. The greedy scheduler will just perform a depth-first search of a tree of clones. The round-robin will provide an equal share of time to all clones, thus implementing a breadth-first visiting. The exploring will prioritize clones that will evaluate less visited areas of a program, the wandering scheduler will just switch between machines in a random manner.

This PR also provides a new implementation of Taint Analysis (the legacy one is still kept). The primus-propagate-taint analysis is made compatible with the old taint-analysis and should work as a drop-in replacement in IDA and Saluki. The new implementation is much more precise and tracks data dependencies on a much longer paths.


rvantonder:
Yowza.

PR #673, opened 2017-07-17
[DONOTMERGEYET] Initial draft of RISC-V in BAP

XVilka:
There is a lot of work to be done, but I think I still need a feedback. I'm afraid to go to a completely wrong direction at the beginning.


plugins/riscv/riscv_main.ml
```diff
@@ -0,0 +1,479 @@
...
+        | None -> invalid_argf "unknown register %s" name ()
+        | Some reg -> Ok reg
+
+  let imm = function
+    | Op.Reg _ | Op.Fmm _ -> Or_error.errorf "expected immediate"
+    | Op.Imm imm -> Ok imm
+
+  (** [rrr_type f d s t] uses lifter [f] for an rrr-type instruction with
+      arguments [d], [s], and [t]. *)
+  let rrr_type f d s t = match reg d, reg s, reg t with
```
ivg: maybe it is better to apply `simpl` once at the toplevel lift function just before the the value is returned? In fact, you may simply omit the `simpl` function, as I will still apply simplification/normalization on later phases. Here it is mostly useful for debugging. Probably, we should add a `--simpl/--no-simpl` and other related options to `bap-mc`.

plugins/riscv/riscv_main.ml
```diff
@@ -0,0 +1,479 @@
...
+      Option.value_exn
+        ~here:[%here]
+        ~error:(Error.create "unsupported operand" op Op.sexp_of_t )
+        (Riscv_op.create op))
+
+  let riscv_ops ops = try_with (riscv_ops_exn ops)
+
+
+  (** Substitute PC with its value  *)
+  let resolve_pc mem = Stmt.map (object(self)
```
ivg: apparently you're not using the `pc` register anywhere, thus you don't need this pass. If you will remove the `pc` register from the scope it all, it will on strengthen this position.

plugins/riscv/riscv_main.ml
XVilka: OK. Will remove it completely then.

XVilka:
Seems RISC-V support finally landed in LLVM, so I will revive this PR as soon LLVM get released with this

PR #675, opened 2017-07-17
Make llvm3.8 preferred

RabidCicada:
This is to update BAP to use llvm3.8 on modern oses :).

### DIFF
--- a/opam/opam
+++ b/opam/opam
@@ -205,10 +205,10 @@ depexts: [
         "libgmp-dev"
         "libzip-dev"
         "libcurl4-gnutls-dev"
-        "llvm-3.4-dev"
+        "llvm-3.8-dev"
         "time"
         "clang"
-        "llvm"
+        "llvm-3.8"
         "m4"
         "dejagnu"
      ]]


ivg:
So, this will break the CI, as on travis there is (was) only 3.4. We will look how it will go out. Second, the `opam/opam` file is used only for the CI, for installation there is a separate `opam-repository` with completely distinct opam files.

RabidCicada:
Ahh...I had cloned the repo, then pinned it and it seemed to use the opam/opam file.  Sorry about that.

ivg:
Thanks, I hope that at some point of time when Travis will move to xenial it will work. But so far, Travis lags behind the progress very much. They just recently moved from Precise to Trusty ... on an experimental basis. 

Maybe we will actually move from Travis to our own integration system, that will check both (all) version of llvm. 

But so far this PR is a little bit ahead of time for us :)

RabidCicada:
BTW...I just happened to separately encounter this: https://docs.travis-ci.com/user/docker/  In response to a request from Travis to support 16.04 [here](https://github.com/travis-ci/travis-ci/issues/5821).  If you want...docker can do 16.04 for you and handle the other version stuff too.  Just an option

Issue #664, opened 2017-05-15
Inconsistent bitvector signedness 


ivg:
The bitvector documentation states that 
> the signedness property won't propagate to the result of the operation

This is not true, however, as unary operators propagate the signedness property. 

It also became evident that we need the `unsigned` operator that will cast away the signed property, as well as the `is_signed` operator that will project it.

PR #680, opened 2017-07-26
Thin bitvector

ivg:
## new bitvector representation

The proposed representation uses the Z.t as a container for both: the
payload, and the metadata.

We reserve 15 bits for the metadata. The first bit represents whether a
bitvector should be treated as signed or unsigned. The rest 14 bits
contain the size of a bitvector. With such representation we can encode
only bitvectors of size 2^14 or less. This is, however, enough in
practice. If more would be needed we may switch to a more complex
prefixed encoding akin to UTF.

```
     +-----------+------+---+
     |  payload  | size | s |
     +-----------+------+---+
      size+15  15 14       0
```

Given this representation all bitvectors that are smaller than
0x1_000_000_000_000 are represented as an OCaml int, and are thus
unboxed, so creating new bitvector, and performing bitvector
arithmetic no longer performs an allocation.

```
 # let deadbeef = Word.of_int32 0xDEADBEEFl;;
 val deadbeef : Bap.Std.Word.t = 0xdeadbeef:32
 # Obj.is_int (Obj.repr deadbeef);;
 - : bool = true
```

The previous representation is still available under
`Word.Stable.V1`. It is useful for serialization and deserialization of
data stored in previous representation. In particular, we are using the
old representation to read byteweight signatures.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -1090,6 +1090,11 @@ module Std : sig
         applied on [t] will be signed *)
     val signed : t -> t
 
+
+    (** [unsigned t] casts [t] to an unsigned type, so that any
+        operations applied to it will interpret [t] as an unsigned word.*)
+    val unsigned : t -> t
+
     (** [is_zero bv] is true iff all bits are set to zero. *)
     val is_zero : t -> bool
 
@@ -1265,6 +1270,16 @@ module Std : sig
         [Width] exception if operands sizes mismatch.
     *)
     module Int_exn : Integer.S with type t = t
+
+    (** Stable marshaling interface.  *)
+    module Stable : sig
+      module V1 : sig
+        type nonrec t = t [@@deriving bin_io, compare, sexp]
+      end
+      module V2 : sig
+        type nonrec t = t [@@deriving bin_io, compare, sexp]
+      end
+    end
 
     (** Prefix trees for bitvectors.
 
@@ -3865,6 +3880,21 @@ module Std : sig
 
     (** Tries over memory  *)
     module Trie : sig
+      module Stable : sig
+        module V1 : sig
+          module R8  : Trie.S with type key = t
+          module R16 : Trie.S with type key = t
+          module R32 : Trie.S with type key = t
+          module R64 : Trie.S with type key = t
+        end
+        module V2 : sig
+          module R8  : Trie.S with type key = t
+          module R16 : Trie.S with type key = t
+          module R32 : Trie.S with type key = t
+          module R64 : Trie.S with type key = t
+        end
+
+      end
       module R8  : Trie.S with type key = t
       module R16 : Trie.S with type key = t
       module R32 : Trie.S with type key = t
--- a/lib/bap_byteweight/bap_byteweight.ml
+++ b/lib/bap_byteweight/bap_byteweight.ml
@@ -94,7 +94,7 @@ module Bytes = struct
         match Memory.view ~from ~words:length mem with
         | Ok mem -> Some mem
         | _ -> None
-    end)(Memory.Trie.R8)
+    end)(Memory.Trie.Stable.V1.R8)
 
   let find bw ~length ~threshold mem =
     let start = Memory.min_addr mem in
--- a/lib/bap_image/bap_memory.ml
+++ b/lib/bap_image/bap_memory.ml
@@ -442,19 +442,30 @@ let pp_hex fmt t =
 
 
 module Trie = struct
-  module Key(Spec : sig val size : size end ) = struct
-    open Spec
-    type nonrec t = t
-    type token = word [@@deriving bin_io, compare, sexp]
+  module Make(Token : sig
+      type t = word [@@deriving bin_io, compare, sexp]
+    end) = struct
+    module Key
+        (Spec : sig val size : size end ) = struct
+      open Spec
+      type nonrec t = t
+      type token = Token.t [@@deriving bin_io, compare, sexp]
 
-    let length m = length m / Size.in_bytes size
-    let nth_token m n = get ~index:n ~scale:size m |> ok_exn
-    let token_hash = Word.hash
+      let length m = length m / Size.in_bytes size
+      let nth_token m n = get ~index:n ~scale:size m |> ok_exn
+      let token_hash = Word.hash
+    end
+
+    module R8  = Trie.Make(Key(struct let size = `r8 end))
+    module R16 = Trie.Make(Key(struct let size = `r16 end))
+    module R32 = Trie.Make(Key(struct let size = `r32 end))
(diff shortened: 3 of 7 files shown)


ivg:
Sorry, for the pure syntactic bap_primus_interpreter.ml file. I was coding it on my laptop previously, and didn't notice, that ocp-indent wasn't functioning correctly.

ivg:
This PR also resolves #664.

maverickwoo:
I am in full support of reducing the memory consumption of BAP and this PR is very much aligned with that direction. However, I am confused at the plan of carrying out the design change:

https://github.com/BinaryAnalysisPlatform/bap/pull/680/files#diff-16c5c010300e0f3d28464302b7a9af79R39 says:

    Speaking of the sign. I would propose to remove it from the
    bitvector, as sign is not a property of a bitvector, it is its
    interpretation.

But then the PR description says:

    The first bit represents whether a bitvector should be treated as signed or unsigned.

Just like what you wrote above, I also agree that fundamentally bitvector should not have a notion of a sign bit. So I believe if we are going to change the design at such a fundamental level, it seems more suitable to eliminate this sign bit at an early stage.

But I also realize this elimination may cause a long ripple, which leads to a difficult-to-manage changeset. So perhaps this PR is only one of the multiple PRs to change the design of bitvector? If so, I'd like to know the overall game plan. (When will the sign bit be eliminated?)

Thanks!

ivg:
@maverickwoo, the problem is that we can't just remove sign, as it will break a backward compatibility. So, if we will decide to remove it at all, then we would first mark all functions such as `signed`/`unsigned` as deprecated while providing another way of handling the signedness. Then, not earlier than on bap 2.0.0 release, we will finally remove the sign. 

Speaking of the signed property in general, I'm still not 100% sure, that it shouldn't be a part of the bitvector. If we will treat a bitvector as just a representation, then we shouldn't put the size into it. I'm thinking of the size as a dynamic type, by which we overload integer operations. The same reasoning is applicable to the signed/unsigned, that can be seen as a refinement of the bitvector type, by which we overload operations in runtime.

maurer:
@ivg I assume this does not change the behavior of `Word.enum_chars` or `Word.of_binary`, since those are talking about bitvector itself and not the internal representation. (that would be my only worry - I think it's important to be able to take output or take in the actual bitvector contents)

Re: Signedness, my Rust projection of the bitvector type drops the signedness, removing it in the future seems reasonable.

maverickwoo:
@ivg Yes, I recognize the long propagation of eliminating the sign bit from bitvector as of now and so I only want to know the plan. I agree that its elimination is better done at BAP 2.

Having said that, I do notice a problem: since the old representation is called `Word.Stable.V1` and this is called `Word.Stable.V2`, then presumably this current representation (`V2`) will be superseded by a `V3` when BAP 2 is released, assuming `V3` is the representation that does not have a sign bit. This looks kind of strange, isn't it? :)

As for the sign bit itself, I remain a strong proponent to eliminate it from bitvector. Just like you said in your code comment, the sign bit is about the *interpretation* of the bitvector's content and not about the bitvector's *content* itself. While one can make a case that interpretation information can be stored together with the content itself, I still think in this case it is a bad idea because the abstraction is named "bitvector". For someone who just looked at the name, one may think a bitvector should be suitable for implementing, say, a 1024-bit Bloom Filter or some bit masks. Clearly there is no sign in these interpretations.

Regarding naming and planning, my question is: what is the plan to call the binary representation of a float, or any of the non-numerical values storable in an XMM (https://en.wikipedia.org/wiki/Streaming_SIMD_Extensions)? Will they all be called a "bitvector" too? The current documentation of bitvector in `bap.mli` says a bitvector is "a numeric value with a 2's-complement binary representation." So it seems that sticking the bits of a float64 into a bitvector is already not suitable (even though a float does have a sign bit).

As for size, if bitvector is intended to have a fixed length at creation time, then I think, if possible, making it phantom and deal with the reification at marshalling time may not be *too* bad of an idea... except for the large amount of code change. :) But if the size of a bitvector needs to change dynamically (which seems a rare use case to me---ymmv), then sure it is easier to explicitly represent and manipulate the size.

ivg:
> Having said that, I do notice a problem: since the old representation is called Word.Stable.V1 and this is called Word.Stable.V2, then presumably this current representation (V2) will be superseded by a V3 when BAP 2 is released, assuming V3 is the representation that does not have a sign bit. This looks kind of strange, isn't it? :)

No, this is a version of representation, not a version of interface. Before 2.0.0 we may change the representation quite a few time. Basically, a representation is just a format of saving the data, the serialization format. 

And the problem with removing the sign is that this is not a representation issue, but the interface one. We can't provide the same interface without changing the interface. When BAP2.0.0 will be released, the interface will be seriously refined, so there wouldn't be any V1, V2 anymore. 

> Just like you said in your code comment, the sign bit is about the interpretation of the bitvector's content and not about the bitvector's content itself.

Ok, I've got your point, but I do not agree. Maybe  your confusion comes from a bad choice of names. Indeed, bitvector is just a vector of bits. And probably in BAP.2.0.0 we will have such basic type. 

A type is not the representation. A type is the interface. Type defines the semantics of values. In our case the semantics of signed and unsigned arithmetic integers is different,  that's why we have the sign in the representation. The same as with the width. 

Of course, when we will implement floating point values, the notion of signedness will no longer have any sense. Because floating points doesn't implement integer arithmetic. The notion of the wideness will be also very different, as floats do not implement modular arithmetic. This just means, that we shouldn't implement floats using the bitvector module, that is a misnomer for an arithmetic type. 

> As for size, if bitvector is intended to have a fixed length at creation time, then I think, if possible, making it phantom and deal with the reification at marshalling time may not be too bad of an idea...

Then we will have `type 'a exp`, with a burden to prove, that an expression is well-typed and preserves the width. Moreover, our expressions may evaluated to storages and change the type. Imagine, how hard it is to implement casts and extracts. The sizes can no longer be runtime values but encoded in types. The bil type will require significant refactoring. The same is true for the IR. 

And does it worth? No it doesn't, because there is a more practical but maybe less fancy solution. We can just typecheck the code in runtime and verify that all sizes are correct. Since the main source of BIL is a lifter and we own the BIL code this invariant is very easy to preserve. At least I can always guarantee that all code in `program` is well-typed.

To summarize, what we call bitvector in BAP, is not a bitvector. `std::vector<bool>` is bitvector. What we have is an arithmetic type. When we will introduce floats, we will factor out the common part from bitvector (and probably retain the bitvector name with this common part) and use it in both floats and words, e.g.,

```ocaml
type 'a bitvector (* an array of bits and noting more *)
type word = integer bitvector  (* provides integer and logic arithmetic *)
type float = floating bitvector  (* provides floating point arithmetic *)
```

PR #682, opened 2017-07-31
reduces the memory representation RAM consumption

ivg:
As a continuation of #680, this PR will reduce the memory footprint of
the memory representation in bap, the `mem` type. As an optimization, we
were precomputing a memory accessors during the construction of a memory
object. Although, it might save some time it didn't at all, as the
memory access operations are rarely used if used at all. Thus, this PR
removes the precomputed getters from the memory, and moves all the check
to the time of an accessor invocation.

As a result, this reduces the size of memory representation by 10
words (80 bytes on amd64). In total with the #680, this will reduce
memory representation from 21 words to only 7.

### DIFF
--- a/lib/bap_image/bap_memory.ml
+++ b/lib/bap_image/bap_memory.ml
@@ -13,28 +13,12 @@ type getter = {
   fast : pos_ref :  int ref -> word;
 }
 
-module Getters = struct
-  type getters = {
-    int8  : getter;
-    uint8 : getter;
-    int16 : getter;
-    uint16: getter;
-    int32 : getter;
-    int64 : getter;
-    int128 : getter;
-    int256 : getter;
-  } [@@deriving fields]
-end
-
-open Getters
-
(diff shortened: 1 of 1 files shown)


ivg:
To summarize the effects on performance of this PR and #680 

1. Maximum resident set size (RAM), as per `/usr/bin/time -v`: -200%
2. Running time (full disassembly): +20%
3. On disk representation (size of a cache program): -100%
4. Loading from cache: -1300%

Thus the tradeoff is 20% of the first time disassembly, for a significantly reduced memory footprint, and much better cache performance. 

If you're interested in absolute values, then I was using a 2Mb file, that was using around 2.5 Gb of RAM before #680, and is using around 750 Gb with this PR (giving us x300 ram overhead, and x100 disk overhead). Note, the memory consumption is not measured by GC, but is reported by /usr/bin/time ,thus the real consumption can be smaller.

PR #688, opened 2017-08-11
Final 1.3 update

ivg:
This PR brings multiple new features from our development branches. In general this PR focuses on three main features:

1. BIL correctness and normalization
2. BIL readability and performance
3. Primus interpreter

Note for reviewers and readers: I wouldn't expect you guys to check the code thoroughly, but you are welcome to ask questions, request clarifications, and discuss changes. Most likely just reading this PR message would be enough for the review. 

Note: the Table of Contents doesn't work since Github doesn't emit anchors on the PR page. They will work (I hope so), when we will move this page to the release page. 

Table of Contents
=============

* [Detailed description](#detailed-description)
    * [BIL Typechecker](#bil-typechecker)
    * [BIL effect analysis](#bil-effect-analysis)
    * [New BIL simplification and constant propagation algorithm](#new-bil-simplification-and-constant-propagation-algorithm)
    * [BIL normal form and normalization utilities](#bil-normal-form-and-normalization-utilities)
    * [SSA transformation plugin](#ssa-transformation-plugin)
    * [Dead code elimination plugin](#dead-code-elimination-plugin)
    * [New pretty printer for bitvectors](#new-pretty-printer-for-bitvectors)
    * [New functions for bitvectors](#new-functions-for-bitvectors)
    * [New pretty printer for BIL](#new-pretty-printer-for-bil)
    * [Identifiable values for the Primus interpreter](#identifiable-values-for-the-primus-interpreter)
    * [New Primus interpreter](#new-primus-interpreter)
    * [New Primus Value representation](#new-primus-value-representation)
    * [New Primus Mark Visitied plugin](#new-primus-mark-visited-plugin)
    * [Diagnostic messages with better backtraces](#diagnostic-messages-with-better-backtraces)
    * [Fixes and updates for the x86 lifter](#fixes-and-updates-for-the-x86-lifter)
    * [Bug fixes in ARM lifter](#bug-fixes-in-arm-lifter)
    * [Support for tracing and other minor fixes](#support-for-tracing-and-other-minor-fixes)
* [Caveats, Limitations, and Regressions](#caveats-limitations-and-regressions)
* [TODO](#todo)

# Detailed Description

## BIL Typechecker

We added a type checker, and now force all BIL code to be well-typed. Lifters must produce only well-typed code. So far, we will issue a warning for each piece of code that is not, along the 1.4 we are planning to fix all issues if any. The `Type` module now have the `infer` function, that will infer type for an expression, and `check` that will type check a BIL program.

## BIL Effect Analysis

A precise effect analysis was added to ensure correctness of program transformations. The effect analysis distinguishes between different effects and coeffects:
   - reads (coeffect)
   - loads (coeffect)
   - writes (effect)
   - raises (effect)

The raises effect includes the division by zero CPU exception and a conservative AI will infer a possible presence of the division by zero in a BIL expression. 
A new module, called `Eff` is added, that provides functions for the effect analysis. 

BIL Simplification and Constant Propagation
-------------------------------------------------------

The old constant propagation and simplification algorithm was buggy, and didn't take into account effects when removing statements and subexpressions. It was rewritten from scratch with an algorithm that is sound and parametrized by the set of effects, that may be ignored during the analysis. The new analysis treats operations accurately with respect to their algebraic properties and takes into account operations associativity. It is also now a big step analysis and no longer requires to be wrapped into the `fixpoint` function. Basically, we are moving from the open and semi-open recursion to the closed recursion wherever it is possible, as it is more efficient and intuitive approach. We will probably remove the `fixpoint` function in BAP.2.0.

BIL normal form and normalization utilities
-----------------------------------------------------

We introduce a notion of a BIL Normal Form, that is a sub-language of BIL that disallows certain constructions, such as `let` and `if/then/else` expressions, complex memory expressions, and complex conditionals. This doesn't impose, that those expressions are not allowed in general, and lifters can still emit code that is easier to write. All such expressions are removed, i.e., are rewritten during the normalization process. The normalization procedure is described in details in the documentation, as well as in comments. Here is a brief highlight:

1. no let expressions
2. no if/then/else expressions
3. all memory operations are one-byte only (size and endianness can be ignored)
4. no anonymous memory storages - load expressions may load only from a variable
5. no loads and stores inside conditional expressions
6. no ill-typed expressions

SSA transformation plugin
---------------------------------

This is a simple plugin that just transfers all functions into SSA form. It was moved from the bap-plugins repository.

Dead code elimination plugin
-------------------------------------

A sound analysis that will remove all unused code. This plugin is very useful, as it will remove lots of unnecessary code that are emitted by lifters. In particular, on x86 it removes flag compuations that are not used anywhere. As a result, this reduces the size of the output to 50% on x86 and makes BIL much more readable. It also highly increases the speed of interpretation. Especially, since many flag computations are quite computationally hard. 

New pretty printer for bitvectors
----------------------------------------

Printers from the Z library proved to be slow, buggy, and ugly. Thus we decided to provide our own pretty-printer that will suit all the needs. A generic printer is added, that allows a user to customize the look of the textual representation of a bitvector. We also provide 9 different instantiations of the generic printer, e.g., `pp_hex`, `pp_dec`, `pp_oct` and `pp_bin` that will print a bitvector in the corresponding format. We also remove suffixes from a bitvector to make output more readable and concise. The full version is available via `pp_hex_full`, `pp_dec_full`, ... printers. We also change a default textual representation (i.e., to_string and pp) functions. The new representation now includes the signedness suffix, and moves away from using `true` and `false` for the one-bit words. 

New functions for bitvectors
----------------------------------

We added the `to_int[XX]_exn` family of functions, that will efficiently cast a bitvector into a corresponding OCaml type without extra allocations whenever possible. For example, `Word.to_int_exn` is usually an identity function w.r.t. the GC (i.e., it still has a check that value fits, and if it is, the it is an identity). 

We also added the `Word.unsigned` function that will cast away the signedness of a bitvector.

The `Unsafe` module was added to the bitvector interface that provides arithmetic instructions that do not check that operands have correct widths. The module can be used for interpreting a well-typed BIL code. Moreover, these operations are included by default in the Bitvector top-level interface, thus, a `Word.(a + b)` will result in the unsafe `(+)` operator.

New pretty printer for BIL
--------------------------------

We implemented a new pretty printer that (i) doesn't emit unnecessary parentheses anymore, and uses more concise syntax whenever it is possible. We followed the C language for the operator precedence (not OCaml). Here is the precedence table:

```
   +-------------------------------------------+----+
   | x[y], <cast>:<N>[x], extract:<N>:<M>[x]   | 10 |
   +-------------------------------------------+----+
   | ~x, -x                                    |  9 |
   +-------------------------------------------+----+
   | *,/,/$, %, %$                             |  8 |
   +-------------------------------------------+----+
   | +,-                                       |  7 |
   +-------------------------------------------+----+
   | <<,>>,~>>                                 |  6 |
   +-------------------------------------------+----+
   | <,>,<=,>=                                 |  5 |
   +-------------------------------------------+----+
   | =, <>                                     |  4 |
   +-------------------------------------------+----+
   | &                                         |  3 |
   +-------------------------------------------+----+
   | ^                                         |  2 |
   +-------------------------------------------+----+
   | |                                         |  1 |
   +-------------------------------------------+----+
   | let.., x with y <- z, if.., . (dot)       |  0 |
   +-------------------------------------------+----+
```
1) Higher precedence means that operators binds tighter.
2) Any bil statement has precedence lower than an expression

The load and store expressions are also simplifed, now it is not necessary to put the endiannes and size if an expression is byte sized. 

Identifiable values for the Primus interpreter
-------------------------------------------------------

We used the word type for values in the Primus interpreter, however, it showed that it is very hard to write a data tracking analysis without being able to identify values. For example, in the taint propagation plugin, we implemented a stack machine for tracking the taint. It is not only hard to maintain and understand, but also is inefficient. This defeats the main reason why we decided to use the word type as  a value representation in the interpreter -- the efficiency. 

We now represent a computation result as a pair of a word and a unique identifier. Basically the same as with the Bil.Result.t except that we are not using a variant type to represent storages and bottom values. In the Primus model, there are no such values, and everything is represented with a machine word, coined with an identifier. 

To make it easier to work with the new values, we lifted most of the bitvector interface into the value interface. 

New Primus interpreter
-----------------------------

The Primus Interpreter is reimplemented and doesn't use the eval or microx code anymore (though there is still lots of code sharing since we factorized a lot of code from microx to the BIL library). The interpreter now tracks data much more closely, and provides an interface that allows code summaries, that are implemented in Lisp or in OCaml or in any other language to describe all code effect precisely. We also revisited observations that are made by the interpreter, and moved memory and environment observations into the interpreter. 

New Primus Mark Visitied plugin
-----------------------------------------

This plugin will mark each term that is visited (evaluated) by the Primus interpreter with the `visited` attribute. It is useful for tracking the coverage. 

Diagnostic messages with better backtraces
--------------------------------------------------------

There is an impedance mismatch between OCaml exceptions and the error monad. Sometimes we switch from one to another and loose the backtracing information. Now the backtraces are always stored in `Or_error.t` values and are printed in case of failure. 

Caveats, Limitations, and Regressions
===============================

1. The normalized form is not represented in a type system, thus it maybe applied more than once. In future we are planning to add phantom types for that. 

2. The dead-code elimination plugin is computationally intensive but is not cached. Thus the startup time of each binary is actually increased. We will later and the caching layer to alleviate this. 

Fixes and updates for the x86 lifter
--------------------------------------------

The shift operation representation was slightly incorrect, as it was using `x := x` instruction to represent that the value of `x` is unchanged. From the analysis perspective this is not entirely true, as it is an update of a variable, an unnecessary usage (that keeps DCE from removing this statement) and, in many cases, it left the right hand side left variable upper exposed, that was introducing a source of undefined values, as well as prohibited certain kind of optimizations on the interprocedural values (we were forced to treat `x` as an implicit input parameter of a function, where it is upper exposed). 

The shift lifter also introduced unnecessary temporal variables, as well as used lots of ite. The former is removed, the latter is translated into one if statement (basically all ite were guarded by the same condition, so we can gather them, that will lead to much better CFG).

Bug fixes in ARM lifter
---------------------------

The ARM lifter was emitting badly typed code. Now it is fixed.

Support for tracing and other minor fixes
----------------------------------------------------

We added a handler for SIGINT so that gprof and other utilites, that requires a correct exit call, can work properly. We also added BAP_DEBUG=true to the Travis environment to ease the debugging on it. 

TODO
=====
Note, this is a work in progress, there are still a few minor issues that should be resolved before this branch would be ready for the merge. Here comes a list

- [x] Mark all new functions with the `@since` tag
- [x] Document all new functions
- [x] update all lifters with the typechecking code
- [x] remove calls to Bil.fixpoint as it is no longer necessary (in most cases)
- [x] check that all new plugins have sufficient manpages and documentation
- [x] check that all usages of Word.string_of_value are consistent, so that we will not break any integrations with external tools
- [ ] update README.md with nice examples of BIL. Ideally, use the well-known strcpy as an example.
- [x] rewrite observer notifications so that flambda can inline them
- [x] cleanup the observations, make sure that we don't have any dead observation handlers.


maverickwoo:
* Do you have a few examples of ill-typed BIL from the previous lifters? Such examples (perhaps as a comment in this PR discussion) can be very helpful to future lifter writers.

* Since the pretty-printer from `Z` is buggy, are you planning to file (or have you already filed) a bug upstream? This seems like a good idea for the community at large.

* Is there a way to preview the documentation? I would have guessed that https://github.com/ivg/bap/blob/final-1.3-update/README.md would point to https://binaryanalysisplatform.github.io/bap/api/v1.3.0/argot_index.html but then I realize that the merge needs to happen first since the documenation generation is automatic? I guess that means technically there is no way to preview the documentation? (Or is the documentation update is going to be a separate PR?)

* Eventually we will need to have an easy method to decide which feature is in the `bap-plugins` repo and which feature is in the `bap` repo. This is not super important, but you can label me confused.

* I like the idea of normalized BIL. In fact, for some time in the future, I suggest we have a table for each architecture, one opcode per row, showing the normalized BIL for that opcode. (I know the x86 lifter wasn't really organized by a standardized opcode like the ARM lifter does, but perhaps over time the x86 lifter will use more of LLVM MC, and so eventually a table row is just an MCOpcode.)

maurer:
I'm mildly familiar with effect type theory, but have never seen coeffect type theory before. I can guess the general idea (coeffects depend on the state of the context; effects change the state of the context), but can't seem to find any literature on coeffects other than a single paper which basically says "coeffects are what we'll call comonoidal actions".

Is all you mean by coeffect here "this is comonoidal", or do you mean something more specific? If something more specific, is there a paper I could read?

maurer:
With regards to BIL Normal Form, I just wanted to mention that breaking reads/writes down to bytewidth-only does two things that may be potentially unhelpful:
* If BAP ever decides to deal with multithreaded programs, this changes the semantics of the instruction.
* If trying to perform a variable detection scheme, certain heuristic queues as to variable boundaries will be erased (e.g. 4 byte accesses vs a 32-bit integer access on the stack).

I'd suggest that the dead code elimination plugin, if possible, should rewrite indirect jumps to have CFI-or-abort in order to bake in the assumption of a complete CFG you need to say that these computations are not used anywhere. This would make both make the soundness claim stronger, and provide a catchable error that would allow the user to extend the CFG and re-run whatever they were doing.

ivg:
> I'm mildly familiar with effect type theory, but have never seen coeffect type theory before. I can guess the general idea (coeffects depend on the state of the context; effects change the state of the context), but can't seem to find any literature on coeffects other than a single paper which basically says "coeffects are what we'll call comonoidal actions".

Yep, your guess is correct, that's just effects are categorical dual of effects - the arrows are in different directions. (Co)Effect analysis is used in OCaml compiler to determine whether an expression is movable, eresable, etc. Concerning the papers: [1][1], [2][2].

[1]: http://www.cl.cam.ac.uk/~dao29/publ/coeffects-icfp14.pdf
[2]: https://www.cl.cam.ac.uk/~dao29/publ/haskell14-effects.pdf

ivg:
> If BAP ever decides to deal with multithreaded programs, this changes the semantics of the instruction.

Not really. When we will start to deal with the multithreaded programs, we will add more statements to BIL, that will deal with the concurrency semantics, as it is not true in general that an instruction is atomic. Thus we would eventually add something like `Atomic of bil` to BIL. 

An in any case, BIL Normal Form is just a form. It is not mandatory. 

> If trying to perform a variable detection scheme, certain heuristic queues as to variable boundaries will be erased (e.g. 4 byte accesses vs a 32-bit integer access on the stack).

Well, first of all it is not mandated, as I told before. Second, it is still one BIR statement, e.g., 
```
R0 := mem[10, el]:u32
```
is normalized to:
```
R0 := mem[13] . mem[12] . mem[11]. mem[10]
```

Sometimes it is easier to work with the latter, sometimes it is easier to work with the former. 


> I'd suggest that the dead code elimination plugin, if possible, should rewrite indirect jumps to have CFI-or-abort

Well, I see your point, but I can't express this CFI-or-abort in terms of BIL, that's the problem. 

And yes, the analysis is, as always, sound under the assumption that CFG is sound. We over-approximate the interprocedural dependencies by assuming that any function may call any other function, thus any upper-exposed variable in any function is preserved by all other functions.

maurer:
It would still change the possible ways in which non-atomic instructions can interleave - normally, a 32-bit write will either all happen or all not happen, not partially happen. This isn't an issue if we just want to declare we're going to rework the memory semantics of BIL when we add multithreading (we'd probably have to), but we would admit additional forms of data races which did not exist before under this normalization.

As far as the CFI, it wouldn't be perfect, but we could reserve negative CPU exceptions for purposes of BIL breakouts. You could imagine adding a chained conditional that checks that the target address is in the list you expect, and if it is not, does CpuExn (-1). Intel only uses exceptions up to 31, and we're using a full OCaml int here, so we have plenty of space.

maverickwoo:
Let's have some preliminary discussions about the memory model for multithreading here.

1. For this PR alone, is it true that if a program does not have a race condition, then the proposed normalization can never introduce a race? (Matt's concern seems to be that if there is already a race, then there are more possible races in the normalized BIL.)

2. It seems that for a language like BIL---one that breaks one machine instruction into multiple effectful statements, we can never safely allow the interleaving of statements from two separate instructions unless we rely on either (i) extensive hygiene by anyone who produces BIL---not a good idea in general, or (ii) some new syntactic construct to declare a sequence of BIL statements is atomic. Are there alternative designs to these two?

ivg:
> As far as the CFI, it wouldn't be perfect, but we could reserve negative CPU exceptions for purposes of BIL breakouts. You could imagine adding a chained conditional that checks that the target address is in the list you expect, and if it is not, does CpuExn (-1). Intel only uses exceptions up to 31, and we're using a full OCaml int here, so we have plenty of space.

OK, maybe I'm not understanding your idea. This is not that we can't express abort in BIR/BIL, we indeed can just raise a CpuExn, or just `call @abort`. The problem that I don't know to solve is how to express the CFI property, basically, how I will express in BIL that it is violated?

ivg:
> For this PR alone, is it true that if a program does not have a race condition, then the proposed normalization can never introduce a race?

Current BIL specification mandates only a single thread of execution for a BIL program, so now race conditions are possible. Neither in normalized or de-normalized program. If we will, arbitrary, say, that a BIL statement is atomic, then the memory expanding (both stores and loads) will not introduce any new potential for the race condition, since it doesn't split a statement into multiple statements. It is still one statement.  However, there are other transformations, that may affect it, namely hoisting complex memory expressions from the while/if conditionals and from the loads (i.e., the following operation `y :=(mem with a <- x)[b]` will be translated to two statements `mem := mem with a <- x; y := mem[b]`). Of course, this will have different semantics, as we move effectful expressions. But nor a lifter should produce such code, neither they do. This is just a problem with BIL that allows to express programs that are not implementable in real hardware. 

> It seems that for a language like BIL---one that breaks one machine instruction into multiple effectful statements, we can never safely allow the interleaving of statements from two separate instructions unless we rely on either (i) extensive hygiene by anyone who produces BIL---not a good idea in general, or (ii) some new syntactic construct to declare a sequence of BIL statements is atomic. Are there alternative designs to these two?

First of all, instructions are not atomic in general. That's why we have the `LOCK` prefix in x86. A simple `inc [eax]` is not atomic. While we were increment the value, another core wrote a different value. The problem is that we can express `inc` as one BIL statement  `mem := mem with eax <-
 mem[eax] + 1`

That basically tells us, that not only instructions are non-atomic, but also BIL statements and even BIL expressions. As a result, introducing a synchronization primitive on the statement level is not enough. I'm not sure whether we can solve it by adding the `Atomic of exp list` construct. In general, we should just revise BIL, and remove effectful such as Store and Load from the expression language.

maverickwoo:
OK, thanks for the clarifications. I now understand that BAP 1 can handle 1 thread and BAP 2 can handle 2 threads. :)

maurer:
Yeah, what you wrote there was exactly what I was suggesting. We could even package it as a separate pass that you do before any analysis that assumes a complete CFG. This would get us two wins:

* If `__cfi_abort` or whatever we call it gets called, we can use what it got called with to revise/expand the CFG
* When running code that has been deadcode-eliminated, there is no uncertainty over whether the execution is viable - any deadcode-eliminated trace that doesn't include a call to `__cfi_abort` is definitely equivalent to a trace of the original program.

ivg:
Yeah, @maurer, I think it would be a good idea to have this pass. The main concern is that it will by default clobber the output. We can alleviate this by rewriting the test like this:
```

when r0 = a0 call r0
when r0 = a1 call r0
...
when r0 = aN call r0
call __cfi_abort with noreturn
```

It will be easier to analyse, and easier to eliminate clauses that are not feasible. 

Yeah, looks nice, we will add this cfi-or-abort pass. But let's it do later, after the release.

Issue #643, opened 2017-03-28
LLVM Version checked against 3.4, does not detect 3.4.x correctly


maurer:
If the llvm version is not explicitly supplied, and is instead supplied by an `llvm-config` binary, it will sometimes specify the minor revision as well. This leads code like [this](https://github.com/BinaryAnalysisPlatform/bap/blob/f871c5de2e6cbcabad30efb6cae90f3ab2f8332e/plugins/x86/x86_mov_offset.ml#L198) to evaluate incorrectly and use 3.8 logic instead.

PR #700, opened 2017-08-24
support of  llvm 4.0

gitoleg:
This PR introduces a  support of `llvm 4.0.0`. 
Some other features:
1) Added `--llvm-version` option, that will print version, used for building `bap-llvm` library
2) relocatable brancher moved to a separate plugin
3) fix https://github.com/BinaryAnalysisPlatform/bap/issues/643 , since `llvm-version` internally represented only by two digits
4) some minor cleaning in `bap-llvm` library


lib/bap_llvm/bap_llvm.ml
```diff
@@ -1,10 +1,14 @@
+open Core_kernel.Std
+
+let strip_version ver =
```
ivg: this will break in 2 years, when llvm 10 will be released :)

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -204,28 +204,29 @@ void symbol_entry(const ELFObjectFile<T> &obj, const SymbolRef &sym, ogre_doc &s
     }
 }
 
-#if LLVM_VERSION_MAJOR == 3 && LLVM_VERSION_MINOR == 8
+#if LLVM_VERSION_MAJOR == 3 && LLVM_VERSION_MINOR == 8          \
+    || LLVM_VERSION_MAJOR == 4 && LLVM_VERSION_MINOR == 0
```
ivg: Per LLVM devs the MINOR version will always be 0. Its only the patch version that will be updated during the release intermezzo. I think that we can safely ignore the patch version though.

plugins/x86/x86_mov_offset.ml
```diff
@@ -194,6 +195,12 @@ end
 module T_34 = Make(Ver_34)
 module T_38 = Make(Ver_38)
 
+module Self = Self ()
+
 let () =
-  if Bap_llvm_disasm.version = "3.4" then T_34.register ()
-  else T_38.register ()
+  if llvm_version = "3.4" then T_34.register ()
+  else if llvm_version = "3.8" || llvm_version = "4.0" then T_38.register ()
```
ivg: This looks inconsistent, where is the T-40?

PR #717, opened 2017-09-28
Fix test env

gitoleg:
1) split tests. so travis will launch either `bap` internal tests or testsuite tests
2) add dependency from `conf-bap-llvm` to bap local opam file


ivg:
3. removes BAP_DEBUG to disable verbose output

Apparently this led to the 'no disk space' error. In case for debugging purpose the variable can be enabled temporary.

PR #724, opened 2017-10-05
fixed xadd insn

gitoleg:
First of all, the original BIL code confused source and destination operands.                                                                                                                                                                                                           
E.g., `xadd [eax], ebx` stores the result in the `ebx` register, although it                                                                                                                                                                                                             
is correct to store it in the memory cell `[eax]`.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                 
The second error was the wrong calculation of flags:                                                                                                                                                                                                                                   
`xadd` swaps destination and source contents, then store sum of them in destination.                                                                                                                                                                        And this was the cause of the problem, those operands were needed for the flags calculation.                                                                                                                                                                                     That's why the new BIL code stores operands values in temporary variables and then uses                                                                                                                                                                        them for flags calculation.

### DIFF
--- a/plugins/x86/x86_lifter.ml
+++ b/plugins/x86/x86_lifter.ml
@@ -1263,14 +1263,19 @@ module ToIR = struct
           assn reg32_t o_rax (Bil.Ite (equal_v, eax_e, dst_low_e));
           assn reg32_t o_rdx (Bil.Ite (equal_v, edx_e, dst_hi_e))
         ]
-      | Xadd(t, dst, src) ->
-        let tmp_res, tmp_dst, tmp_src = tmp t, tmp t, tmp t in
-        Bil.Move (tmp_src, op2e t src)
-        :: Bil.Move (tmp_dst, op2e t dst)
-        :: Bil.Move (tmp_res, Bil.(op2e t dst + op2e t src))
-        :: assn t src (op2e t dst)
-        :: assn t dst (Bil.Var tmp_res)
-        :: set_flags_add !!t (Bil.Var tmp_dst) (Bil.Var tmp_src) (op2e t dst)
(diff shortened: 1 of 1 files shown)


plugins/x86/x86_lifter.ml
```diff
@@ -1264,12 +1264,13 @@ module ToIR = struct
           assn reg32_t o_rdx (Bil.Ite (equal_v, edx_e, dst_hi_e))
         ]
       | Xadd(t, dst, src) ->
-        let tmp = tmp t in
-        Bil.Move (tmp, Bil.(op2e t dst + op2e t src))
+        let tmp_res, tmp_dst, tmp_src = tmp t, tmp t, tmp t in
+        Bil.Move (tmp_src, op2e t src)
```
ivg: 1. Do not use `Bil` data constructors to create BIL expressions and statements. The should be considered as private. We will make them private at some point of time.

2. Try to write code that looks like pseudocode, the C-like pseudocode. Basically someone without knowledge of OCaml should understand this code. I understand that it is hard in this particular case without rewriting everything from scratch, but sticking with using BIL EDSL and keeping to the most possible syntax (even if it impacts the performance) would be the right first steps. For example,

```ocaml
| Xadd (t, dst_op, src_op) -> 
   let tmp_src = tmp t in
   let tmp_dst = tmp t in
   let tmp_res = tmp t in 
   let src = op2e src_op in
   let dst = op2e dst_op in
   Bil.[
     tmp_src := src;
     tmp_dst := dst;
     tmp_res := dst + src;
     assn t src_op dst;
     assn t dst_op (var res);
  ] @ set_flag_add !!t (Bil.var tmp_dst) (Bil.var tmp_src) dst
```

This is the first approximation, but still is much more readable.

Issue #706, opened 2017-09-19
Rotate instructions produce bogus self assignments


ivg:
I thought, that I've fixed this in 1.3 but apparently I missed the case with the rotate instructions (`Rotate`), that produces the bogus `x := x` assignments (should be fixed the same way as I did with the `Shift` instruction).

```
$ bap-mc "c000ff" --show-bil --arch=X86 --show-insn=asm --x86-lifter=legacy
rolb $0xff, (%eax)
{ 
  orig_count1 := 0x1F
  mem := mem 
             with [EAX] <- mem[EAX] << orig_count1 | mem[EAX] >> 8 - orig_count1
  if (orig_count1 = 0) {
    CF := CF
  }
  else {
    CF := low:1[mem[EAX]]
  }
  if (orig_count1 = 0) {
    OF := OF
  }
  else {
    if (orig_count1 = 1) {
      OF := CF ^ high:1[mem[EAX]]
    }
    else {
      OF := unknown[OF undefined after rotate of more then 1 bit]:u1
    }
  }
}
```

gitoleg:
fixed in https://github.com/BinaryAnalysisPlatform/bap/pull/725

PR #725, opened 2017-10-06
Fixing rotate instructions

gitoleg:
According to the manual, `ror/rol` instructions has
the following calculation of the number of bit positions:
```
tempCOUNT ← (COUNT & COUNTMASK) MOD SIZE
```
We didn't  take a `mod size` in x86 lifter, while we should.

Also, BIL contains weird assignments like `OF := OF`,
when the flag is unchanged, so we will get rid of them.
(fix for https://github.com/BinaryAnalysisPlatform/bap/issues/706)
Finally, rotate instructions look like the following:
```
bap-mc "c000ff" --show-bil  --show-asm --arch=X86
rolb $0xff, (%eax) 
{
  orig_count1 := 7
  mem := mem
             with [EAX] <- mem[EAX] << orig_count1 | mem[EAX] >> 8 - orig_count1
  if (orig_count1 = 0) {
    CF := low:1[mem[EAX]]
  }
  else {
    if (orig_count1 = 1) {
      OF := CF ^ high:1[mem[EAX]]
    }
    else {
      OF := unknown[OF undefined after rotate of more then 1 bit]:u1
    }
  }
}
```

### DIFF
--- a/plugins/x86/x86_lifter.ml
+++ b/plugins/x86/x86_lifter.ml
@@ -923,46 +923,42 @@ module ToIR = struct
           ] []
         ]
       | Rotate(rt, s, dst, shift, use_cf) ->
-        (* SWXXX implement use_cf *)
         if use_cf then unimplemented "rotate use_vf";
-        let origCOUNT = tmp ~name:"orig_count" s in
+        let count_var = tmp ~name:"orig_count" s in
+        let count = Bil.var count_var in
         let e_dst = op2e s dst in
-        let shift_val = match s with
-          | Type.Imm 64 -> 63
-          | _ -> 31
-        in
-        let s' = !!s in
-        let e_shift = Bil.(op2e s shift land int_exp shift_val s') in
-        let size = int_exp !!s s' in
+        let shift_mask = match s with
+          | Type.Imm 64 -> int_exp 63 !!s
+          | _ -> int_exp 31 !!s in
+        let size = int_exp !!s !!s in
+        let e_shift = Bil.((op2e s shift land shift_mask) mod size) in
+        let lsb_dst = Bil.(cast low 1 e_dst) in
+        let msb_dst = Bil.(cast high 1 e_dst) in
         let new_cf = match rt with
-          | LSHIFT -> Bil.(Cast (LOW, !!bool_t, e_dst))
-          | RSHIFT -> Bil.(Cast (HIGH, !!bool_t, e_dst))
+          | LSHIFT -> lsb_dst
+          | RSHIFT -> msb_dst
           | _ -> disfailwith "impossible" in
         let new_of = match rt with
-          | LSHIFT -> Bil.(cf_e lxor Cast (HIGH, !!bool_t, e_dst))
-          | RSHIFT -> Bil.(Cast (HIGH, !!bool_t, e_dst) lxor Cast (HIGH, !!bool_t, e_dst lsl int_exp 1 s'))
+          | LSHIFT -> Bil.(cf_e lxor msb_dst)
+          | RSHIFT -> Bil.(msb_dst lxor (msb_dst lsl int_exp 1 !!s))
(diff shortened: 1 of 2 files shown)


plugins/x86/x86_lifter.ml
```diff
@@ -923,46 +923,42 @@ module ToIR = struct
...
           | _ -> disfailwith "impossible" in
-        let result = Bil.(ret1 lor ret2) in
-        [
-          Bil.Move (origCOUNT, e_shift);
-          assn s dst result;
-          (* cf must be set before of *)
-          Bil.Move (cf, ifzero cf_e new_cf);
-          Bil.Move (oF, ifzero of_e Bil.(Ite (Var origCOUNT = int_exp 1 s', new_of, unk_of)));
-        ]
+        Bil.([
```
ivg: This is again unreadable. Too OCamlish. 
1) Don't use the bang-bang operator
2) Try not to use pattern matching where it is possible
3) Build the specification using the following pattern:
```
Insn(p1,p2,p3,p4,...) ->
  <let-bindings-of-all-variables>
   Bil.[
    <pseudocode>
   ]
```

Try not to use BIL pseudocode in the `<let-bindings>` section, try to keep it to simple OCaml. Move as much as possible BILish expression into the last section

    


```ocaml
Rotate(rotation_type, exp_type, dst_op, shift_op, use_cf) ->
  (* sanity checks - as soon as possible *)
  if use_cf then unimplemented "rotate use_vf";

  (* define environment first *)
  let word_size = bitwidth_of_type exp_type in
  let mask_size = word_size - 1 in

  (* now declare variable *)
  let count = tmp exp_type in

  (* now declare a mapping between operands/environment and expressions *) 
  let zero = int_exp 0 exp_type in
  let one = int_exp 1 exp_type in
  let dst = op2e exp_type dst_op in
  let shift = op2e exp_type shift_op in
  let shift_mask = int_exp mask_size word_size in
  let size = int_exp word_size word_size in 

Bil.[
   count := (shift land shift_mask) mod size;
   if shift_type = LSHIFT
   then assn exp_type dst (dst lsl count) lor (dst lsr (size - count))
   else assn exp_type dst (dst lsr count) lor (dst lsl (size - count));
   if_ (count = zero) [
      cf := if shift_type LSHIFT then cast low 1 dst else cast high 1 dst;
   ] (*else*) [
     if_ (count = one) [
        of := if shift_type = LSHIFT 
         then var cf lxor cast high 1 dst
         else cast high 1 dst lxor (cast high 1 dst lsl one)
     ] (* else *) [
        of := unknown "undefined after rotate" bool_t
     ]
   ]
]
```

Alternatively, we can hoist the OCaml `if` out of the BIL:

```ocaml
if shift_type = LSHIFT
then Bil.[
   count := (shift land shift_mask) mod size;
   assn exp_type dst (dst lsl count) lor (dst lsr (size - count));
   if_ (count = zero) [
      cf :=cast low 1 dst;
   ] (*else*) [
     if_ (count = one) [
        of := var cf lxor cast high 1 dst;
     ] (* else *) [
        of := unknown "undefined after rotate" bool_t
     ]
   ]
]
else Bil.[
   count := (shift land shift_mask) mod size;
   assn exp_type dst (dst lsr count) lor (dst lsl (size - count));
   if_ (count = zero) [
      cf := cast high 1 dst;
   ] (*else*) [
     if_ (count = one) [
        of := cast high 1 dst lxor (cast high 1 dst lsl one)
     ] (* else *) [
        of := unknown "undefined after rotate" bool_t
     ]
   ]
]
```

So, even with our super ugly `assn` and `int_exp` and other "helpers" the code is already readable.

ivg: I personally think that hoisting if in this case is not necessary if not counter-productive. In the first version we can easily compare the differences between the left and right shifts.

PR #726, opened 2017-10-06
fixing `pshufb` insn

gitoleg:
The `pshufb` instruction was implemented previously as one expression
that has lots of nested if-then-else expressions.
It's possible to write a more readable code using a loop.

And if we compare a previous implementation with a new one, a difference
is quite noticeable:

```
bap-mc "660f3800042542424242" --show-bil --arch=X86 | wc --lines
327678
```
```
bap-mc "660f3800042542424242" --show-bil --arch=X86 | wc --lines
165
```

Also, the previous implementation didn't check a memory boundary in case of 128-bit operands.

### DIFF
--- a/plugins/x86/x86_lifter.ml
+++ b/plugins/x86/x86_lifter.ml
@@ -778,46 +778,46 @@ module ToIR = struct
           | 64 -> 3
           | 128 | 256 -> 4
           | _ -> disfailwith "invalid size for pshufb" in
+	let loop f = List.concat @@ List.init (op_size / 8) ~f in
 
-        let byte = int_exp 8 op_size in
         let src = op2e exp_type src_op in
         let dst = op2e exp_type dst_op in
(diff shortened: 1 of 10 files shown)


plugins/x86/x86_lifter.ml
```diff
@@ -772,34 +772,53 @@ module ToIR = struct
...
+        let mask_byte_i = tmp exp_type in
+        let tmp_byte = tmp exp_type in
+        let tmp_dst = tmp exp_type in
+        let msb = tmp exp_type in
+        let ind = tmp exp_type in
+
+        let check_mem_allignment = match src_op with
+          | Oaddr addr when (op_size = 128) ->
+            let zero = Bil.int (Word.zero 4) in
+            [ Bil.(if_ (extract 3 0 addr <> zero) [cpuexn 13] []) ]
```
ivg: use the bit_and operation to check for the alignment.

PR #733, opened 2017-10-26
fixed minor bugs

gitoleg:
forgotten parentheses in shift instructions

compare with bap.1.2.0:

```
  match st with
  | LSHIFT -> Bil.(Cast (LOW, !!bool_t, Var origDEST lsr (size - Var origCOUNT)))
  | RSHIFT | ARSHIFT ->
         Bil.(Cast (HIGH, !!bool_t, Var origDEST lsl (size - Var origCOUNT)))
  | _ -> failwith "impossible"
```

### DIFF
--- a/plugins/x86/x86_lifter.ml
+++ b/plugins/x86/x86_lifter.ml
@@ -886,9 +886,9 @@ module ToIR = struct
           (* undefined for SHL and SHR instructions where the count is greater than
              or equal to the size (in bits) of the destination operand *)
           match st with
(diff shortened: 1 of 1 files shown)


ivg:
Since this PR and therefore, I would like that every change that affects instruction semantics, i.e., something that deals with BIL transformations or lifters, should be accompanied with the run of veri, that proves that we don't have any regressions. I think that we need to create a repository that will contain logs from bap-veri runs, so that the proof that we don't have a regression is encoded as a commit to this repo.

PR #739, opened 2017-11-21
powerpc lifter

gitoleg:
It's an initial PR in order to bring PowerPC lifter
into a life. It contains only one group of instructions (`load`),
just for discussing, along with service modules like
`Powerpc_model`, `Powerpc_dsl` and others.


lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
+
+(**
+
+   ## Intro
```
ivg: In ocamldoc sections look like `{3 Intro}`

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+
+    So proposed usage is just to open it at the very beginning of your
+    module:
+
+    open Bap_powerpc.Std
+    ...
+
+    Note, that everywhere in this module, where some operation has
+    a notion of bit position (byte/word ...), it's assumed that
+    numeration starts from most significant bit (byte/word ...).
```
ivg: We need to clarify this even further, as a casual user won't believe this :)

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+
+    Any lifter function must have two arguments.
+    The first one is a model of CPU that contains all information like
+    registers, memory access, instruction address, etc.
+    The second argument is an array of instruction operands, and its
+    user responsibility to treat each element of this array
+    according to instruction semantics. Those, for add instruction
+    which sum register and immediate and save a result in a register,
+    a user must point that the first and the second element of
+    operands array registers and the third argument is immediate.
```
ivg: I can't understand the last sentence.

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+
+    In general, what one should do to add an instruction is to write
```
ivg: It sounds to cryptic or, even, biblical. Also not sure that this is the right place for this sentence. Usually, one should start thy description, with such sentence (pun intended). Also try to follow the following simple rules when writing documentation:

1. Thouh shall not use the passive mode. 
2. Thouh shall not use complex sentences.

In the latter, I mean, that you should use too complex constructions, especially those that include lots of indirections and possessives such as "which", "those". In general, if your sentence has too much punctuation that simplify it. This is technical documentation, so there is no need to write nice prose.

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+   ## RTL
+
+    The central part of this module is RTL. It contains expressions,
+    operations over expressions and statements.
+    Basically, any line that ended with ';' is a statement,
+    and any part of it is either expression(s), or operation over
+    expression(s).
+
+    ### Bitwidth and Signess
+    Any expression in RTL has a notion of signess. Any operation over
```
ivg: it is "signedness", not signess

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+    expressions of different sign causes a casting to signed.
+
+    Any expression in RTL has a notion of bitwidth. All operations
+    over expressions of different width will cause a casting to the
+    more biggest bitwidth.
```
ivg: Let's rephrase this a little bit:

"If an operation is applied to expressions with different bitwidths, an expression with the smaller bitwidth is implicitly casted (extended) to the bitwidth of another expression."

Also, I have a question here. During the extension, what happens with the sign bit? Is this signed extension or not?

Probably the next question is trying to answer these questions, but I'm not sure I can understand it.

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+
+    Note, that any expression that tends to take part of other
+    expression is always unsigned. The same is true for any
```
ivg: I can't understand this question. What do you mean by "an expression that tends to take a part of other expression"?

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+    expression that is a concatenation of expressions.
+    But it's still possible to use them as signed expressions, e.g.
+    assign them to signed expression or use signed comparison.
+
+   ### Expressions
+    There are only a few ways to construct expression:
```
ivg: ... only few ... an expression

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+     - from instruction operand
+     - from constant
+     - from string
+     - by defining temporary variable
+
+    To construct an expression that denotes a register and treat
+     its content as an unsigned value, one should write:
+     ...
```
ivg: you should put this into the verbatim block. Did you try to compile this into html documentation?

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,614 @@
...
+                 ^      ^    ^
+                 |      |    |
+      content is |      |    |
+      unsigned __|      |    |
+                        |    |
+           claim register  from operands array at index 0
+
+    Immediate instructions operands are constructed in the same
+    way:
+     ...
```
ivg: this should be put into the code block

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,675 @@
...
+      let x = unsigned const word 42
+    ]}
+
+    And create an expression from string:
+
+    {[
+      let x = unsigned of_string "0xFFFF_FFFF_FFFF"
+    ]}
+
+    {4 Taking a part}
```
ivg: "Taking a part" is a colloquial expression that means "to participate in something", it is better to say, to extract a portion of a value. 

Also, make sure that you do not confuse two terms - expression and value. It is common when you see an expression as a value, and people do make such assumptions. But when you say "to extract a portion of expression" (or to take its part, whatever), I think of a poor expression that is taken apart, like you're extracting a variable, or a plus sign from an expression.

lib/bap_powerpc/bap_powerpc.mli
```diff
@@ -0,0 +1,675 @@
...
+    There are few useful constructions that either a part of RTL
+    ([if_], [foreach]) or simplify code ([when_], [ifnot], [switch]).
+    Also, one needs to register a lifted function, so
+    it could be called when appropriative instruction will be
+    encountered. There are two operators for this purpose:
+    - [(>>)] - just registers a function (see example below)
+    - [(>.)] - does the same, plus does some extra job (see
+             a description below)
+
+    {2 Comlete example}
```
XVilka: Typo - `complete`

plugins/powerpc/powerpc_main.ml
```diff
@@ -0,0 +1,44 @@
+open Core_kernel.Std
+open Bap.Std
+open Bap_powerpc.Std
+
+include Self()
+
+let () = Config.manpage [
+    `S "DESCRIPTION";
+    `P "Provides lifter PowerPC architecture.";
```
XVilka: Missing "for" I think.

XVilka:
Is it possible to move RTL library outside of PowerPC, so can be reused in other architectures?

lib/bap_powerpc/bap_powerpc_cpu.mli
```diff
@@ -0,0 +1,40 @@
+open Bap.Std
+open Bap_powerpc_rtl
+open Bap_powerpc_dsl
+
+type cpu = {
```
ivg: it's the third copy of the record definitions, please move it to a separate module, e.g., `bap_powerpc_types.ml` that has only type definitions and thus do not require an mli file.

lib/bap_powerpc/bap_powerpc_dsl.mli
```diff
@@ -0,0 +1,91 @@
...
+val signed : 'a ec -> 'a
+val unsigned : 'a ec -> 'a
+
+val zero : exp
+val one  : exp
+
+(** [extract e lx rx] extracts portion of [e] starting
+    at bit [lx] and ending at bit [rx], all bounds
+    are inclusive. Bits indexes start from the most
+    significant bit. *)
```
ivg: do not duplicate documentation, there should be only one single point of truth. Please erase all duplicating comments (unless they are internal for the package)

ivg:
@XVilka, I think it would be a good idea. 

When we were designing this RTL the idea was to build a eDSL that is specific to the PowerPC architecture and do no try to build something general and abstract (we have BIL for that). However,  at the end of the day, we managed to build something that looks reusable across all platforms. We should be careful here and do not introduce unnecessary abstractions and false dependencies. I think that some part of the RTL language itself (i.e., that doesn't mention anything powerpcish could be reused).

We are not planning to extract the RTL library right now, I believe that it is premature. However, feel free to copy paste parts of it into MIPS, and we will soon see what is reusable across different platforms or what is not. Let the abstractions to emerge instead of creating them. 

And with all that said, I think that in general, I think that the approach that we use in the PowerPC lifter should be definitely reused. The resulting code is very nice.

lib/bap_powerpc/bap_powerpc_dsl.mli
gitoleg: yes, they are for internal usage in the case of `Bap_powerpc_dsl` module

lib/bap_powerpc/bap_powerpc_rtl.ml
```diff
@@ -0,0 +1,451 @@
...
+  width : int;
+} [@@deriving bin_io, compare, sexp]
+
+type t =
+  | Move of exp * exp
+  | Jmp of exp
+  | Store of var * exp * exp * endian * size
+  | If of exp * t list * t list
+  | Foreach of bool * exp * exp * t list
+  | Message of string
```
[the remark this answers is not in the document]
gitoleg: ` [message m] - embeds a string [m] in code `
In fact, it just the other name for `Special` BIl variant. Looks better, as I think - what is a `Special`? 
The example  from `move` instruction code:
```
switch (num) [
      case lr_num   [ cpu.lr  := rs ];
      case ctr_num  [ cpu.ctr := rs ];
      default [ message "Unknown cpu.register number"; ]
    ];
```

lib/bap_powerpc/bap_powerpc.mli
gitoleg: Thanks!

plugins/powerpc/powerpc_main.ml
gitoleg: Thanks again!

plugins/powerpc/powerpc_branch.ml
```diff
@@ -0,0 +1,277 @@
...
+
+let bclrl = bclr ^ update_link_register
+
+(** Branch Instructions extended mnemonic, branch to LR unconditionally.
+    Page 792 of IBM Power ISATM Version 3.0 B
+    examples:
+    4e 80 00 20   blr
+    4e 80 00 21   blrl *)
+let blr cpu ops =
+  let sh = unsigned of_string "0b00" in
```
ivg: I don't like this implementation. We shall try to follow the actual CPU microcode as close as possible. I understand that this is how it is written in the manual, that an word is concatenated with two zero bits. But it would be much nicer (and closer to what actually happens in the CPU) if we will just use the shift left operation.

Also, I don't think it is a good idea to use the string representation of a binary word here. Not sure whether it is worthwhile anywhere.

plugins/powerpc/powerpc_add.ml
```diff
@@ -0,0 +1,188 @@
...
+    37 de fd 28    addic. r30, r30, -728 *)
+let addic cpu ops =
+  let rt = signed cpu.reg ops.(0) in
+  let ra = signed cpu.reg ops.(1) in
+  let im = signed imm ops.(2) in
+  let tm = signed var doubleword in
+  RTL.[
+    tm := ra;
+    rt := ra + im;
+    cpu.ca := low cpu.word_width rt < low cpu.word_width tm;
```
ivg: Will this comparison be signed? If yes, then probably this is incorrect.

plugins/powerpc/powerpc_arith.ml
```diff
@@ -0,0 +1,420 @@
+
```
ivg: that's strange modularization, it sounds like that the addition is not a part of arithmetics :)

plugins/powerpc/powerpc_arith.ml
```diff
@@ -0,0 +1,420 @@
...
+open Powerpc.Std
+
+(** Fixed-Point Arithmetic Instructions - Negate
+    Page 72 of IBM Power ISATM Version 3.0 B
+    example:
+    7c 22 00 d0   neg  r1, r2
+    7c 22 00 d1   neg. r1, r2 *)
+let neg cpu ops =
```
ivg: can you please explain how you are handling the dots? I don't see any CRx specific code here, though you're writing in the comments that this code corresponds to both `neg` and `neg.`. 

The same question for all other dot instructions.

plugins/powerpc/powerpc_arith.ml
```diff
@@ -0,0 +1,420 @@
...
+(** Fixed-Point Arithmetic Instructions - Substract From
+    Page 69 of IBM Power ISATM Version 3.0 B
+    example:
+    7c 22 18 50    subf  r1, r2, r3
+    7c 22 18 51    subf. r1, r2, r3 *)
+let subf cpu ops =
+  let rt = unsigned cpu.reg ops.(0) in
+  let ra = unsigned cpu.reg ops.(1) in
+  let rb = unsigned cpu.reg ops.(2) in
+  RTL.[rt := (lnot ra) + rb + one]
```
ivg: there is a strange controversy between addition and subtraction. In addition you're using signed operands while in subtraction they are unsigned. On the first glance, this looks incorrect. Also addition and subtraction are idempotent to the signedness of their operands. Probably it only matters when conditional flags are involved, especially when you express them via the comparison operation, that are indeed dependent on the interpretation of the sign bit. And my intuition here tells me, that we shall compare unsigned. Please, can you provide some insight on this?

plugins/powerpc/powerpc_arith.ml
```diff
@@ -0,0 +1,420 @@
...
+(** Fixed-Point Arithmetic Instructions - Substract From Immediate Carrying
+    Page 70 of IBM Power ISATM Version 3.0 B
+    example:
+    20 22 10 92    subfic r1, r2, 4242 *)
+let subfic cpu ops =
+  let rt = unsigned cpu.reg ops.(0) in
+  let ra = unsigned cpu.reg ops.(1) in
+  let si = unsigned imm ops.(2) in
+  RTL.[
+    rt := (lnot ra) + si + one;
```
ivg: You're inconsistent with using parentheses around the unary operators. They are not needed here, though you're still using them. At the same time, at other places you're not using. People may start to believe that parentheses here matter.

plugins/powerpc/powerpc_arith.ml
```diff
@@ -0,0 +1,420 @@
...
+    cpu.ca32 := low word si < low word ra;
+  ]
+
+(** Fixed-Point Arithmetic Instructions - Substract From Carrying
+    Page 70 of IBM Power ISATM Version 3.0 B
+    example:
+    7c 22 18 10    subfc  r1, r2, r3
+    7c 22 18 11    subfc. r1, r2, r3 *)
+let subfc cpu ops =
+  let rt = signed cpu.reg ops.(0) in
```
ivg: now it really suspicious, why are you using unsinged operands for `subfic` and signed for `subfc` if they only differ in the sort of the third operand, i.e., register vs. immediate.

plugins/powerpc/powerpc_add.ml
```diff
@@ -0,0 +1,188 @@
...
+    3f de 00 02     addis   r30,r30,2
+    3d 6b f0 00     addis   r11,r11,-4096 *)
+let addis cpu ops =
+  let rt = signed cpu.reg ops.(0) in
+  let ra = signed cpu.reg ops.(1) in
+  let im = signed imm ops.(2) in
+  let sh = unsigned const word 16 in
+  RTL.[ rt := ra + (im lsl sh); ]
+
+let lis cpu ops =
```
ivg: modularization: why is it in the add module?

plugins/powerpc/powerpc_branch.ml
```diff
@@ -0,0 +1,277 @@
...
+
+let bl cpu ops =
+  let im = signed imm ops.(0) in
+  let tm = signed var (bitwidth 26) in
+  let sh = unsigned of_string "0b00" in
+  let ad = unsigned const byte 4 in
+  RTL.[
+    tm := im ^ sh;
+    cpu.jmp (cpu.pc + tm);
+    cpu.lr := cpu.pc + ad;
```
ivg: you can't update the link register after the jump is already executed. The jump takes effect immediately.

plugins/powerpc/powerpc_branch.ml
```diff
@@ -0,0 +1,277 @@
...
+    ];
+    ctr_ok := nth bit x 2 lor ((cpu.ctr <> zero) lxor (nth bit x 3));
+    cond_ok := nth bit x 0 lor (bi lxor (lnot (nth bit x 1)));
+    when_ (ctr_ok land cond_ok) [
+      tm := bd ^ sh;
+      cpu.jmp tm;
+    ]
+  ]
+
+let bcl = bc ^ update_link_register
```
ivg: this won't work, as it will update the link register after the jump, not before.

plugins/powerpc/powerpc_cpu.ml
```diff
@@ -0,0 +1,74 @@
...
+  let x = int_of_bitwidth x in
+  match Size.of_int x with
+  | Ok s -> s
+  | Error _ -> ppc_fail "invalid size: %d" x
+
+let find m reg =
+  let (module M : PowerPC) = m in
+  let open M in
+  let open E in
+  let find_reg regs reg = String.Map.find regs (Reg.name reg) in
```
ivg: use `Map.find` instead of `String.Map.find`. First of all it less verbose, second it confusing, third it make the code less general, and last but not least, it will break in core v0.10. They are removing Maps from the modules.

plugins/powerpc/powerpc_cpu.ml
```diff
@@ -0,0 +1,74 @@
...
+  let find_vr  = find_reg vr in
+  let find_fpr = find_reg fpr in
+  let find_cr_bit = find_reg crn in
+  let find_cr_field = find_reg cr_fields in
+  let reg_searches =
+    [find_gpr; find_cr_bit; find_cr_field; find_fpr; find_vr] in
+  List.filter_map reg_searches ~f:(fun f -> f reg) |> function
+  | [] -> Exp.of_word (Word.zero gpr_bitwidth)
+  | hd :: [] -> hd
+  | _ -> ppc_fail "Register name %s is ambiguous!!!" (Reg.name reg)
```
ivg: You did it! 🙏

gitoleg:
Most important changes after review:
1. There are not any special `signed` version of operators: `<$`, `>$`, `<=$`, `>=$`, `\$`, `%$` in RTL. Instead, a signed version will be chosen for signed arguments. So, if `x` is signed, a `<` is a signed less-then operator and the value of `y` is `1`:
```
      let x = signed const halfword 0xFAAA in
      let y = unsigned var bit in
      RTL.[
        y := x < zero;
      ]
```
and if `x` is unsigned, a `<` is a unsigned less-then operator and the value of `y` is `0`
```
      let x = unsigned const halfword 0xFAAA in
      let y = unsigned var bit in
      RTL.[
        y := x < zero;
      ]
```
2. There are not any logical shift operators `lsl` and `lsr`. Instead, we introduce shift operators `<<` and `>>`. A shift is logical if its arguments are unsigned, and a shift is arithmetical if signed. 
```
      let x = signed const halfword 0xFAAA in
      let s = unsigned const halfword 4 in
      let y = unsigned var halfword in
      RTL.[
        y := x >> s;
      ]
  ```
If [x] is signed, like in example above, then shift is arithmetical, and result is 0xFFAA. If [x] is unsigned, then shift  is logical and result is Ox0FAA.

PR #743, opened 2018-01-07
Fix grammar inside a comment

murmour:
P.S. The Monads library is very neat. Thank you!


ivg:
Thanks! 

The CI failed because one of the machines timed out. I've restarted it, so let's wait a bit. If it will continue to fail, I will override the CI and merge. I have a strong belief that your change won't break any code :)

PR #746, opened 2018-01-12
few minor changes in Bap_helpers

gitoleg:
1) fixed bug for XOR simplification
2) added/edited branches for extract and cast simplification

### DIFF
--- a/lib/bap_types/bap_helpers.ml
+++ b/lib/bap_types/bap_helpers.ml
@@ -163,6 +163,8 @@ module Apply = struct
     | Cast.HIGH     -> ext ~lo:Int.(Bitvector.bitwidth u - sz) u
     | Cast.LOW      -> ext ~hi:Int.(sz - 1) u
 
+  let extract hi lo w = Bitvector.extract_exn ~hi ~lo w
+
 end
 
 module Type = struct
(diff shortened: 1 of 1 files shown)


lib/bap_types/bap_helpers.ml
```diff
@@ -163,6 +163,8 @@ module Apply = struct
     | Cast.HIGH     -> ext ~lo:Int.(Bitvector.bitwidth u - sz) u
     | Cast.LOW      -> ext ~hi:Int.(sz - 1) u
 
+  let extract hi lo w = Bitvector.extract_exn ~hi ~lo w
```
ivg: why is it needed?

lib/bap_types/bap_helpers.ml
```diff
@@ -469,16 +471,17 @@ module Simpl = struct
...
-    and cast t s x =
-      let x = exp x in
-      match Type.infer_exn x with
-      | Type.Imm s' when s' = s -> x
+    and cast t s x = match exp x with
+      | Int w -> Int (Apply.cast t s w)
       | _ -> Cast (t,s,x)
+    and extract hi lo x = match exp x with
+      | Int w -> Int (Apply.extract hi lo w)
+      | _ -> Extract (hi,lo,x)
```
ivg: you're losing an effect of applying `exp` to the `x` argument, it should be
```ocaml
| x -> Extract (hi,lo,x)
```

lib/bap_types/bap_helpers.ml
```diff
@@ -520,7 +523,7 @@ module Simpl = struct
       | OR,x,y  when x = y -> x
       | XOR,x,y when x = y -> zero width
       | XOR,x,y when is0 x -> y
-      | XOR,x,y when is0 y -> y
+      | XOR,x,y when is0 y -> x
```
ivg: nice catch!
we definitely need a test case that will catch such transformations. Probably, some automatic expression generator. We need to think about this.

Issue #747, opened 2018-01-13
bitstring < 3.0.0


percontation:
bitstring seems to have removed their camlp4 preprocessor in version 3.0.0, causing parts of bap's opam install to fail (e.g. bap-elf).

ivg:
I see, we should constrain the version of bitstrings. As a workaround, you can specify constraints on the command line:

```
opam install bap "bitstring<3.0.0"
```

ethan42:
I think you can probably close this now :)

ivg:
not yet... apparently you're not the only one who doesn't know how our build system is working :) I missed the fact, that the `bap-elf` library is disabled by default. So, if it is enabled, and #748 patch is applied, the build fails. @gitoleg  is currently investigating the issue.

ivg:
marking this issue as partially solved. The issue is now tracked in #756

PR #748, opened 2018-01-15
switch to bitstring.ppx

percontation:
Changes deps from bitstring<3.0.0 -> bitstring>=3.0.0


percontation:
Re: #747. It actually looked pretty trivial to update to bitstring 3.0.0's ppx stuff.

(I didn't quite get bap building from source so the build is probably broken; this is more of a "look how easy I think it is PR" than something that's necessarily mergable)

PR #751, opened 2018-01-16
Making YMM registers available for IA32 and AMD64 users.

ethan42:
The PR makes YMM registers available for IA32 and AMD64 users. The PR also fixes a few 
inconsistencies in the interface exposed internally in plugin modules (e.g., the x86_lifter):
the number of YMM registers now varies based on the architecture and the `nums`
array was removed and replaced by references to the AMD64 `r` array. The `r` array definition was
changed to empty for IA32 internals (it can still throw runtime errors and needs to be removed in
the future, but this version is better than the previous approach).

### DIFF
--- a/lib/x86_cpu/x86_cpu.ml
+++ b/lib/x86_cpu/x86_cpu.ml
@@ -10,7 +10,7 @@ module Make_CPU(Env : ModeVars) = struct
   let gpr = Var.Set.of_list @@ [
       rax; rcx; rdx; rsi; rdi;
       rbx; rbp; rsp;
-    ] @ Array.to_list nums
+    ] @ Array.to_list r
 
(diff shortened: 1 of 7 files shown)


ethan42:
That makes sense (this is why I asked on gitter, I wasn't sure if that can be included in the current version). The other part of the PR moves ymms from the global scope to the architecture-parameterized scope which depending on how much you care about the X86_env could also be an api-breaking change (none of these are exposed at X86_cpu so it seemed like it should be fine). On the other hand, the previous version was incorrect, so it's up to you :)

ethan42:
Fixed the deprecation warning and also fixed up the test for pshufb.

ivg:
Thanks! Travis finally did it.

PR #757, opened 2018-01-18
Minor bug fixes

gitoleg:
minor bug fixed:
1) in primus lisp - read-word didn't read a requested count of words
2) in x86-lifter - forgotten llvm version constraint for `movoffset` instruction
   after added llvm 5.0 support
3) in bap_llvm library fixed a bug that crushes bap build with llvm 3.4

### DIFF
--- a/plugins/x86/x86_mov_offset.ml
+++ b/plugins/x86/x86_mov_offset.ml
@@ -199,7 +199,8 @@ module Self = Self ()
 
 let () =
   if llvm_version = "3.4" then T_34.register ()
-  else if llvm_version = "3.8" || llvm_version = "4.0" then T_38.register ()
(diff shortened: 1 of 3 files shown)


plugins/x86/x86_mov_offset.ml
```diff
@@ -195,7 +195,8 @@ module Self = Self ()
 
 let () =
   if llvm_version = "3.4" then T_34.register ()
-  else if llvm_version = "3.8" || llvm_version = "4.0" then T_38.register ()
+  else if llvm_version = "3.8" || llvm_version = "4.0" || llvm_version = "5.0"
+  then T_38.register ()
```
ivg: Sooner or later we should rename this module.

Issue #694, opened 2017-08-17
The file loading is broken


ivg:
Looks like that we messed somewhere in the project loaders, but it doesn't work anymore:

```
bap bin/arm-linux-gnueabi-echo -dmarshal:arm-echo.marshal
$ bap arm-echo.marshal --source-type=project
Failed to create a project: ("File format is not supported"
  "Raised at file \"src/error.ml\", line 7, characters 20-30\
 \nCalled from file \"lib/bap/bap_project.ml\", line 242, characters 48-55\
 \nCalled from file \"src/or_error.ml\", line 59, characters 9-15\
 \n")
```

ivg:
Apparently, not only the project loading is broken, but all other uses cases do not work, e.g.,

````
printf '\x83\x3d\xe9\xb5\x21\x00\x01' > exe
bap ./exe --source-type=x86_64-code
Failed to create a project: ("File format is not supported"
  "Raised at file \"src/import0.ml\" (inlined), line 234, characters 22-32\
 \nCalled from file \"src/error.ml\" (inlined), line 9, characters 14-30\
 \nCalled from file \"src/or_error.ml\", line 70, characters 17-32\
 \nCalled from file \"lib/bap/bap_project.ml\", line 242, characters 48-55\
 \nCalled from file \"src/or_error.ml\", line 62, characters 9-15\
 \n")
```

PR #758, opened 2018-01-23
Fix source-type cmdline option

gitoleg:
Cmdline option source-type had no effect. This PR solves this issue.
So there are not any errors when loading a project from `.marshal` file
or when using a raw code with bap.
fix https://github.com/BinaryAnalysisPlatform/bap/issues/694

### DIFF
--- a/src/bap_main.ml
+++ b/src/bap_main.ml
@@ -118,8 +118,7 @@ let main o =
     | Ok project ->
       Project.Cache.save (digest o) project;
       project in
-  let proj_of_file file =
-    let fmt,ver = parse_filename file in
+  let proj_of_file ?ver ?fmt file =
     In_channel.with_file file
       ~f:(fun ch -> Project.Io.load ?fmt ?ver ch) in
   let project = match Project.Cache.load (digest o) with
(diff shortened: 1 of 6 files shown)


src/bap_main.ml
```diff
@@ -95,27 +95,48 @@ let process options project =
...
-      and brancher = brancher o
-      and reconstructor = reconstructor o
-      and symbolizer = symbolizer o in
-      let input =
+    | None -> match o.source with
+      | `File "builtin" ->
+        let fmt,ver = parse_filename o.filename in
+        proj_of_file ?fmt ?ver o.filename
+      | `File fmt ->
+        proj_of_file ~fmt o.filename
```
ivg: Ok, we have lots of confusion here. Originally,  `File <name>` mapped to the `<name>-custom` where `<name>` "should be a name of factory method registered for a file source in project factory", as per `bap --help` documentation. It is not a format name, it is a name of the method, i.e., the loader name. 

Your interpretation basically makes the custom syntax just another way to specify a format, so we have two equivalent variants:
```
bap ./exe.marshal --source-type=project
bap ./exe --source-type=marshal-custom
```

It looks like that we don't need the `*-custom` form anymore, as it is superseded by the `loader` option. 

So, please, can you rectify this by:

1. removing the custom form from the source-type parse
2. fixing the ouptut representation, by changing `File of string` to `Project`
3. updating the source-type documentation

ivg:
and also, please add test cases to bap-testsuite, so that we don't need to fix it again.

ivg:
LGTM as soon as you will update the tests

PR #759, opened 2018-01-25
Merge sources

gitoleg:
rewritten merge-sources function. We don't need to wait for all sources to produce a result anymore.

### DIFF
--- a/src/bap_main.ml
+++ b/src/bap_main.ml
@@ -35,13 +35,13 @@ let merge_streams ss ~f : 'a Source.t option =
 let merge_streams ss ~f : 'a Source.t option =
   let stream, signal = Stream.create () in
   List.iter ss ~f:(fun s -> Stream.observe s (fun x -> Signal.send signal x));
-  let stream : 'a Source.t = Stream.parse stream ~init:None
-    ~f:(fun state x -> match state with
-        | None -> Some (Ok x), Some x
-        | Some y ->
-          printf "called merge!\n";
-          Some (Ok (f x y)), Some (f x y)) in
-  Option.some stream
+  let stream = Stream.parse stream ~init:None
+      ~f:(fun state x -> match x, state with
+          | Ok x, None -> Some (Ok x), Some x
+          | Ok x, Some y -> Some (Ok (f x y)), Some (f x y)
+          | error, Some y -> Some error, Some y
+          | error, None -> Some error, None) in
+  Some stream
 
 let merge_sources create field (o : Bap_options.t) ~f =  match field o with
   | [] -> None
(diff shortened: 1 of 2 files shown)


src/bap_main.ml
```diff
@@ -20,17 +20,20 @@ let brancher = find_source (module Brancher.Factory) brancher
...
-      | Ok a, Ok b -> Ok (f a b)
-      | Error a, Error b -> Error (Error.of_list [a;b])
-      | Error e,_|_,Error e -> Error e))
+  let stream, signal = Stream.create () in
+  List.iter ss ~f:(fun s -> Stream.observe s (fun x -> Signal.send signal x));
+  let pair x = Some x, Some x in
+  Stream.parse stream ~init:None
+    ~f:(fun state x -> match x, state with
+        | Ok x, None -> pair (Ok x)
+        | Ok x, Some (Ok y) -> pair (Ok (f x y))
```
ivg: You are flipping the order of argument to the function `f` and since `f` is not commutative (e.g., `Symbolizer.chain`) it may lead to a confusion. Basically, your merge function applies `f new old` while it should do `f old new`

src/bap_main.ml
```diff
@@ -20,17 +20,20 @@ let brancher = find_source (module Brancher.Factory) brancher
...
+        | Error e, Some (Ok _)
+        | Ok _, Some (Error e) -> pair (Error e)
+        | Error x, None -> Some (Error x), None
+        | Error x, Some (Error y) ->
+          pair (Error (Error.of_list [x;y]))) |>
```
ivg: also `[y,x]` for consistency.

src/bap_main.ml
```diff
@@ -20,17 +20,20 @@ let brancher = find_source (module Brancher.Factory) brancher
...
 let merge_streams ss ~f : 'a Source.t option =
-  List.reduce ss ~f:(Stream.merge ~f:(fun a b -> match a,b with
-      | Ok a, Ok b -> Ok (f a b)
-      | Error a, Error b -> Error (Error.of_list [a;b])
-      | Error e,_|_,Error e -> Error e))
+  let stream, signal = Stream.create () in
+  List.iter ss ~f:(fun s -> Stream.observe s (fun x -> Signal.send signal x));
+  let pair x = Some x, Some x in
+  Stream.parse stream ~init:None
+    ~f:(fun state x -> match x, state with
```
ivg: instead of `state` you can use something more readable, e.g., `fun prev curr` or just `fun x x'` or even `fun x y` is better and indicates that the first parameter represents the information obtained on the previous step, and the second represents the information obtained on the current step.

PR #760, opened 2018-01-26
restores read-roots plugin

gitoleg:
This PR brings back the `read-roots` plugin. Actually, it works similar to 
`read-symbols`, but expected addresses only. Rooter is registered under the name "file".
And a name of registered services in `read-symbols` plugin changed.
Now it is a concatenation of "file:" and filename, where symbols are read from.


plugins/read_roots/root_loader_main.ml
```diff
@@ -1,24 +1,33 @@
...
-    extension name >>= fun fmt -> Addr.find_reader fmt >>| fun _ ->
-    In_channel.with_file filename ~f:(fun chan ->
-        Addr.Io.load_all ?ver ~fmt chan |>
-        Seq.of_list |>
-        create)
-end
+let register filename =
+  Stream.map Project.Info.arch (fun arch ->
+      Ok (Rooter.create (extract filename arch))) |>
+  Rooter.Factory.register "file"
```
ivg: the rooter name shall be consistent with the the read_symbols plugin, i.e., it should be `file:<filename>`

plugins/read_roots/root_loader_main.ml
```diff
@@ -1,24 +1,33 @@
...
+
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "Read roots information from a file and provide a rooter under
+          the name $(b,file). So print $(b,--rooter=file) to use only this
```
ivg: make it consistent with the read-symbols plugin

plugins/read_roots/root_loader_main.ml
```diff
@@ -1,24 +1,55 @@
...
+let rooter filename arch =
+  let name,ver = split_right ~after:'-' filename in
+  match extension name with
+  | None -> extract_default filename arch
+  | Some fmt -> extract_with_fmt filename fmt ver
+
+let register filename =
+  let name = sprintf "file:%s" filename in
+  Stream.map Project.Info.arch (fun arch ->
+      Ok (Rooter.create (rooter filename arch))) |>
```
ivg: What if I will pass a 64 bit addresses for 32 bit arch, would it still be `Ok`?

gitoleg:
updated and added a couple of tests:
https://github.com/BinaryAnalysisPlatform/bap-testsuite/pull/35

plugins/read_roots/root_loader_main.ml
```diff
@@ -1,24 +1,86 @@
...
+      | Ok s -> Ok (Rooter.create s)
+      | Error e -> Error e) |>
+  Rooter.Factory.register name
+
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "Read roots information from a file and provide a rooter.
+          The name of registered rooter is a concatenation of
+          \"file:\" and a filename with roots. So once roots are read,
```
ivg: What was the motivation for splicing the filename inside of rooter's provider name? 

The user experience is awkward (even worse than it was for the read_symbols), e.g.,
```
bap --rooter=file:/a/path/to/file --read-symbols-from=/a/path/to/file
```

Why do we need to specify the filename twice? 

TBH, I remember we were discussing this issue IRL and came to this conclusion. Today, four months later, I can't recall why we did this, and now it doesn't look good.

plugins/read_roots/root_loader_main.ml
```diff
@@ -1,24 +1,86 @@
...
+          use $(b,--rooter)=$(b,file:filename) to use them.
+          Default file format is just a list of numbers. One can use
+          a common convention to specify other format, see $(bap --help)
```
ivg: one not knows about any "common conventions" :)

Seriously, these words do not say anything useful. And `bap --help` is a pretty long read, and without any guidance, it is impossible to find that "common convention" (which, by the way, doesn't exist). So, either provide more details or specify the format description right here. I'm voting for the latter, since the file extension matters as it defines the format of a file, moreover, although you're trying to be consistent with the way how bap handles file names, you're still doing a lot of work here, in the plugin, that means that in future the conventions may diverge.

ivg:
hold this after the bap-service is merged

ivg:
closing it since bap-service was forfeited

PR #761, opened 2018-01-31
fixes the search of destinations in the reconstructor

gitoleg:
In the reconstructor block destinations are computed directly from the BIL representation, that is not correct, as BIL doesn't take into account the relocations. And if the relocatable brancher is used, then the CFG contains correct edges, but this information is ignored when we build a set of call destinations in the reconstructor.  This PR addresses this problem by searching a destination directly in CFG.

### DIFF
--- a/lib/bap_disasm/bap_disasm_reconstructor.ml
+++ b/lib/bap_disasm/bap_disasm_reconstructor.ml
@@ -19,23 +19,18 @@ type reconstructor = t
 let create f = Reconstructor f
 let run (Reconstructor f) = f
 
-
-let dest_of_bil bil =
-  (object inherit [word] Stmt.finder
-    method! enter_jmp dst goto = match dst with
-      | Bil.Int dst -> goto.return (Some dst)
-      | _ -> goto
-  end)#find bil
-
 let find_calls name roots cfg =
   let starts = Addr.Table.create () in
   List.iter roots ~f:(fun addr ->
       Hashtbl.set starts ~key:addr ~data:(name addr));
   Cfg.nodes cfg |> Seq.iter ~f:(fun blk ->
-      let term = Block.terminator blk in
-      if Insn.(is call) term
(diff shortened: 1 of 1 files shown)


lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,23 +19,18 @@ type reconstructor = t
...
   Cfg.nodes cfg |> Seq.iter ~f:(fun blk ->
-      let term = Block.terminator blk in
-      if Insn.(is call) term
-      then Option.iter (dest_of_bil (Insn.bil term))
-          ~f:(fun w -> Hashtbl.set starts ~key:w ~data:(name w)));
+      match Cfg.Node.outputs blk cfg |> Seq.hd with
+      | None -> ()
+      | Some e ->
+        let term = Block.terminator blk in
+        if Insn.(is call) term && Cfg.Edge.label e = `Jump then
```
ivg: there could be conditional calls, so the it should be 
```
Insn.(is call) term && Cfg.Edge.label <> `Fall
```

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,23 +19,18 @@ type reconstructor = t
...
 let find_calls name roots cfg =
   let starts = Addr.Table.create () in
   List.iter roots ~f:(fun addr ->
       Hashtbl.set starts ~key:addr ~data:(name addr));
   Cfg.nodes cfg |> Seq.iter ~f:(fun blk ->
-      let term = Block.terminator blk in
-      if Insn.(is call) term
-      then Option.iter (dest_of_bil (Insn.bil term))
-          ~f:(fun w -> Hashtbl.set starts ~key:w ~data:(name w)));
+      match Cfg.Node.outputs blk cfg |> Seq.hd with
```
ivg: You shall consider _all_ edges, so you shall iterate over all edges.

PR #762, opened 2018-02-01
Initial MIPS lifter draft

XVilka:
Heavily inspired by PowerPC lifter in https://github.com/BinaryAnalysisPlatform/bap/pull/739

I tested many opcodes by hand but still need to figure out how to write autotests for BAP.
I hope RTL part can be shared with PowerPC and split into separate library. Sometimes using it causes more verbose code, but certainly easier to read and understand - very good invention, imho.

Probably some opcodes are still missing, will work on them along with tests.
So waiting for your feedback.


gitoleg:
@XVilka, could you say please, what documents did you use when writing this lifter?

ivg:
@XVilka  so it works nearly out of the box, that's good news! There are some issues and few instructions were missing, the patch is attached below for the reference point (its crappy, just to make things done, we are working on it). We can't make pull requests due to the indentations issues, so we will stick to patches until you push the indentation update. We may push the changes ourselves, or if at the time of reading we will be sleeping at the time when you're reading this (BAP is the empire on which the sun never sets) you can just grab the latest patch and apply it yourself (we will keep updating the patch, or uploading new patches). 

Here are the problems that the patch fixes:

1. we have a mess with register aliases, basically there are registers that are referenced by numbers ($0,$31) and there are named registers that are actually aliases to it, for example A0 is $4 (that is R4 in your cpu model). You also defined them as 64 bit registers, even in 32-bit architectures. But in any case, it looks like that the backend never uses the `$x` notation, and always uses `Ax`, ` Vx`, etc. The only place where the gpr registers were used is in the ABI. I've fixed this in a crappy way, @gitoleg will rectify it by using right values exported via the CPU interface.

2. the ABI needs a little bit of demangling, not a big issue.

3. all jumps were off by 4 bytes 

4. `jal`, `jialrc` shall assign the fall-through address to the `RA` (aka $31) register

5. we added `bne`, `slt`, `sltu`, and `sltiu` instructions.  

6. in `jal` you were assigning to the PC register, that's an error, @gitoleg will fix the docs. 

That's all. So far the plan is to merge it before the 1.4 release (the pre-release is planned on Feburary, 14th). It is already good to go, especially the 32-bit variant. I have some concerns about the 64 bit variant, as some constructions in the lifter do not look cross-platform (i.e., when we set RA register, we shall use word-width and do not hard code 8, and so on, @gitoleg  is currently looking at those issues). 

The patch is below, apply it with `git apply --ignore-whitespace <PATCH>`

```diff

diff --git a/plugins/mips/mips_abi.ml b/plugins/mips/mips_abi.ml
index 656858d..f5b9ba1 100644
--- a/plugins/mips/mips_abi.ml
+++ b/plugins/mips/mips_abi.ml
@@ -23,15 +23,16 @@ module Abi32 = struct
   open Mips.Std
   open MIPS_32
 
-    let reg i = Int.Map.find_exn gpri i |> Bil.var
+  let a i = Map.find_exn gpr (sprintf "A%d" i) |> Bil.var
+  let v i = Map.find_exn gpr (sprintf "V%d" i) |> Bil.var
   let name = "mips32"
   let size = object
     inherit C.Size.base `ILP32
   end
   let arg = function
-        | Ret_0 -> reg 2
-        | Ret_1 -> reg 3
-        | Arg n -> reg (n + 4)
+    | Ret_0 -> v 0
+    | Ret_1 -> v 1
+    | Arg n -> a n
 
 end
 
@@ -74,6 +75,19 @@ let api abi proto =
 
 let dispatch abi sub attrs proto = api abi proto
 
+
+let strip_leading_dot s =
+  match String.chop_prefix s ~prefix:"." with
+  | None -> s
+  | Some s -> s
+
+let demangle demangle prog =
+  Term.map sub_t prog ~f:(fun sub ->
+      let name = demangle (Sub.name sub) in
+      Tid.set_name (Term.tid sub) name;
+      Sub.with_name sub name)
+
+
 let main proj = match Project.arch proj with
   | `mips ->
     info "using MIPS ABI";
@@ -84,8 +98,8 @@ let main proj = match Project.arch proj with
     let api = C.Abi.create_api_processor Abi32.size abi in
     Bap_api.process api;
     let prog = Project.program proj in
+    let prog = demangle strip_leading_dot prog in
     Project.set (Project.with_program proj prog) Bap_abi.name Abi32.name
   | _ -> proj
 
 let setup () = Bap_abi.register_pass main
-
diff --git a/plugins/mips/mips_branch.ml b/plugins/mips/mips_branch.ml
index a050b01..e5ab9ac 100644
--- a/plugins/mips/mips_branch.ml
+++ b/plugins/mips/mips_branch.ml
@@ -67,7 +67,7 @@ let bltzalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt < zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -77,7 +77,7 @@ let blezalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt <= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -87,7 +87,7 @@ let bgezalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt >= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -97,7 +97,7 @@ let bgtzalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt > zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -107,7 +107,7 @@ let beqzalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt = zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -117,7 +117,7 @@ let bnezalc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt <> zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -128,7 +128,7 @@ let bltc cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs < rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -139,7 +139,7 @@ let bgec cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs >= rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -150,7 +150,7 @@ let beqc cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs = rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -161,7 +161,7 @@ let bnec cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs <> rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -172,7 +172,7 @@ let bltuc cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs < rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -183,7 +183,7 @@ let bgeuc cpu ops =
   let off = signed imm ops.(2) in
   RTL.[
     when_ (rs >= rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -193,7 +193,7 @@ let bltzc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt < zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -203,7 +203,7 @@ let blezc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt <= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -213,7 +213,7 @@ let bgezc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt >= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -223,7 +223,7 @@ let bgtzc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rt > zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -233,7 +233,7 @@ let beqzc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rs = zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -243,7 +243,7 @@ let bnezc cpu ops =
   let off = signed imm ops.(1) in
   RTL.[
     when_ (rs <> zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+      cpu.jmp (cpu.cia + off);
     ];
   ]
 
@@ -287,8 +287,9 @@ let jump cpu ops =
 (* JAL target *)
 let jal cpu ops =
   let target = signed imm ops.(0) in
+  let insn_size = unsigned const byte 8 in
   RTL.[
-        cpu.cia := cpu.cia + unsigned const byte 8;
+    cpu.gpr 31 := cpu.cia + insn_size;
     cpu.jmp target;
   ]
 
@@ -305,8 +306,9 @@ let jalr cpu ops =
 let jialc cpu ops =
   let rt = signed cpu.reg ops.(0) in
   let off = signed imm ops.(1) in
+  let insn_size = unsigned const byte 8 in
   RTL.[
-        cpu.cia := cpu.cia + unsigned const byte 4;
+    cpu.gpr 31 := cpu.cia + insn_size;
     cpu.jmp (rt + off);
   ]
 
@@ -325,6 +327,26 @@ let jr cpu ops =
     cpu.jmp rs;
   ]
 
+
+let bne cpu ops =
+  let ra = unsigned cpu.reg ops.(0) in
+  let rb = unsigned cpu.reg ops.(1) in
+  let off = signed imm ops.(2) in
+  RTL.[
+    when_ (lnot (ra = rb)) [
+      cpu.jmp (cpu.cia + off)
+    ]
+  ]
+
+let bnez cpu ops =
+  let ra = unsigned cpu.reg ops.(0) in
+  let off = signed imm ops.(1) in
+  RTL.[
+    when_ (lnot (ra = zero)) [
+      cpu.jmp (cpu.cia + off)
+    ]
+  ]
+
 let () =
   "BAL" >> bal;
   "BEQ" >> beq;
@@ -359,4 +381,51 @@ let () =
   "JIALC" >> jialc;
   "JIC" >> jic;
   "JR" >> jr;
+  "BNE" >> bne;
+  "BNEZ" >> bnez
+
 
+
+(* 0062102a        slt     v0,v1,v0 *)
+let slt cpu ops =
+  let rd = signed cpu.reg ops.(0) in
+  let rs = signed cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[
+    if_ (rs <$ rt) [
+      rd := one;
+    ] [
+      rd := zero
+    ]
+  ]
+
+(* 0051182b        sltu    v1,v0,s1 *)
+let sltu cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[
+    if_ (rs < rt) [
+      rd := one;
+    ] [
+      rd := zero
+    ]
+  ]
+
+(* 2c420007        sltiu   v0,v0,7 *)
+let sltiu cpu ops =
+  let rt = signed cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let im = signed imm ops.(2) in
+  RTL.[
+    if_ (rs < im) [
+      rt := one;
+    ] [
+      rt := zero
+    ]
+  ]
+
+let () =
+  "SLT"   >> slt;
+  "SLTu"  >> sltu;
+  "SLTiu" >> sltiu
diff --git a/plugins/mips/mips_model.ml b/plugins/mips/mips_model.ml
index 197b8cb..ee5ec85 100644
--- a/plugins/mips/mips_model.ml
+++ b/plugins/mips/mips_model.ml
@@ -114,7 +114,7 @@ module Make_MIPS(S: Spec) : MIPS = struct
 
   let gpr = Array.to_list gprs |> reglist_to_map
   (* FIXME: better representation *)
-    let gpri = make_regs_i (Type.imm fpr_bitwidth) "R" range32
+  let gpri = make_regs_i (Type.imm gpr_bitwidth) "R" range32
 
   module E = struct
     include Exps
@@ -178,4 +178,3 @@ end
 
 module MIPS_32_cpu = Make_cpu(MIPS_32)
 module MIPS_64_cpu = Make_cpu(MIPS_64)
-
diff --git a/testsuite b/testsuite
index 669bf1c..b54cf30 160000
--- a/testsuite
+++ b/testsuite
@@ -1 +1 @@
-Subproject commit 669bf1c855d491dd3dceeaa020235f20cb9dfd2d
+Subproject commit b54cf308c97adfa50bf2dee0266517750c936b33-dirty
```

gitoleg:
@XVilka, there are few changes and I tried to preserve your identation to make it easier to view changes and apply patch. 
So, speaking about setting register in instructions like branch and link, a proposed solution is to add to the cpu model  `word_width` and `word_bitwidth` fields. Anyway, we don't need to have both `addr_width` and `gpr_bitwidth` in cpu model since they are equal to each other (and this will be also fixed in powerpc lifter). Once we added those fields, we can write a generic code for both 32-bit and 64-bit architectures.
Also we nees to create a separate file for compare instructions `slt`, `sltu`, `sltiu`.

```diff
diff --git a/plugins/mips/mips_abi.ml b/plugins/mips/mips_abi.ml
index 656858d..e9249b7 100644
--- a/plugins/mips/mips_abi.ml
+++ b/plugins/mips/mips_abi.ml
@@ -23,15 +23,16 @@ module Abi32 = struct
     open Mips.Std
     open MIPS_32
 
-    let reg i = Int.Map.find_exn gpri i |> Bil.var
+  let a i = Map.find_exn gpr (sprintf "A%d" i) |> Bil.var
+  let v i = Map.find_exn gpr (sprintf "V%d" i) |> Bil.var
     let name = "mips32"
     let size = object
         inherit C.Size.base `ILP32
     end
     let arg = function
-        | Ret_0 -> reg 2
-        | Ret_1 -> reg 3
-        | Arg n -> reg (n + 4)
+    | Ret_0 -> v 0
+    | Ret_1 -> v 1
+    | Arg n -> a n
 
 end
 
@@ -74,6 +75,19 @@ let api abi proto =
 
 let dispatch abi sub attrs proto = api abi proto
 
+
+let strip_leading_dot s =
+  match String.chop_prefix s ~prefix:"." with
+  | None -> s
+  | Some s -> s
+
+let demangle demangle prog =
+  Term.map sub_t prog ~f:(fun sub ->
+      let name = demangle (Sub.name sub) in
+      Tid.set_name (Term.tid sub) name;
+      Sub.with_name sub name)
+
+
 let main proj = match Project.arch proj with
     | `mips ->
             info "using MIPS ABI";
@@ -84,8 +98,8 @@ let main proj = match Project.arch proj with
             let api = C.Abi.create_api_processor Abi32.size abi in
             Bap_api.process api;
             let prog = Project.program proj in
+    let prog = demangle strip_leading_dot prog in
             Project.set (Project.with_program proj prog) Bap_abi.name Abi32.name
     | _ -> proj
 
 let setup () = Bap_abi.register_pass main
-
diff --git a/plugins/mips/mips_branch.ml b/plugins/mips/mips_branch.ml
index a050b01..e2031aa 100644
--- a/plugins/mips/mips_branch.ml
+++ b/plugins/mips/mips_branch.ml
@@ -4,8 +4,10 @@ open Mips.Std
 let bal cpu ops =
     let rs = signed cpu.reg ops.(0) in
     let off = signed imm ops.(1) in
+    let step = unsigned const byte 2 in
+    let byte = unsigned const byte 8 in
     RTL.[
-        rs := cpu.cia + unsigned const byte 8;
+        rs := cpu.cia + step * cpu.word_width / byte;
         cpu.jmp (cpu.cia + off);
     ]
 
@@ -54,9 +56,11 @@ let blez cpu ops =
 let bgezal cpu ops =
     let rs = signed cpu.reg ops.(0) in
     let off = signed imm ops.(1) in
+    let step = unsigned const byte 2 in
+    let byte = unsigned const byte 8 in
     RTL.[
-        (* )cpu.lr := cpu.cia + unsigned const byte 8; *)
         when_ (rs >= zero) [
+            cpu.gpr 31 := cpu.cia + step * cpu.word_width / byte;
             cpu.jmp (cpu.cia + off);
         ];
     ]
@@ -67,7 +71,7 @@ let bltzalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt < zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -77,7 +81,7 @@ let blezalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt <= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -87,7 +91,7 @@ let bgezalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt >= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -97,7 +101,7 @@ let bgtzalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt > zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -107,7 +111,7 @@ let beqzalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt = zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -117,7 +121,7 @@ let bnezalc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt <> zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -128,7 +132,7 @@ let bltc cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs < rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -139,7 +143,7 @@ let bgec cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs >= rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -150,7 +154,7 @@ let beqc cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs = rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -161,7 +165,7 @@ let bnec cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs <> rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -172,7 +176,7 @@ let bltuc cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs < rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -183,7 +187,7 @@ let bgeuc cpu ops =
     let off = signed imm ops.(2) in
     RTL.[
         when_ (rs >= rt) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -193,7 +197,7 @@ let bltzc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt < zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -203,7 +207,7 @@ let blezc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt <= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -213,7 +217,7 @@ let bgezc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt >= zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -223,7 +227,7 @@ let bgtzc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rt > zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -233,7 +237,7 @@ let beqzc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rs = zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -243,7 +247,7 @@ let bnezc cpu ops =
     let off = signed imm ops.(1) in
     RTL.[
         when_ (rs <> zero) [
-            cpu.jmp (cpu.cia + off + unsigned const byte 4);
+            cpu.jmp (cpu.cia + off);
         ];
     ]
 
@@ -287,8 +291,10 @@ let jump cpu ops =
 (* JAL target *)
 let jal cpu ops =
     let target = signed imm ops.(0) in
+    let step = unsigned const byte 2 in
+    let byte = unsigned const byte 8 in
     RTL.[
-        cpu.cia := cpu.cia + unsigned const byte 8;
+        cpu.gpr 31 := cpu.cia + step * cpu.word_width / byte;
         cpu.jmp target;
     ]
 
@@ -296,8 +302,10 @@ let jal cpu ops =
 let jalr cpu ops =
     let rd = signed cpu.reg ops.(0) in
     let rs = signed cpu.reg ops.(1) in
+    let step = unsigned const byte 2 in
+    let byte = unsigned const byte 8 in
     RTL.[
-        rd := cpu.cia + unsigned const byte 8;
+        rd := cpu.cia + step * cpu.word_width / byte;
         cpu.jmp rs;
     ]
 
@@ -305,8 +313,10 @@ let jalr cpu ops =
 let jialc cpu ops =
     let rt = signed cpu.reg ops.(0) in
     let off = signed imm ops.(1) in
+    let step = unsigned const byte 2 in
+    let byte = unsigned const byte 8 in
     RTL.[
-        cpu.cia := cpu.cia + unsigned const byte 4;
+        cpu.gpr 31 := cpu.cia + step * cpu.word_width / byte;
         cpu.jmp (rt + off);
     ]
 
@@ -325,6 +335,17 @@ let jr cpu ops =
         cpu.jmp rs;
     ]
 
+(* BNE ra, rb, offset  *)
+let bne cpu ops =
+    let ra = unsigned cpu.reg ops.(0) in
+    let rb = unsigned cpu.reg ops.(1) in
+    let off = signed imm ops.(2) in
+    RTL.[
+        when_ (lnot (ra = rb)) [
+            cpu.jmp (cpu.cia + off)
+        ]
+    ]
+
 let () =
     "BAL" >> bal;
     "BEQ" >> beq;
@@ -359,4 +380,49 @@ let () =
     "JIALC" >> jialc;
     "JIC" >> jic;
     "JR" >> jr;
+    "BNE" >> bne;
+
+
+    (* SLT  v0,v1,v0 *)
+let slt cpu ops =
+  let rd = signed cpu.reg ops.(0) in
+  let rs = signed cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[
+    if_ (rs <$ rt) [
+      rd := one;
+    ] [
+      rd := zero
+    ]
+  ]
+
+(* SLTU  v1,v0,s1 *)
+let sltu cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[
+    if_ (rs < rt) [
+      rd := one;
+    ] [
+      rd := zero
+    ]
+  ]
+
+(* SLTiu  v0,v0,7 *)
+let sltiu cpu ops =
+  let rt = signed cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let im = signed imm ops.(2) in
+  RTL.[
+    if_ (rs < im) [
+      rt := one;
+    ] [
+      rt := zero
+    ]
+  ]
 
+let () =
+  "SLT"   >> slt;
+  "SLTu"  >> sltu;
+  "SLTiu" >> sltiu
diff --git a/plugins/mips/mips_cpu.ml b/plugins/mips/mips_cpu.ml
index 4ae2775..9c70adc 100644
--- a/plugins/mips/mips_cpu.ml
+++ b/plugins/mips/mips_cpu.ml
@@ -52,11 +52,9 @@ let make_cpu addr_size endian memory =
             mips_fail "%s with number %d not found" name n in
     let gpr n = find "GPR" gpri n in
     let fpr n = find "FPR" fpri n in
-    let addr_size = match addr_size with
-    | `r32 -> word
-    | `r64 -> doubleword in
-    let gpr_width = addr_size in
-    { load; store; jmp; cia; addr_size; gpr_width;
+    let word_width, word_bitwidth = match addr_size with
+    | `r32 -> unsigned const byte 32, word
+    | `r64 -> unsigned const byte 64, doubleword in
+    { load; store; jmp; cia; word_width; word_bitwidth;
         reg; gpr; fpr;
     }
-
diff --git a/plugins/mips/mips_model.ml b/plugins/mips/mips_model.ml
index 197b8cb..8423fd3 100644
--- a/plugins/mips/mips_model.ml
+++ b/plugins/mips/mips_model.ml
@@ -114,7 +114,7 @@ module Make_MIPS(S: Spec) : MIPS = struct
 
     let gpr = Array.to_list gprs |> reglist_to_map
     (* FIXME: better representation *)
-    let gpri = make_regs_i (Type.imm fpr_bitwidth) "R" range32
+    let gpri = make_regs_i (Type.imm gpr_bitwidth) "R" range32
 
     module E = struct
         include Exps
@@ -178,4 +178,3 @@ end
 
 module MIPS_32_cpu = Make_cpu(MIPS_32)
 module MIPS_64_cpu = Make_cpu(MIPS_64)
-
diff --git a/plugins/mips/mips_types.ml b/plugins/mips/mips_types.ml
index 52627c6..9055f03 100644
--- a/plugins/mips/mips_types.ml
+++ b/plugins/mips/mips_types.ml
@@ -11,10 +11,9 @@ type cpu = {
     store : exp -> exp -> bitwidth -> rtl;
     jmp : exp -> rtl;
     cia : exp;
-    addr_size : bitwidth;
-    gpr_width : bitwidth;
+    word_width : exp;
+    word_bitwidth : bitwidth;
     reg : (op -> exp) ec;
     gpr : int -> exp;
     fpr : int -> exp;
 }
-
```

XVilka:
I applied your patch, added some comments and a few missing instructions and re-indented all code.

ivg:
An open question about the unaligned access. The right approach is to add an if statement that checks an address and raises a CPU exception if the address is not aligned. This will, however, bloat the control flow graph, as every load instruction will be mapped into a load-or-branch instruction that will result in a couple of basic blocks. 

My suggestion is to ignore this issue right now. But later, after the release add an option to the lifter, that will add this check (through the cpu.load/store/jump operations).

bapdoc.ml
```diff
@@ -27,6 +27,7 @@ let libraries = [
   [
     "bap-arm", "ARM", "ARM architecture";
     "bap-x86-cpu", "X86_cpu", "x86 family";
+    "bap-mips", "MIPS", "MIPS architecture";
```
gitoleg: Can you remove it, please? 
here is said that there is a `bap-mips` library, what actually is not true.
So it will break our documentation.

plugins/mips/mips_arithmetic.ml
```diff
@@ -0,0 +1,234 @@
...
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[ rd := rs + rt; ]
+
+(* CLO rd, rs
+ * Count Leading Ones in Word, MIPS32
+ * Page 136 *)
+(* TODO: check that RT = RD *)
+let clo cpu ops =
+  let rs = unsigned cpu.reg ops.(0) in
+  let rt = unsigned cpu.reg ops.(1) in
```
gitoleg: There are few places like that, when I have something like 
`Warning 26: unused variable rt.`
So could you please fix it by sticking to one of the following: either just remove such lines, or, if you would like to emphasize that the operand is not used, prefix it with `_` : `let _rt = `.

plugins/mips/mips_arithmetic.ml
```diff
@@ -0,0 +1,234 @@
...
+  let rs = signed cpu.reg ops.(1) in
+  let rt = signed cpu.reg ops.(2) in
+  RTL.[ rd := rs + rt; ]
+
+(* ADD.S fd, fs, ft , MIPS32
+ * ADD.D fd, fs, ft , MIPS32
+ * ADD.PS fd, fs, ft , MIPS32, MIPS64
+ * Floating Point Add
+ * Page 37 *)
+let add_fmt cpu ops =
```
gitoleg: I don't think that it's a good idea to add instruction that does nothing (if it isn't `nop` of course). `bap` doesn't support floating point right now, so there is definitely nothing to code in this case. But leave an empty RTL is not correct too because we code a wrong instruction semantic by doing this.  So I suggest to remove it. Or write here something like 
```
RTL.[
   message("add fmt not implemented")
]
```
But in the latter case, we probably should do it for all instructions we didn't implement. So do we really need this insn?

plugins/mips/mips_arithmetic.ml
```diff
@@ -0,0 +1,234 @@
...
+  RTL.[
+    rd := (rs lsl (sa + unsigned const byte 1)) + rt;
+  ]
+
+(* MOVE rd, rs  - WTF nonexistent!? *)
+
+(* SEB rd, rt
+ * Sign-Extend Byte, MIPS32 Release 2
+ * Page 426 *)
+let seb cpu ops =
```
gitoleg: it's a not correct implementation. If I understood it right, we should take a least significant byte from rt, sign-extend it, and store in rs. 
So I would do it something like following:
```
let seb cpu ops =
    let rt = unsigned cpu.reg ops.(0) in
    let rs = signed cpu.reg ops.(1) in
    RTL.[
        rs := last byte rt;
    ]
```

XVilka:
Addressed all the comments.

plugins/mips/mips_logic.ml
```diff
@@ -0,0 +1,182 @@
...
+  let rs = unsigned cpu.reg ops.(1) in
+  let im = unsigned imm ops.(2) in
+  RTL.[
+    rd := rs lxor im;
+  ]
+
+(* BITSWAP rd, rt
+ * Swaps (reverses) bits in each byte, MIPS32 Release 6
+ * Page 95 *)
+let bitswap cpu ops =
```
ivg: We shall try to make this operation without a loop, e.g. as [here][1]. Adding a loop here will increase the complexity of a binary and hamper analysis, though the operation by itself has a very limited side effect, it is really a pain to verify this formally. 


[1]: https://stackoverflow.com/questions/21511533/reverse-integer-bitwise-without-using-loop

plugins/mips/mips_logic.ml
```diff
@@ -0,0 +1,182 @@
...
+      rd := rd lor (((rt lsr cnt) land one)
+                    lsl (unsigned const byte 31 - cnt));
+      cnt := cnt + one;
+    ];
+  ]
+
+(* DBITSWAP rd, rt  - 64bit instruction
+ * Swaps (reverses) bits in each byte, MIPS64 Release 6
+ * Page 95 *)
+let dbitswap cpu ops =
```
ivg: see `bitswap`

plugins/mips/mips_multiply.ml
```diff
@@ -0,0 +1,127 @@
...
+
+(* MUL rd, rs, rt
+ * Multiply Words Signed, Low Word, MIPS32
+ * Page 351 *)
+let mul cpu ops =
+  let rd = signed cpu.reg ops.(0) in
+  let rs = signed cpu.reg ops.(1) in
+  let rt = signed cpu.reg ops.(2) in
+  RTL.[
+    rd := low word (rs * rt);
```
ivg: double checking, @gitoleg are the `low`, `high`, and the rest of word projectors, updated wrt the MIPS ordering?

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+ * Shift Word Left Logical Variable, MIPS32
+ * Page 441 *)
+let sllv cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rt lsl rs;
+    (* TODO: sign extend *)
```
ivg: this TODO comment is misleading, why do you think that the result should be sign-extended?

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+ * Shift Word Left Logical, MIPS32
+ * Page 440 *)
+let sll cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rs lsl sa;
+    (* TODO: sign extend *)
```
ivg: this TODO comment is misleading, why do you think that the result should be sign-extended?

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+  ]
+
+(* SLL rd, rs, imm
+ * Shift Word Left Logical, MIPS32
+ * Page 440 *)
+let sll cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  let tm = unsigned var word in
```
ivg: why do we need a temporary variable here?

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+  ]
+
+(* SLLV rd, rs, rt
+ * Shift Word Left Logical Variable, MIPS32
+ * Page 441 *)
+let sllv cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  let tm = unsigned var word in
```
ivg: this is probably bogus

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+(* SRA rd, rs, imm
+ * Shift Word Right Arithmetic, MIPS32
+ * Page 446 *)
+let sra cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rs lsr sa;
```
ivg: this is not an arithmetic shift.

@gitoleg do you have `asr` in RTL or the operand should me specified as signed for the `lsr` to become an `asr`?

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+(* SRAV rd, rs, rt
+ * Shift Word Right Arithmetic Variable, MIPS32
+ * Page 448 *)
+let srav cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rt lsr rs;
```
ivg: change to `asr` (see the comment above)

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+
+(* DSRA rd, rs, imm
+ * Shift Doubleword Right Arithmetic, MIPS64
+ * Page 201 *)
+let dsra cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  RTL.[
+    rd := rs lsr sa;
```
ivg: change to asr (see the comment above)

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+
+(* DSRA32 rd, rs, imm
+ * Shift Doubleword Right Arithmetic Plus 32, MIPS64
+ * Page 202 *)
+let dsra32 cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  RTL.[
+    rd := rs lsr (sa + unsigned const byte 32);
```
ivg: change to asr (see the comment above)

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+
+(* DSRAV rd, rs, rt
+ * Shift Doubleword Right Arithmetic Variable, MIPS64
+ * Page 203 *)
+let dsrav cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  RTL.[
+    rd := rt lsr rs;
```
ivg: change to asr (see the comment above)

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+ * Shift Word Right Logical, MIPS32
+ * Page 449 *)
+let srl cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let sa = unsigned imm ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rs lsr sa;
+    (* TODO: sign extend *)
```
ivg: there is no need to sign extend here, afaik

plugins/mips/mips_shift_and_rot.ml
```diff
@@ -0,0 +1,239 @@
...
+ * Shift Word Right Logical Variable, MIPS32
+ * Page 450 *)
+let srlv cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  let rt = unsigned cpu.reg ops.(2) in
+  let tm = unsigned var word in
+  RTL.[
+    tm := rt lsr rs;
+    (* TODO: sign extend *)
```
ivg: there is no need to sign extend here, afaik

plugins/mips/mips_store.ml
```diff
@@ -0,0 +1,167 @@
...
+  let rt = signed cpu.reg ops.(0) in
+  let base = signed cpu.reg ops.(1) in
+  let off = signed imm ops.(2) in
+  RTL.[
+    cpu.store (base + off) rt doubleword;
+  ]
+
+(* SCE rt, offset(base)
+ * Store Conditional Word EVA, MIPS32
+ * Page 404 (for real) *)
```
ivg: :rofl:

plugins/mips/mips_logic.ml
```diff
@@ -0,0 +1,182 @@
...
+    rd := rs lor im;
+  ]
+
+(* WSBH rd, rs
+ * Word Swap Bytes Within Halfwords, MIPS32 Release 2
+ * Page 509 *)
+let wsbh cpu ops =
+  let rd = unsigned cpu.reg ops.(0) in
+  let rs = unsigned cpu.reg ops.(1) in
+  RTL.[
```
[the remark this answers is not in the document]
XVilka: You need to enable it in mc - `llvm-mc -arch=mips -mattr=+mips32r6` for example

ivg:
Ok, I'm ready to get it merged, as long as we get at least one smoke test for MIPS32. I'm fine if we will omit the MIPS64 in this version. We can add it later.

PR #767, opened 2018-02-09
fixes mips release 6 problems

gitoleg:
As for now, we have a problem with disassembling and lifting code built for Mips VI release. Although `llvm-mc` works perfectly, if we pass a special argument, e.g. `-mattr=mips32r6`

One possible solution would be to add additional arches for the 6th release like we did it for ARM architecture. E.g., add `mips32r6`. But at first, it's not very beautiful, because we will need to add 4(!!!) architectures: for 32 and 64 bit, for little and big endian. At second, it may hit as in future with new Mips release, whenever it will be. At third, on the `mips` lifter side, this addition will face us with a bunch of new problems, because we will have to handle a bigger amount of architectures.
And at fourth, the most important one. In ARM case, llvm knows how to handle different arm types, there are special modules and functions for this purpose. So, `armv7` is a valid architecture name. But it is not true in Mips case.

A better solution would be to add a command line argument for this purpose.

Let's look how `llvm-mc` creates subtarget:
```
TheTarget->createMCSubtargetInfo(TripleName, MCPU, FeaturesStr));
```
where `FeaturesSet` is exactly the place where "-mattr=mips32r6" goes. And how we do the same:
```
target->createMCSubtargetInfo(triple, cpu, "");
```
So next question is: How to transfer features from command line? But it turned out, that in `llvm-mc`, the `mips32r6` could transfered through `-mcpu` command line argument. At least, llvm tests use it in this way. And what I suggest is just to add the same cmdline argument. Anyway, it already exists in disassembler interface, but not for all functions. So we need just fix it and also add it to `Bap_project` interface.

And, at last, an example:
```
$: echo "00 43 08 9a" | bap-mc --arch=mips --show-insn=asm --show-bil
Invalid instruction at offset 0:
00000000 (00 43 08 9A)                                    |.C..            |
```
```
$: echo "00 43 08 9a" | bap-mc --arch=mips --show-insn=asm --cpu=mips32r6 --show-bil
div $1, $2, $3
{
  AT := V0 /$ V1
}
```
And for little endian it works too:
```
$: echo "9a 08 43 00" | bap-mc --arch=mipsel --show-insn=asm --cpu=mips32r6 --show-bil
div $1, $2, $3
{
  AT := V0 /$ V1
}
```


XVilka:
Will this one make into the release?

ivg:
@XVilka, nope, we will address this issue in 1.5. The solution that Oleg suggests has some issues and, although it works, it actually hides (and makes worse) the underlying problem - our handling of architectures is broken. This PR will break it even more. For example, we may have `--arch=arm --cpu=mips` :)

gitoleg:
I'm closing this PR, a discussion will continue here: https://github.com/BinaryAnalysisPlatform/bap/issues/797

PR #768, opened 2018-02-11
improves relocations handling

gitoleg:
This PR reworks relocations handling in bap.
Previously, we had to support some invariants at different stages of project creation. And it was a kind of breaking of abstractions.  
And now, after https://github.com/BinaryAnalysisPlatform/bap/pull/761  fix, we finally can refactor it. So we rolled back all changes in a disassembler and lifting to IR.  

Also, this PR changes a default brancher, so relocations will work out-of-box.


lib/bap/bap_inject_externals.mli
```diff
@@ -0,0 +1,9 @@
+open Bap_types.Std
+open Bap_image_std
+open Bap_sema.Std
+open Bap_disasm_std
+
+(** [run prg insn spec] performs search for relocations in [spec] and
+    add inject synthetic subroutines for each external function. Also
```
ivg: "add inject" -> inserts or injects

I'm not sure, why you're using word injection for this procedure at all, for me injection is an injective mapping, i.e., when you map one set to another without losing information.

lib/bap/bap_inject_externals.mli
```diff
@@ -0,0 +1,9 @@
+open Bap_types.Std
+open Bap_image_std
+open Bap_sema.Std
+open Bap_disasm_std
+
+(** [run prg insn spec] performs search for relocations in [spec] and
```
ivg: performs search = searches. Keep away from the bureaucratic style

Also, try to document functions, not in terms of how they are implemented, but in terms of what they do, i.e., in a more declarative style. Also, try to minimize the context that is required to understand what the function is doing.

The documentation shall answer the following questions:

1) what is sought in the spec exactly (specify the attributes)
2) what is inserted into the project (explain how functions are named, note that they are tagged with the [synthetic] attribute)
3) what are the assumptions, if any.

lib/bap/bap_project.ml
```diff
@@ -330,7 +319,12 @@ let create_exn
       let disasm = Disasm.create g in
       let program = MVar.read program in
       let program =
-        symbolize_synthetic program (Disasm.insns disasm) spec in
+        if MVar.is_updated spec then
```
ivg: why are you injecting here, instead of putting the code in a plugin and using an external injection point? Are there any invariants on which the core code now depends and which are provided by this injector?

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -48,6 +49,9 @@ let reconstruct name roots cfg =
   let find_block addr =
     Cfg.nodes cfg |> Seq.find ~f:(fun blk ->
         Addr.equal addr (Block.addr blk)) in
+  let find_callers blk =
```
ivg: this function doesn't find anything but maps, the name is misleading.

ivg: also, input edges of a block are not "callers", they are called predecessors. Moreover, there is the `[Cfg.Node.preds]` function for that. 

Please, learn the full interface of a module that you're using.

lib/bap/bap_inject_externals.mli
```diff
@@ -0,0 +1,9 @@
...
+    add inject synthetic subroutines for each external function. Also
+    fixups calls to externals in [prg] *)
+val run : program term -> (mem * insn) seq -> Ogre.doc -> program term
```
ivg: it is totally not obvious and is not reflected in the documentation what is the second argument for.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -62,7 +66,10 @@ let reconstruct name roots cfg =
                       if Block.equal n entry
                       then Cfg.Node.insert n t
                       else return t)) in
-          Symtab.add_symbol syms (name,entry,cfg))
+          let fn = name,entry,cfg in
+          let symtab = Symtab.add_symbol syms fn in
+          Seq.fold (find_callers entry) ~init:symtab
+            ~f:(fun syms a -> Symtab.add_call syms a fn))
```
ivg: This is a wrong assumption, that all predecessors of an entry block are callers. A function can have a back edge that points to the entry block. E.g.,



```
digraph {
    entry -> check;
    check -> exit;
    check -> entry;
}
```

lib/bap_disasm/bap_disasm_symtab.ml
```diff
@@ -28,6 +28,7 @@ module Fn = Opaque.Make(struct
 type t = {
   addrs : fn Addr.Map.t;
   names : fn String.Map.t;
+  calls : fn Addr.Map.t;
```
ivg: what is the justification to put this into the Symtab? This adds a completely new responsibility for this data structure, and we need to review each function in its implementation that it preserves its invariants. For example, you at least broke the `remove` function. The mere fact that you didn't touch it means, that it is broken, as what will happen if you will remove a function that is in the calls?

lib/bap_sema/bap_sema_lift.ml
```diff
@@ -49,24 +49,22 @@ let label_of_fall cfg block =
...
-let linear_of_stmt ?addr return insn stmt : linear list =
+let linear_of_stmt ?addr ?dst return insn stmt : linear list =
   let (~@) t = match addr with
     | None -> t
     | Some addr -> annotate_addr (annotate_insn t insn) addr in
   let goto ?cond id =
     `Jmp ~@(Ir_jmp.create_goto ?cond (Label.direct id)) in
   let jump ?cond exp =
-    let target = Label.indirect exp in
+    let target = match dst with
```
ivg: that doesn't look good or right. Even if it is correct, it should be rewritten in a good way.

The jump function takes exp and returns a corresponding term. Here, you're ignoring the exp if the `[dst]` parameter is passed. In this case, all jumps will lead to the same destination. 

This might be correct if `[dst]` is passed only if the statement has at most one jump. But this is a kind of precondition that you shall never enforce. It is broken.

lib/bap_sema/bap_sema_lift.ml
```diff
@@ -158,12 +156,20 @@ let is_conditional_jump jmp =
   Insn.(may affect_control_flow) jmp &&
   has_jump_under_condition (Insn.bil jmp)
 
-let blk cfg block : blk term list =
+let find_dst block symtab =
+  if Insn.(is call) (Block.terminator block) then
+    Symtab.find_call symtab (Block.addr block) |>
```
ivg: A call usually refers to a call site, i.e., a place where the call occurs, e.g., the `call` assembly instruction, the `call` term in BIR, or a IRL call. In your case, the call is the Block.terminator which is a call. 
What you're finding is called _callee_, I believe, though it is hard to understand, because there is no documentation for the new Symtab functions. And they are very ambiguous and easy to use incorrectly.

plugins/relocatable/rel_brancher.ml
```diff
@@ -33,48 +32,45 @@ module Rel = struct
   let relocations =
     addr_width >>= fun width ->
     relocations >>= fun rels ->
-    external_symbols >>= fun ext ->
-    Fact.return (of_aseq width rels, of_aseq width ext)
+    external_symbols >>= fun exts ->
+    let to_addr = Addr.of_int64 ~width in
+    Fact.return (of_seq rels ~fkey:to_addr ~fdata:to_addr,
```
ivg: it would be better to first map both sequences, and the use the normal `of_aseq`

plugins/relocatable/rel_brancher.ml
```diff
@@ -33,48 +32,45 @@ module Rel = struct
...
+                 of_seq exts ~fkey:to_addr ~fdata:ident)
 
 end
 
-let width_of_mem m = Word.bitwidth (Memory.min_addr m)
-
 let get mem data =
-  let rec find addr =
+  let rec loop addr =
```
ivg: `find` was a much better name

lib/bap_disasm/bap_disasm_symtab.ml
gitoleg: Justification is:
we lose information about calls in reconstructor when we take apart a whole program cfg, that is very harmful in case of relocations. And the `Symtab` looks like a good place to store this information. 
And we definitely need to review each function, what I could do incorrect, but `remove` function, as far as I see, removes calls, therefore I did touch it

gitoleg:
I placed most of the relocations related stuff to a plugin and cleaned code from previous decisions. So neither `Bap_project`, nor `Bap_sema_lift`, nor `Bap_disasm_rec`
are not taking in account relocations at all.
And what do we have now to handle relocations :
1) `relocatable` brancher is still the same as it was before, with just a small portion of 
  refactoring
2) a new pass that substitutes any `jmp` term with a call to  a subroutine in case the address 
    of the `jmp` corresponds to the adderess of a relocation of any kind. And in case of
   calls to external functions a synthetic subroutine created.

lib/bap_disasm/bap_disasm_symtab.ml
ivg: So, you actually did touch it. 
What do you mean by "harmful in case of relocations"?
And yes, I agree, that we lose information when we proceed from the whole program CFG to a set of function CFGs, and since we pass only symtab downstream we need to preserve this information. 

I have a couple more questions:

Does the `calls` data structure relates callsites to the called function?
Is it guaranteed that domain of call is equal to the domain of addrs ? (I.e., that forall x, exists y, calls[x] = addrs[y])

ivg:
Can you also implement #858 in this PR so that we do not produce intermediate results? I mean, that we can also get rid of rela_brancher in this PR, what do you think?

gitoleg:
so, what do we do: 
1) remove the relocatable brancher, because we finally undestand that this guy actually fixes a behaviour of the main brancher, and we need just to fix main. So relocations handling is a part of his job now.

2) add a relocatable symbolizer, that uses an information about relocations and instructions to provide names for external functions (by address of a caller side)

3) in `symtab` we we add `add_callee` and `find_callee` functions to store information about caller addresses and callees names.

4) in `reconstructor` we add callees names to `symtab`: for each call we name the destination side. And if the destination is unknown (no output edges or only fall edge), we just name the caller's address. The latter is most likely occurs in case of external functions.

5) in `sema lifter` we take a look at every indirect jump. We check if there is a known callee name at the address of the jump (`Symtab.fin_callee`). And if there is, then there are two options:
    - If there is a function with the given name in Symtab  
      (`Symtab.find_by_name`), then we can say that this indirect jump can be replaced as a call to  this function
    - If there is no such function, then we can say that there is a call to
   external and we create a synthetic subroutine then.

lib/bap_disasm/bap_disasm_symtab.mli
```diff
@@ -20,3 +20,6 @@ val dominators : t -> mem -> fn list
 val intersecting : t -> mem -> fn list
 val to_sequence : t -> fn seq
 val span : fn -> unit memmap
+
+val add_callee : t -> addr -> string -> t
```
ivg:
```ocaml
(* remembers a call to an external function from the given block *)
val add_external : t -> block -> string -> t

(* finds if there are any calls to an external symbol from the given block *)
val find_external : t -> addr -> string option
```

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -46,39 +46,68 @@ let is_start roots cfg blk =
...
-          else entries)
+          if is_fall e then entries
+          else Set.add entries (Cfg.Edge.dst e))
   else entries
 
-let collect_entries cfg roots =
-  let roots = Addr.Set.of_list roots in
-  Seq.fold (Cfg.nodes cfg) ~init:Block.Set.empty
-    ~f:(entries_of_block cfg roots)
+let terminator_addr blk =
```
ivg: use the block address as a designator of a callsite. A basic block is a unit of a control flow relation not a specific instruction.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -46,39 +46,68 @@ let is_start roots cfg blk =
...
+  List.rev (Block.insns blk) |>
+  List.hd_exn |>
+  fst |>
+  Memory.min_addr
+
+let maybe_external blk cfg =
```
ivg: `is_unresolved` might be a better name.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -46,39 +46,68 @@ let is_start roots cfg blk =
...
+  let deg = Cfg.Node.degree ~dir:`Out blk cfg in
+  deg = 0 ||
+  (deg = 1 && is_fall (Seq.hd_exn (Cfg.Node.outputs blk cfg)))
+
+let add_callees syms name cfg blk =
+  if is_call blk then
+    let call_addr = terminator_addr blk in
+    if maybe_external blk cfg then
```
ivg: We're here assuming that any unresolved call is an external call. I think that it is too overly, we need somehow to limit the set of presumed externals. So, we will consider the call to be external if it has a resolved name, i.e., something that is not `sub_XXX`. Of course, we won't look at the name for that, instead we will extend the symbolizer interface with a new function `is_resolved` that will return true if a symbolizer returned `Some name`.

lib/bap_disasm/bap_disasm_reconstructor.ml
ivg: Is this code dead now? If yes, them remove it, otherwise make sure that it is dead and still remove.

lib/bap_disasm/bap_disasm_brancher.ml
```diff
@@ -40,9 +114,7 @@ let has_jumps =
     end)
 
 let rec dests_of_bil bil : dests =
-  if has_jumps bil then
-    Bil.fold_consts bil |> List.concat_map ~f:dests_of_stmt
-  else []
+  Bil.fold_consts bil |> List.concat_map ~f:dests_of_stmt
```
ivg: do we really need to do this after the BIL transformation pipeline PR is merged?

plugins/relocatable/rel_symbolizer.ml
```diff
@@ -0,0 +1,73 @@
...
+        | _ -> fun _ -> None in
+      Ok (Symbolizer.create name))) |>
+  Symbolizer.Factory.register "relocatable"
+
+let () =
+  init ();
+  Config.manpage [
+    `S "SYNOPSIS";
+    `S "DESCRIPTION";
+    `P "Provides a symbolizer for external symbols in relocatable files.";
```
ivg:
```ocaml
`P "Extracts symbol information from the program relocations. "

`P "The relocation symbolizer leverages the relocation information stored in files
to extract symbol names. Since a relocation references an external symbol which
doesn't have an address we use an address of a callsite. "
```

plugins/relocatable/rel_symbolizer.ml
```diff
@@ -0,0 +1,73 @@
...
+        address where fixup should be done, and this address
+        doesn't match with an address of any instruction, because it
+        points somewhere to the middle of the instruction bytes, after
+        prefixes, opcodes and etc. Relocatable symbolizer uses OGRE
+        to query relocations and disassembled instructions to find
+        an instruction address of external function calls.";
+    `S "SEE ALSO";
+    `P "$(b,bap-plugin-llvm)(1) code";
+  ];
+  Config.when_ready (fun _ -> ())
```
ivg: you shall call `init` on `when_ready`, so it should be `Config.when_read init`

PR #781, opened 2018-02-21
moves to travis stages

gitoleg:
Starting from this PR our CI will get profit from `travis` stages feature. 
So what do we have:
- three parallel builds, that just compile and install `bap` for each supported compiler. One of the builds also warms up a cache for next stages;
 - parallel stages for unit tests and `testsuite` checks.

### DIFF
--- a/.run_travis_tests.sh
+++ b/.run_travis_tests.sh
@@ -1,0 +1,20 @@
+TASK=$1
+
+eval `opam config env`
+
+comp=`opam config var switch`
+
+cp -r $HOME/save_opam/bin/* $HOME/.opam/$comp/bin/
+cp -r $HOME/save_opam/share/* $HOME/.opam/$comp/share/
+cp -r $HOME/save_opam/lib/* $HOME/.opam/$comp/lib/
+
+bap --version
+bap-byteweight update
+
+if [ "$TASK" == "checks" ]; then
+    bash -exc 'make check'
+fi
+
+if [ "$TASK" == "unit_tests" ]; then
+    bap_run_tests
+fi
--- a/.run_travis_tests.sh
+++ b/.run_travis_tests.sh
@@ -4,9 +4,9 @@ eval `opam config env`
 
 comp=`opam config var switch`
 
-cp -alu $HOME/save_opam/bin/* $HOME/.opam/$comp/bin/
-cp -alu $HOME/save_opam/share/* $HOME/.opam/$comp/share/
-cp -alu $HOME/save_opam/lib/* $HOME/.opam/$comp/lib/
+cp -r $HOME/save_opam/bin/* $HOME/.opam/$comp/bin/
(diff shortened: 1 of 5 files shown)


.run_travis_tests.sh
```diff
@@ -0,0 +1,20 @@
+TASK=$1
+
+eval `opam config env`
+
+comp=`opam config var switch`
+
+cp -r $HOME/save_opam/bin/* $HOME/.opam/$comp/bin/
```
ivg: you can use `cp -al` instead of `-r` that shall be much faster.

.travis.yml
```diff
@@ -35,11 +29,29 @@ addons:
     - dejagnu
     - jq
     - opam
+
+cache:
+  directories:
+  - $HOME/.opam
+  - $HOME/save_opam
```
ivg: why do we need to copy from `save_opam` to `.opam` if they both are cached?

.travis_install.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ "$WITH_BUILD_CACHE" == "true" ]; then
+    export POST_INSTALL_HOOK='
+x=`ocamlfind query bap`
+if [ $x != "" ]; then
+   echo START TO PACK BAP
+   OPAM_SWITCH=`opam config var switch`
+   mkdir -p $HOME/save_opam
+   mkdir -p $HOME/save_opam/lib
+   cp -r $HOME/.opam/$OPAM_SWITCH/lib/bap* $HOME/save_opam/lib/
```
ivg: again, `cp -al` shall be faster.

Makefile
```diff
@@ -39,12 +39,12 @@ distclean:
 .PHONY: check
 
 test: build
-ifeq ("$(BAP_RUN_TEST)","true")
+ifneq ("$(BAP_NO_TESTS)","true")
```
ivg: are we using this environment variables at all? If not, then let's just drop this ifelses from the Makefile.

.travis.yml
gitoleg: We can environment variables and add `save_opam` to path lookup. Also, we can change tests in `bap-testsuite`, e.g. for `run` tests, we will have to add where is a `site-lisp` directory and so on. But I don't think it's a good solution.

Makefile
gitoleg: yes, it's not hard to get rid of them

.travis.yml
gitoleg: Ah, better to answer that `bap` is removed while caching of `.opam`, because 
travis does rather a straightforward sequence of actions: install the package and then remove the package. So the only place we can catch this - by setting `POST_INSTALL_HOOK` variable, to save all we can

ivg: yeah, the last answer clarifies it to me.

Makefile
ivg: So let's drop them, and also let's try to add `-al` just curious if it will work.

PR #787, opened 2018-03-05
updates CI tools

ivg:
This PR adds 7 docker files and 3 vagrant files that are using in our
CI system to test bap installability.

Also this PR updates our `opam-release.sh` script, that is now capable
of performing the whole update procedure of the OPAM repo in fully
automated way (you need to create only PR manually)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ derivatives. For other distributions we provide tgz archives. To
 install bap on a Debian derivative:
 
 ```bash
-wget https://github.com/BinaryAnalysisPlatform/bap/releases/download/v1.3.0/{bap,libbap,libbap-dev}_1.3.0.deb
-sudo dpkg -i {bap,libbap,libbap-dev}_1.3.0.deb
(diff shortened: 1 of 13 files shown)


gitoleg:
Can you provide some examples for new `opam-release`? Given that we have next tree of packages:
```
packages
    bare 
         bare.1.0
         bare.master
    ogre
          ogre.master
    bap-std
          bap-std.1.4.0
          bap-std.master
    conf-bap-llvm
           conf-bap-llvm.1.1
```
How do we call? What output we will get?

ivg:
we call it just as `opam-release.sh release` (though instead of `release` you can pass any other function name to call it directly, or not to pass anything, so that you can source it if you like). 

Concerning your, broken, example, this is what we will get (give the current state of our github repostories):

```
packages
    bare 
         bare.1.0
         bare.1.4.0
    ogre
          ogre.1.4.0
    bap-std
          bap-std.1.4.0
    conf-bap-llvm
           conf-bap-llvm.1.1
```

and `bap-std.1.4.0` could be modified if there was a difference between the master version and the `1.4.0` version. The `master`  version will take the precedence. 

The result of the script in real life could be found here: https://github.com/ocaml/opam-repository/pull/11531/files

PR #789, opened 2018-03-12
fixes BIL normalization procedure

ivg:
As it was observed and reported by @ethan42 the normalization procedure may duplicate and rearrange expressions that have observable effects (i.e., expressions that are non-generative in our parlance), that may change semantics of normalized expressions. The root of the problem is the presence of expressions, that have side effects, such as load and store expressions. This PR resolves this issues by restricting the First BIL Normal Form (BNF1) from having non-generative expressions in certain positions, exactly those that would be duplicated during the BNF2 normalization. The absence of non-generative expressions in the duplication points is ensured by hoisting them into separate statements, while preserving the order and the number of operations. Since, this transformation produces BIL statements we can't apply it purely on the expression language, thus function `Exp.normalize` is defined only for expressions in BNF1, that is now stated explicitly in the comment.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -2127,10 +2127,7 @@ module Std : sig
         preceding statement or is not bound with [let] expression *)
     val free_vars : stmt list -> vars
 
-    (** [fold_consts] evaluate constant expressions.
-        Note: this function performs only one step, and has no loops,
-        it is supposed to be run using a fixpoint combinator.
-    *)
+    (** [fold_consts] evaluates constant expressions and statements. *)
     val fold_consts : stmt list -> stmt list
 
     (** [fixpoint f] applies transformation [f] until fixpoint is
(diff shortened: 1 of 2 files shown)


dbrumley:
It's not clear to me where BNF1 and BNF2 are defined. Also, could you add some language that says what it means to be generative?

ivg:
@dbrumley, BNF1 and BNF2 are defined in the [BAP documentation][1] and [normalization was introduced][2] in the 1.3 release. 

> Also, could you add some language that says what it means to be generative?

In fact it is in the PR description the first sentence, but I probably used to many indirection references so that it become non clear.  Those expressions that have observable side-effects are called non-generative. Conversely, those expressions that do no have any observable side-effects are generative. The term is not coined by me, but I believe that it means that such expressions could generated, replicated, without affecting the semantics of a program, hence the name. 

[1]: http://binaryanalysisplatform.github.io/bap/api/v1.4.0/Bap.Std.Stmt.html#VALnormalize
[2]: https://github.com/BinaryAnalysisPlatform/bap/pull/688#bil-normal-form-and-normalization-utilities

PR #790, opened 2018-03-12
bumps the bap version to 1.5.0-dev

ivg:
to prevent the confusion with the already released version


ivg:
we need to merge this PR before any other PRs.

PR #794, opened 2018-03-14
fixes relative addresses in llvm backend

gitoleg:
There was a kind of bug in llvm backend: relative address for any binary entry was defined as non-negative value. Probably, it wasn't a very bad choice, since we don't want to
deal with negative addresses. But from another point of view, we always convert relative address to absolute and negative address should not happen. And we actually inject wrong
information in ogre when storing non-negative numbers: converting them to absolute addresses assigns a real value to every entry, dependless if it had a meaningful address or not at all.

### DIFF
--- a/lib/bap_llvm/llvm_primitives.cpp
+++ b/lib/bap_llvm/llvm_primitives.cpp
@@ -12,7 +12,7 @@ std::string arch_of_object(const llvm::object::ObjectFile &obj) {
     return Triple::getArchTypeName(static_cast<Triple::ArchType>(obj.getArch()));
 }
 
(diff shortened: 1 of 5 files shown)


lib/bap_llvm/llvm_primitives.cpp
```diff
@@ -13,8 +13,7 @@ std::string arch_of_object(const llvm::object::ObjectFile &obj) {
 }
 
 uint64_t relative_address(uint64_t base, uint64_t abs) {
-    if (abs >= base) return (abs - base);
-    else return abs;
+    return (abs - base);
```
ivg: shouldn't the return type of the function be fixed also?

PR #795, opened 2018-03-15
adds llvm 6.0 support

gitoleg:
adds `llvm 6.0` support


lib/bap_llvm/llvm_binary_38_40.hpp
```diff
@@ -1,5 +1,5 @@
 #if LLVM_VERSION_MAJOR == 3 && LLVM_VERSION_MINOR == 8          \
-    || LLVM_VERSION_MAJOR == 4 || LLVM_VERSION_MAJOR == 5
+    || LLVM_VERSION_MAJOR >= 4
```
ivg: let's not be that optimistic, and either use an interval, e.g., `LLVM_VERSION_MAJOR >= 4 && LLVM_VERSION_MAJOR < 7` or just enumerate them. I'm for the former.

gitoleg: It's as you wish, but  I perform a check anyway - just once:
```

Issue #776, opened 2018-02-16
the `--api-path` parameter doesn't understand dots (`.`)


ivg:
It convenient to create a folder with API specification directly in the working directory. However, when we specify the current working directory using the dot notation, bap fails with an exception:
```
$ bap bin/exe --api-path=. 
Uncaught exception:
  
  FilePath.EmptyFilename

Raised at file "src/FilePath.ml", line 246, characters 12-31
Called from file "plugins/api/api_main.ml", line 77, characters 20-57
Called from file "plugins/api/api_main.ml", line 80, characters 54-68
Called from file "plugins/api/api_main.ml", line 84, characters 19-36
Called from file "src/list.ml", line 693, characters 12-16
Called from file "src/list.ml" (inlined), line 700, characters 26-47
Called from file "plugins/api/api_main.ml", line 95, characters 4-58
Called from file "src/list.ml", line 322, characters 13-17
Called from file "src/list.ml" (inlined), line 346, characters 15-31
Called from file "plugins/api/api_main.ml", line 260, characters 2-38
Called from file "plugins/api/api_main.ml", line 286, characters 12-30
Called from file "plugins/api/api_main.ml", line 379, characters 4-22
```

A workaround is to use `pwd`, 

```
bap bin/exe --api-path=`pwd`
```

But it is awkward, not portable, and doesn't work in the environments without the expansion, e.g., we can't use it in BAP recipes

PR #796, opened 2018-03-16
fixes paths comprehension in api plugin

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/776

So everybody is welcome to set `api` path like `$ bap bin/exe --api-path=. `


ivg:
will `bap bin/exe --api-path=./hello` work with this fix?

gitoleg:
I would say that your example will work even without this fix, i.e. it worked before.

ivg:
Good, that's what I was expecting. Thanks

PR #799, opened 2018-03-21
improves macho loader

gitoleg:
Previously, only little endian binaries were expected in our macho loader, so this caused unexpected behavior with big endian binaries. This PR fix it, as well adds some more guards for reliability.

Also, this PR fixes problem with an entry point for old macho binaries: they just don't contain so-called `LC_MAIN` command, and entry point as part of the binary structure is absent.
But we can workaround this problem by emitting entry point as an address of a first executable section.

### DIFF
--- a/lib/bap_llvm/llvm_macho_loader.hpp
+++ b/lib/bap_llvm/llvm_macho_loader.hpp
@@ -155,8 +155,7 @@ void entry_point(const macho &obj, ogre_doc &s) {
     uint64_t entry = max;
     for (auto sec : prim::sections(obj)) {
         auto addr = prim::section_address(sec);
-        if (!addr) continue;
-        if (section_flags(obj, sec) & MachO::S_ATTR_PURE_INSTRUCTIONS)
+        if (addr && section_flags(obj, sec) & MachO::S_ATTR_PURE_INSTRUCTIONS)
             entry = std::min(*addr, entry);
     }
     if (entry != max)
@@ -311,40 +310,42 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
         macho_command(obj, cmd, s);
 }
 
-MachO::relocation_info get_rel(const macho &obj, uint32_t off) {
+// Size in bytes of relocation entry. It coulbe either size of
+// relocation_info, scattered_relocation_info, any_relocation_info
+#define RELOCATION_ENTRY_SIZE 8
+
+error_or<MachO::relocation_info> get_rel(const macho &obj, uint32_t off) {
     const char *ptr = prim::get_raw_data(obj) + off;
-    MachO::relocation_info rel;
+    MachO::any_relocation_info any;
+    std::copy_n(ptr, RELOCATION_ENTRY_SIZE, reinterpret_cast<char *>(&any));
     if (obj.isLittleEndian() != sys::IsLittleEndianHost) {
-        MachO::any_relocation_info any;
-        memcpy(&any, ptr, sizeof(MachO::any_relocation_info));
         MachO::swapStruct(any);
-        memcpy(&rel, &any, sizeof(MachO::relocation_info));
-    } else {
-        memcpy(&rel, ptr, sizeof(MachO::relocation_info));
     }
-    return rel;
+    if (!obj.is64Bit() && obj.isRelocationScattered(any)) {
+        return failure("scattered relocations are not supported");
+    } else {
+        MachO::relocation_info rel;
+        std::copy_n(reinterpret_cast<char *>(&any),
+                    RELOCATION_ENTRY_SIZE,
+                    reinterpret_cast<char *>(&rel));
+        return success(rel);
+    }
 }
 
-// Note, that is r_extern is set to 1, than r_symbolnum contains an index in symbtab, and
-// section index otherwise
+//Note, that if r_extern is set to 1, then r_symbolnum contains an index in symbtab, and
+//section index otherwise
 void iterate_dyn_relocations(const macho &obj, uint32_t off, uint32_t num, ogre_doc &s) {
-    uint32_t next = off;
     for (std::size_t i = 0; i < num; ++i) {
-        MachO::relocation_info rel = get_rel(obj, next);
-        if (rel.r_address & MachO::R_SCATTERED) {
-            // not supported
-            next = next + sizeof(MachO::scattered_relocation_info);
-        } else {
-            if (rel.r_extern)
-                symbol_reference(obj, rel.r_symbolnum, rel.r_address, s);
-            next = next + sizeof(MachO::relocation_info);
-        }
+        auto rel = get_rel(obj, off + i * RELOCATION_ENTRY_SIZE);
+        if (rel && rel->r_extern)
+            symbol_reference(obj, rel->r_symbolnum, rel->r_address, s);
     }
 }
 
 void indirect_symbols(const macho &obj, const MachO::dysymtab_command &dlc, ogre_doc &s) {
     auto base = image_base(obj);
     uint64_t obj_size = obj.getData().size();
+    MachO::symtab_command symtab = obj.getSymtabLoadCommand();
     for (auto sec : prim::sections(obj)) { // sections
         auto typ = section_type(obj, sec);
         if (typ == MachO::S_NON_LAZY_SYMBOL_POINTERS ||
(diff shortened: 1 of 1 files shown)


lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -152,6 +137,38 @@ MACHO_SECTION_FIELD(reserved2, uint32_t)
...
+            const MachO::entry_point_command *entry_cmd =
+                reinterpret_cast<const MachO::entry_point_command*>(info.Ptr);
+            uint64_t entry = entry_cmd->entryoff;
+            if (obj.isLittleEndian() != sys::IsLittleEndianHost)
+                sys::swapByteOrder(entry);
+            s.entry("entry") << entry;
+            return;
+        }
+    }
+    std::vector<uint64_t> addrs;
```
ivg: Big binaries can have quite a few section, so I don't think that we should use such inefficient algorithm. You can use `std::optional` if you're afraid of using explicit sentinel values, such as `max_int`.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -152,6 +137,38 @@ MACHO_SECTION_FIELD(reserved2, uint32_t)
...
+            addrs.push_back(*addr);
+    }
+    std::sort(addrs.begin(), addrs.end());
+    if (addrs.size() > 0)
+        s.entry("entry") << addrs.front();
+}
+
+void image_info(const macho &obj, ogre_doc &s) {
+    if (!is_exec(obj))
+        s.raw_entry("(entry 0)");
```
ivg: why is it 0, but not the lowest address? If it is a library then will this take into account the rebasing?

lib/bap_llvm/llvm_macho_loader.hpp
ivg: it looks like that `optional` was added only in std+17, so you can't use it. So either use a sentinel or a boolean flag.

lib/bap_llvm/llvm_macho_loader.hpp
gitoleg: it equals to zero because we use relative addresses. so after rebasing entry point will be equal to the lowest address

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -152,6 +139,37 @@ MACHO_SECTION_FIELD(reserved2, uint32_t)
...
+            const MachO::entry_point_command *entry_cmd =
+                reinterpret_cast<const MachO::entry_point_command*>(info.Ptr);
+            uint64_t entry = entry_cmd->entryoff;
+            if (obj.isLittleEndian() != sys::IsLittleEndianHost)
+                sys::swapByteOrder(entry);
+            s.entry("entry") << entry;
+            return;
+        }
+    }
+    uint64_t entry = INT_MAX;
```
ivg: The `INT_MAX` is not the maximum value of the `uint64_t` type, it is 

```
#include <limit>
std::numeric_limits<uint64_t>::max();
```

gitoleg:
So, finally,  I wrote `swap`  for the `relocation_info` struct and it is really better (and correct) than it was earler. Also added few comments for future readers.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,24 +311,40 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
+
+// Note, that is r_extern is set to 1, than r_symbolnum contains an index in symbtab, and
+// section index otherwise
+void iterate_dyn_relocations(const macho &obj, uint32_t off, uint32_t num, ogre_doc &s) {
+    uint32_t next = off;
     for (std::size_t i = 0; i < num; ++i) {
-        const MachO::relocation_info *rel = reinterpret_cast<const MachO::relocation_info *>(ptr + next);
-        if (rel->r_address & MachO::R_SCATTERED) {
+        MachO::relocation_info rel = get_rel(obj, next);
+        if (rel.r_address & MachO::R_SCATTERED) {
```
ivg: `rel.r_address & MachO::R_SCATTERED` doesn't imply that the relocation `rel` is scattered, it also depends on the architecture, e.g., for x86_64 and arm64 the relocation would be an immediate value, i.e., 64 bit, so this iteration will go off track in case of 64 bit system. If needed I can give the same testing binary for the x86_64 arch.

There is a `isRelocationScattered` member-function of the `MachOObjectFile` class that can tell you for sure if it scattered or now. Is there any _good_ reason why you are not using this function? Also there are plenty of relocation related member functions that apparently takes the endianness into account. Why aren't you using them?

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -336,13 +368,17 @@ void indirect_symbols(const macho &obj, const MachO::dysymtab_command &dlc, ogre
             for (uint32_t j = 0; j < count && n + j < dlc.nindirectsyms; j++) {
                 auto i_addr = addr + j * stride;
                 auto i_offs = offs + j * stride;
-                uint32_t indirect_symbol = obj.getIndirectSymbolTableEntry(dlc, n + j);
+
+                uint64_t index = n + j;
+                uint64_t offset = dlc.indirectsymoff + index * sizeof(uint32_t);
+                if (offset > obj_size)
```
ivg: what if `offset` is equal to `obj_size`?

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -351,13 +387,19 @@ void indirect_symbols(const macho &obj, const MachO::dysymtab_command &dlc, ogre
     } // sections
 }
 
+// This part is mainly for bundles, since it's not enough just to iterate over
+// sections in order to get relocations, e.g. like we do it for MH_OBJECT files.
+// We are going to take a look only to indirect symbols and external relocations.
```
ivg: s/to/for

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -351,13 +387,19 @@ void indirect_symbols(const macho &obj, const MachO::dysymtab_command &dlc, ogre
...
+// Speaking about local relocations references from dyn sym tab, even modern (6.0)
+// does not perform a lot of job about them, so let's do the same. One cat take a look at
```
ivg: s/cat/can

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,24 +311,40 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
+            if (rel.r_extern)
+                symbol_reference(obj, rel.r_symbolnum, rel.r_address, s);
             next = next + sizeof(MachO::relocation_info);
         }
     }
 }
 
 void indirect_symbols(const macho &obj, const MachO::dysymtab_command &dlc, ogre_doc &s) {
-
+    auto base = image_base(obj);
```
ivg: from here and below there is a lot of voluntarism with the handling  of the raw `base` pointer. Please, use some smart pointer, that will bound check every dereference. If you don't find any in C++ stdlib or LLVM, then implement your own.

gitoleg:
I will try to avoid to use the word `finally`) but anyway I like this implementation.

### Summary
I added more comments, and really hope they will be helpful.  Also added more guards and checks, so bug `shall not pass`, and refactored and simplified the code. There is only one place when we work directly with a raw pointer to object data, and I guarded it with checks too. 

### Relocations structures and endianess.
It turned out, that we should not use that `relocation_info` structure at all. I wrote a detailed explanation directly in code, but briefly - there are not any easy ways to read bit fields from it. We have to use masks, swaps, shifts. And actually, everything is already done in `llvm`, so we just need to use it. 

In the end, I checked everything on different files - little and big endian, and now I can say that it works like we expect.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,69 +313,136 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
-        if (rel->r_address & MachO::R_SCATTERED) {
-            // not supported
-            next = next + sizeof(MachO::scattered_relocation_info);
-        } else {
-            if (rel->r_extern)
-                symbol_reference(obj, rel->r_symbolnum, rel->r_address, s);
-            next = next + sizeof(MachO::relocation_info);
-        }
+        auto rel = get_rel(obj, off + i * RELOCATION_ENTRY_SIZE);
+        if (rel && obj.getPlainRelocationExternal(*rel))
```
ivg: the `getPlainRelocationExternal` function depends on the endiannes, however, you already swapped the relocation bytes. Do you think that it is correct? I would suggest you to re-implement those LLVM functions and do not rely on them. Looks like this part of LLVM is buggy and was never tested, probably it is just easier to reimplement this stuff here.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,69 +313,136 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
-                    if (auto name = prim::symbol_name(*Sym)) {
-                        s.entry("symbol-entry") << *name << i_addr << stride << i_offs ;
-                        s.entry("code-entry") << *name << i_offs << stride;
+    // this is actually a check that we will not get out of object
+    // bounds while calling getIndirectSymbolTableEntry.
+    // llvm does approximately the same to raise a nasty
+    // runtime LLVM_ERROR
+    uint64_t offset = dlc.indirectsymoff + index * sizeof(uint32_t);
+    symbol_iterator s = prim::end_symbols(obj);
+    if (offset < obj_size)
```
ivg: again, you can't compare offsets and sizes.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,69 +313,136 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
+        s = get_symbol(obj, obj.getIndirectSymbolTableEntry(dlc, index));
```
ivg: instead of the pattern:
```
x = invalid;
if (valid(s)) {
  x = make(s);
  return x;
}
return x;
```

you can use much more safe and readable pattern:
```
if (valid(s)) 
    return make(s);
else
   return invalid;
```

This will make your code easier to understand (you do not need to track variables and their flow) and sustainable to changes (the invalid value doesn't roam in the scope waiting to be dereferenced).

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,69 +313,136 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
+    return s;
+}
+
+// returns stride for indirect symbols entries
+uint32_t indirect_symbols_stride(const macho &obj, const SectionRef &sec) {
```
ivg: an easier to understand and less error prone version:
```C++
uint32_t indirect_symbols_stride(const macho &obj, const SectionRef &sec) {
    if (section_type(obj, sec) == MachO::S_SYMBOL_STUBS)
        return section_reserved2(obj, sec);
    else
        return obj.is64Bit() ? 8 : 4;
}
```

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,69 +313,136 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
+        stride = section_reserved2(obj, sec);
+    else
+        stride = obj.is64Bit() ? 8 : 4;
+    return stride;
+}
+
+// true for symbol stubs or symbol pointers sections
+bool contains_sym_stubs(const macho &obj, const SectionRef &sec) {
+    auto typ = section_type(obj, sec);
+    return
```
ivg: you do not need to put parentheses around the return expression. You can also use switch/case

```c++
switch (section_type(obj, sec)) {
  case X:
  case Y:  
  case Z:   return true;
  default:  return false;
}
```

That's is very close to OCaml's match.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -376,7 +461,13 @@ void symbols(const macho &obj, ogre_doc &s) {
     symbols(obj, sizes, s);
 }
 
+// It's safe to call getSymtabLoadCommand without any checks,
```
ivg: is it true for all versions? I remember that somewhere in LLVM 3.4 we were getting a segmentation fault when we were asking for a symtab. It might be in ELF backend though.

gitoleg: it's true for llvm versions >= 3.8

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -295,70 +329,122 @@ void iterate_macho_commands(const macho &obj, ogre_doc &s) {
...
-void iterate_dyn_relocations(const macho &obj, std::size_t off, std::size_t num, ogre_doc &s) {
-    const char *ptr = prim::get_raw_data(obj);
-    std::size_t next = off;
+bool is_rel_scattered(const macho &obj, const MachO::any_relocation_info &rel) {
+    return !obj.is64Bit() && (rel.r_word0 & MachO::R_SCATTERED);
+}
+
+bool is_rel_extern(const macho &obj, const MachO::any_relocation_info &rel) {
+    if (obj.isLittleEndian())
+        return (rel.r_word1 >> 27) & 1;
```
ivg: bless me father for I have sinned :)

lib/bap_llvm/llvm_macho_loader.hpp
ivg: OK, for the record, in case of 3.4 this function will not be used

PR #810, opened 2018-03-30
adds ogre dump format

gitoleg:
this PR adds ogre dump format:
```
$ bap /bin/echo --dump=ogre
(declare arch (name str))
(declare base-address (addr int))
...
(arch x86_64)
(base-address 4194304)
...
```

### DIFF
--- a/plugins/print/print_main.ml
+++ b/plugins/print/print_main.ml
@@ -45,6 +45,11 @@ let sec_name memory fn sub =
   | None -> "bap.virtual"
   | Some name -> name
 
+let print_spec ppf proj =
+  match Project.get proj Image.specification with
+  | None -> warning "The file specification is not found"
+  | Some spec ->  Format.fprintf ppf "%a" Ogre.Doc.pp spec
+
 let print_symbols subs secs demangler fmts ppf proj =
   let demangle = create_demangler demangler in
   let symtab = Project.symbols proj in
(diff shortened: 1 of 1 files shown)


plugins/print/print_main.ml
```diff
@@ -45,6 +45,11 @@ let sec_name memory fn sub =
   | None -> "bap.virtual"
   | Some name -> name
 
+let print_spec ppf proj =
+  match Project.get proj Image.specification with
+  | None -> Format.fprintf ppf "Image specification not found\n"
```
ivg: Don't print errors into the standard output. People will be parsing it so everything you're printing here should be in the specified format, otherwise it is a bug on your side, that will make people very angry.Use stderr for errors and diagnostic messages.  In our case, it is hardly an error, maybe not even a warning. So just print nothing, and issue some dignostic in the log, using `info/warning` functions. Its your call.

plugins/print/print_main.ml
```diff
@@ -317,8 +323,9 @@ let main attrs ansi_colors demangle symbol_fmts subs secs =
   Project.add_writer ~ver "bil.adt"
     ~desc:"print BIL instructions in ADT format" pp_disasm_bil_adt;
   Project.add_writer ~ver "bil.sexp"
-    ~desc:"print BIL instructions in Sexp format" pp_disasm_bil_sexp
-
+    ~desc:"print BIL instructions in Sexp format" pp_disasm_bil_sexp;
+  Project.add_writer
+    ~desc:"print image specification" ~ver "ogre" pp_spec
```
ivg: `print the file specification in the OGRE format`

Issue #735, opened 2017-10-27
caching system is confused with the comma separated command line arguments


ivg:
When some comma separated argument is added or changed, then the cached data is invalidated.

ivg:
Closing together with #811

PR #811, opened 2018-04-01
improves bap cache

gitoleg:
This PR is just a first approach for rectifying bap cache.  

Here we define categories of passes that are critical to cache, namely `brancher`, `disassembler`, `lifter`, `loader`, `reconstructor`, `rooter`, `symbolizer`. So passes of categories other then listed are not considered as cache sensitive and `bap` will not take them into account for cache matters.

Examples:
```
$ bap --cache-clean
$ bap exe
$ bap exe --my-pass-arg=42
$ bap exe --x86-lifter=legacy
``` 
First `bap` call with `exe` file will obviously create a cache, since no cache is present.
Second call will use a created cache assuming that `my-pass` is neither `lifter` nor `disassembler` etc.
And third call will not use this cache, since the `x86-lifter` pass arguments changed. 

Also, we take care about coma-separated values, so `bap` calls like
```
$ bap --my-pass-arg=c,d,a
```
and
```
$ bap --my-pass-arg=a,d,c
```
are considered as the same from a cache perspective, so this PR is also fix https://github.com/BinaryAnalysisPlatform/bap/issues/735


lib/bap/bap_project.ml
```diff
@@ -564,11 +564,14 @@ module type S = sig
   type t
   val empty : t
   val of_image : image -> t
+  val service : Bap_service.service
   module Factory : Bap_disasm_source.Factory with type t = t
 end
 
 let register x =
   let module S = (val x : S) in
+  let _provider = Bap_service.Provider.declare
```
ivg: Move it to a plugin, and provide a normal description

lib/bap_disasm/bap_disasm_brancher.ml
```diff
@@ -65,4 +65,15 @@ let dests_of_bil arch =
...
+let service = Bap_service.Service.declare "brancher"
+    ~desc:"A brancher service"
+    ~uuid:"af1f5cce-08fd-438c-8fc3-906f4f3ca07c"
+
+module Factory = struct
+  include Source.Factory.Make(struct type nonrec t = t end)
+
+  let register name source =
+    let desc = sprintf "brancher %s" name in
+    let _provider = Bap_service.Provider.declare ~desc name service in
```
ivg: The unused provider indicates that you're using the system incorrectly. And probably even that you have implemented it wrong

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,158 @@
+open Core_kernel.Std
```
ivg: Remove this module duplication

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,158 @@
...
+      type nonrec t = t [@@deriving bin_io, compare, sexp]
+      let module_name = Some "Bap.Service.Provider"
+      let hash = Hashtbl.hash
+      let version = "0.1"
+      let pp fmt t = Format.fprintf fmt "%s %s" t.name t.desc
+    end)
+end
+
+module Product_t = struct
+  type t = {
```
ivg: It should be a list of(digest, provider)

lib/bap_disasm/bap_disasm_brancher.ml
gitoleg: It’s all about
>  our existing factories will overrider the register function so that it will silently register the provider with the given name (and bogus description).

lib/bap_disasm/bap_disasm_brancher.ml
```diff
@@ -65,4 +65,15 @@ let dests_of_bil arch =
...
-module Factory = Source.Factory.Make(struct type nonrec t = t end)
+let service = Bap_service.Service.declare "brancher"
+    ~desc:"A brancher service"
+    ~uuid:"af1f5cce-08fd-438c-8fc3-906f4f3ca07c"
+
+module Factory = struct
+  include Source.Factory.Make(struct type nonrec t = t end)
+
+  let register name source =
+    let desc = sprintf "brancher %s" name in
```
ivg: should be "no description provided"

lib/bap_disasm/bap_disasm_brancher.ml
ivg: Ah, OK,

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,103 @@
...
+
+  let services = Table.create ()
+
+  let declare ~desc ~uuid name =
+    let t = {desc=desc; data=uuid; name;} in
+    Hashtbl.set services t (Stream.create ());
+    t
+
+  let provide t product =
+    let p = Hashtbl.find_exn services t in
```
ivg: this will fail, since you're not checking that the declared services are unique.
```
let s1 = Service.declare ~desc:"" ~uuid:"" "" 
let s2 = Service.declare ~desc:"" ~uuid:"" "" 
let p = <create-product> 
provide s1 p;; 
```

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,103 @@
...
+    Signal.send (snd p) product
+
+  let request t = fst (Hashtbl.find_exn services t)
```
ivg: will also fails as above.

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,103 @@
...
+  let all_by_name n = all (fun p -> String.equal p.name n)
+  let all_by_service s = all (fun p -> Service.equal p.data s)
+
+  let select ?by_service ?by_name () =
+    let find = Option.value_map ~default:[] in
+    find ~f:all_by_service by_service @
+    find ~f:all_by_name by_name
+
+
+  include Regular.Make(struct
```
ivg: neither provider is a regular type

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,103 @@
...
+  let combine p p' = {
+    digest = String.Set.union p.digest p'.digest;
+    providers = p.providers @ p'.providers;
+  }
+
+  include Regular.Make(struct
+      type nonrec t = t [@@deriving bin_io, compare, sexp]
+      let module_name = Some "Bap.Service.Product"
+      let hash = Hashtbl.hash
+      let version = "0.1"
```
ivg: it should be `1.0`

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,103 @@
...
+  type t = product [@@deriving bin_io, compare, sexp]
+
+  let create ~digest provider =
+    let digest = Set.add String.Set.empty digest in
+    { digest; providers = [provider] }
+
+  let digest t = Set.to_list t.digest |> String.concat
+  let providers t = t.providers
+
+  let combine p p' = {
```
ivg: `Product.(combine x y  <> combine y x)` that violates the monoid requirement (i.e., `combine` should be associative and commutative). (This is, btw, accroding to your own specification of the equality).

My suggestion is to use the string set for the providers fields also.

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,109 @@
...
+  let request = products
+
+  let add_provider serv prov =
+    Hashtbl.set (info serv).providers prov.name prov
+
+  let check_uuid x =
+    if Option.is_none (Uuidm.of_string x) then
+      invalid_arg "Invalid UUID format";
+    if Hashtbl.mem all x then
+      invalid_argf "UUID %s is already in use" x ()
```
ivg: Please provide more information to a user. Suppose you are the user, who has no idea about those services an you get the following error:
```
$ bap --version
Program terminates: UUID 6a0037e5-5bc9-4bfd-a6a6-a495fc4387fa is already in use.
```

By whom, what for, why is it bad. 

The right error message would be:

```
Unable to register the `XXX` service as there is a service with the same UUID.
Please provide a unique service identifier. 
```

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,109 @@
...
+  type t = product [@@deriving bin_io, compare, sexp]
+
+  let digest t = t.digest
+  let provider t = t.provider
+
+  let provide ~digest provider =
+    let sid = Provider.service provider in
+    Signal.send (Service.update sid) {digest; provider}
+
+  include Regular.Make(struct
```
ivg: why do we need Regular here?

lib/bap_service/bap_service.ml
```diff
@@ -0,0 +1,109 @@
...
+    uuid
+end
+
+module Provider = struct
+  type t = provider
+  [@@deriving bin_io, compare, sexp]
+
+  let declare ~desc name service =
+    let p = {name; desc; data=service} in
+    Service.add_provider service p;
```
ivg: why is it necessary to register the provide in the service? Why provider is not just a pair `service * meta`?

lib/bap_service/bap_service.mli
```diff
@@ -0,0 +1,69 @@
...
+type product [@@deriving bin_io, compare, sexp]
+
+type void
+type literal = (void,void,void) format
+
+module Service : sig
+  type t = service
+
+  (** [declare ~desc uuid name] creates a service from description
+      [desc], unique id [uuid] and [name] *)
```
ivg: this documentation says nothing. It basically reiterates the signature. 
It doesn't tell a user that the [uuid] must be statically unique and will be used to distinguish services globally, that `name` is a short and descriptive name, and description provides a short description of the provided service. 
It also doesn't explain what effects this declaration has. 

Again, imagine that you're a user and you're trying to understand how to use this function.

ivg:
I hope that much of this work will migrate to BAP 2.0. I don't think that it is really worthwhile to implement all this for 1.x, it's fine if its caching infrastructure would be a little bit over-conservative. We still would need to use this work when we will update command line and configuration file interfacing in BAP 2.x. However, as it is, I'm keen to close this.

Issue #808, opened 2018-03-29
Term naming is non-injective


ivg:
That means, that we can have two functions (sub terms) having the same name. In practice, this hits with ELF shared libraries, that have both a PLT entry stub and an implementation for a particular stub. Our existing symbolizers, have a tradition to treat the PLT entry stub as a real function, and it works fine with regular binaries, that do no expose and use functions at the same time. But with the shared libraries our symbolizers (objdump and IDA) tells us the same name for two different addresses. Careful thinking tells us, that we can't actually fix this on the symbolizer level. Neither we actually do want to fix symbolizers. We should enforce, however, an invariant that for each name we have at most one address/term. Or speaking mathematically, that our naming function is injective. 

## The desired properties

We should enforce an invariant, that for each name we have at most one address, while several names can share the same address. That means that the `address : N -> A` function is surjective, where `N` is the set of names and `A` is the set of addresses. And `name : A -> N` is injective. This pair of functions also induces a partitioning of names, `principal : N -> N = name . address`, i.e., it establishes a principal name for each name, that is the representative for the class of names of the given name. We use the principal name as the name of the symbol and treat the rest of the names in the class as aliases. 

## Our current implementation

1. We use symbolizers, that provide the `name` function. We do not check that this function is injective, i.e., that for each name we have at most one address.

2. On the IR level, we allow arbitrary renaming. We do not control how terms are associated with addresses and vice verse. We apply renaming during different phases of reconstruction and disassembling, and in general every pass can rename a term. 

## Implementation Limitations

1. We cannot enforce injectivity on the `name` function, so we must assume that this function is neither injective nor surjective. 

2. We use ugly `Term.set_name` function that uses internal hashtable to establish a mapping between terms and names. 

## Proposed Solution

1. We will enforce the naming invariant only on the IR level. Lower levels, such as symtab, disassembly, etc represent what we had as input, so it is better to leave them as they are. Thus we will enforce the invariant on the domain of terms, not the set of addresses. But given that `term : A -> T` is injective, our invariant is even more strict, than the desired one. 

2. We will not try very hard to enforce the invariant on the `Term.set_name` function. However, it will work correctly, unless it was abused. We shall consider removing this function in BAP 2.0, and we should mark it as obsolete asap, to refrain users from touching it. We shall also dedocument it. 

3. To choose the principal name, we will treat the last proposed name as the more important one, assuming that every consequent analysis pass has more information than previous. In other words, the last set name, is the principal name. 

[...]

PR #813, opened 2018-04-04
adds mangling of duplicated subroutines

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/808

We track every addition of a subroutine to a program by storing names and tids in a table. Every addition replaces previous entries with the same name. When someone requests a result, it's easy to tell if a subroutine name needs in demangling: if a `tid` in the table for this name is differ from a subroutine's one, then other subroutine started to use this name, and we need to demangle current name. The same is true for subroutine aliases.

### DIFF
--- a/lib/bap_types/bap_ir.ml
+++ b/lib/bap_types/bap_ir.ml
@@ -1396,14 +1396,8 @@ module Ir_program = struct
     Option.value ~default:[]
       (Term.get_attr s Ir_sub.aliases)
 
-  let same_names s s' =
-    String.equal (Ir_sub.name s) (Ir_sub.name s')
-
-  let strings_mem xs s = List.mem xs s ~equal:String.equal
-
-  let need_to_mangle s reserved =
-    let names = Ir_sub.name s :: sub_aliases s in
-    List.exists names ~f:(fun n -> strings_mem reserved n)
+  let sub_addr s =
+    Term.get_attr s Bap_attributes.address
 
   let mangle_name addr tid name =
     let mname = match addr with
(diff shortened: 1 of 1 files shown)


lib/bap_types/bap_ir.ml
```diff
@@ -1389,15 +1392,59 @@ module Ir_program = struct
...
+    let subs' = Vec.foldi subs ~init:[] ~f:(fun i acc sub ->
+        if not (need_to_mangle sub names) then acc
+        else
+          let may_mangle = may_mangle (sub_addr sub) (Term.tid sub) in
+          let als = List.map (sub_aliases sub) ~f:may_mangle in
+          let nam = may_mangle (Ir_sub.name sub) in
+          let sub = Term.set_attr sub Ir_sub.aliases als in
+          let sub = Ir_sub.with_name sub nam in
+          (i, sub) :: acc) in
+    List.iter ~f:(fun (i, s) -> Vec.set subs i s) subs'
```
ivg: I really don't like using vector here... it is abusing.

lib/bap_types/bap_ir.ml
```diff
@@ -1389,15 +1392,59 @@ module Ir_program = struct
...
+
   module Builder = struct
-    type t = tid option * sub term vector
+    type t = tid option * tid String.Table.t * sub term vector
```
ivg: You shall not do this on the builder interface.

ivg:
You shall enforce this invariant on the program data structure, so every time a value of type program is created you need to check it (non-invasive) and if the invariant holds, then return the program unchanged otherwise rename the offending subroutines. 

I would suggest you to make the `program` type private to ensure that it is not created anywhere without a check (that means that you need to hide it inside a module). Of course, this should all be done in the implementation, nothing should be changed in the interface.

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,60 @@ type sub = {
...
+
+  let fix_names subs =
+    let names = String.Table.create () in
+    Array.iter subs ~f:(fun s ->
+        let name = s.self.name in
+        let tid = s.tid in
+        Hashtbl.change names name ~f:(function
+            | None -> Some tid
+            | Some tid' ->
+              if Tid.(tid' < tid) then Some tid
```
ivg: So this policy prioritizes a subroutine with a smaller tid, though it doesn't give the priority to the last chosen name.

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,90 @@ type sub = {
...
+  let names = String.Table.create () in
+  Array.iter subs ~f:(fun s ->
+      Hashtbl.change names s.self.name ~f:(function
+          | None -> Some [s.tid]
+          | Some tids -> Some (s.tid :: tids)));
+  names
+
+let fix_names ?(old_names=String.Table.create ()) subs =
+  let names = names_of_subs subs in
+  let conflicts = Hashtbl.count names ~f:(fun x -> List.length x > 1) in
```
ivg: `fun x -> List.length x > 1` is in general very inefficient pattern. You do not need to know the full length of the list, to check that it is greater than 1, just do `match xs with | [] | [_] -> true | _ -> false` and you will go from linear to O(1) algorithm.

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,90 @@ type sub = {
...
+  if conflicts = 0 then subs
+  else
+    let remove name tid = match Hashtbl.find names name with
```
ivg: it is a bad style to put nested function in the else branch. Are those function has to defined only if the precondition holds? Please, move them out the function and define at the top-level. If you're concerned with a namespace pollution, then put them in a module.

ivg: instead of using `find`/`set` you can use `change`. Though, this code already is too bad to be discussed at all...

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,90 @@ type sub = {
...
+      match Hashtbl.find old_names s.self.name with
+      | Some old_tid -> Tid.equal s.tid old_tid
+      | _ -> false in
+    Array.map ~f:(fun sub ->
+        match Hashtbl.find_exn names sub.self.name with
+        | [] | [_]-> sub
+        | tids ->
+          if is_old sub then mangle sub
+          else
+            let max_tid = Option.value_exn
```
ivg: There should not be any dependence on the tid ordering. The order relation between two tids has nothing to do with a corresponding term freshness or anything. We may switch later to uuid as tids and assign them randomly. So please, remove any dependencies.

gitoleg: let me quote you: 
> If both names are new, then pick the rename a name that corresponds to a tid that is smaller.

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,90 @@ type sub = {
...
+  let addr = Dict.find s.dict Bap_attributes.address in
+  let name = mangle_name addr s.tid s.self.name in
+  Tid.set_name s.tid name;
+  let self = {s.self with name} in
+  {s with self}
+
+let names_of_subs subs =
+  let names = String.Table.create () in
+  Array.iter subs ~f:(fun s ->
+      Hashtbl.change names s.self.name ~f:(function
```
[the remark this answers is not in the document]
gitoleg: which signature is  `add_multi : ('a, 'b list) t -> key:'a -> data:'b -> unit`
So a type of a hash table is not what I need here

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,90 @@ type sub = {
...
+  | None -> sprintf "%s%%%s" name (Tid.to_string tid)
+
+let mangle_sub s =
+  let addr = Dict.find s.dict Bap_attributes.address in
+  let name = mangle_name addr s.tid s.self.name in
+  Tid.set_name s.tid name;
+  let self = {s.self with name} in
+  {s with self}
+
+let names_of_subs subs =
```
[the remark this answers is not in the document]
gitoleg: Justification is in elements order in an array of subroutines. It I would use map, I 
have to do something like `Array.rev_in_place @@ Array.fold `, but definitely not
`Array.map`

lib/bap_types/bap_ir.ml
```diff
@@ -143,18 +143,67 @@ type sub = {
...
+      | None -> true in
+    let keep_name tids name tid = Map.add tids ~key:name ~data:tid in
+    let tids = Array.fold news ~init:String.Map.empty ~f:(fun tids sub ->
+        match Map.find tids sub.self.name with
+        | None -> keep_name tids sub.self.name sub.tid
+        | Some _ ->
+          if is_new sub.tid sub.self.name then
+            keep_name tids sub.self.name sub.tid
+          else tids) in
+    Array.map news ~f:(fun sub ->
```
ivg: `Array.map` will create a new array and copy the contents of the old array even if the `f` function is `ident`:
```
let xs = Array.init 8 ~f:ident;;
val xs : int Core_kernel.Array.t = [|0; 1; 2; 3; 4; 5; 6; 7|]
# let ys = Array.map xs ~f:ident;;
val ys : int Core_kernel.Array.t = [|0; 1; 2; 3; 4; 5; 6; 7|]
# phys_equal xs ys;;
- : bool = false
```

In our case, it will result in double copying of the whole program even if the change was benign and didn't break the naming injectivity. 

So you need to prefix the map function with the check `if Array.length news <> Map.length tids then Array.map ... else news`

PR #820, opened 2018-04-09
OCaml 4.06 (-safe-string) + Jane Street 0.11.x library compatibility + minor fixes

drvink:
# Do not merge, needs review!

This depends on other changes:

* `FrontC`: `safe-string` compatibility (needs vendoring)
* `optcomp`: `safe-string` compatibility (needs vendoring unless diml/optcomp#4 is merged)
* `piqi`: `safe-string` compatibility (works if pinned to dev repo)
* `piqilib`: `safe-string` compatibility (works if pinned to dev repo)
* `ppxlib`: restoration of `ocamlbuild` plugin (needs vendoring)


lib/bap_primus/bap_primus_lisp_program.ml
```diff
@@ -150,8 +150,8 @@ let pp_term pp_exp ppf = function
 let pp_word = pp_term Int64.pp
 let pp_var = pp_term String.pp
 
-let rec concat_prog =
-  List.concat_map ~f:(function
+let rec concat_prog p =
```
ivg: yet another definition rejected by https://github.com/ocaml/ocaml/pull/556 PR.

lib/bap_primus/bap_primus_lisp_program.ml
```diff
@@ -359,7 +359,7 @@ module Reindex = struct
         rename v >>= fun v ->
         map x >>| fun x ->
         {t with data = Set (v,x)}
-    and map_all = State.List.map ~f:map in
+    and map_all xs = State.List.map xs ~f:map in
```
ivg: rejected since ocaml/ocaml#556

lib/bap_primus/bap_primus_sexp.ml
```diff
@@ -89,4 +89,4 @@ let rec sexp_of_exp = function
       sexp_of_exp x;
       sexp_of_exp y;
     ]
-and sexps_of_exps = List.map ~f:sexp_of_exp
+and sexps_of_exps exps = List.map exps ~f:sexp_of_exp
```
ivg: rejected since ocaml/ocaml#556

lib/bap_traces/bap_trace_meta.ml
```diff
@@ -41,7 +41,7 @@ module Binary = struct
        @\nmd5sum: %s\
        @\n@[<2>\
        args:   %a@]\
-       @\n@[<2>envp:@\n%a@]@]@\n}" t.path (Digest.to_hex t.md5sum) Args.pp t.args Envp.pp t.envp
+       @\n@[<2>envp:@\n%a@]@]@\n}" t.path t.md5sum Args.pp t.args Envp.pp t.envp
```
ivg: This will print digest as a string of bytes instead of a nice and common to us hex representation of the MD5sum

lib/bap_types/bap_helpers.ml
```diff
@@ -545,7 +545,7 @@ module Simpl = struct
     and while_  c ss = match exp c with
       | Int x when Word.is_zero x -> []
       | c -> [While (c, bil ss)]
-    and bil = List.concat_map ~f:stmt in
+    and bil stmts = List.concat_map stmts ~f:stmt in
```
ivg: yet another ocaml/ocaml#556 rejection

lib/monads/monads.mli
```diff
@@ -726,9 +726,9 @@ module Std : sig
             evaluation is unspecified.  *)
         val all : ('a,'e) m t -> ('a t, 'e) m
 
-        (** [all_ignore cs] performs all computations in [cs] in an
+        (** [all_unit cs] performs all computations in [cs] in an
             unspecified order, and discards the results. *)
-        val all_ignore : ('a,'e) m t -> (unit,'e) m
+        val all_unit : ('a,'e) m t -> (unit,'e) m
```
ivg: we need to preserve the old interface, otherwise we need to release Monads 2.0. 
I don't like the idea of bumping the major version just because of changes in Core.

lib/regular/regular_cache.ml
```diff
@@ -35,7 +35,7 @@ end
 module Digest = struct
   include String
 
-  let make s = s |> Digest.string |> Digest.to_hex
+  let make s = s |> Md5.digest_string |> Md5.to_hex
```
ivg: Probably it is easier just to rely on `Caml.Digest`

lib/bap_build/bap_build.ml
```diff
@@ -15,7 +15,7 @@ module Plugin_rules = struct
 
   let default_predicates = [
     "custom_ppx";
-    "ppx_driver";
+    "ppxlib";
```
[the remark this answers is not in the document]
drvink: More or less, yes. As of `ppx_driver.v0.11.0`:
```
arch-vm[1124]:~% opam show ppx_driver.v0.11.0

=-=- ppx_driver: information on all versions =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
name                   ppx_driver
all-installed-versions v0.10.4 [4.05.0+flambda]
all-versions           113.09.00  113.24.00  113.33.00  113.33.00+4.03  113.33.01+4.03  113.33.02+4.03  113.33.03  113.33.04  v0.9.0
                       v0.9.1  v0.9.2  v0.10.0  v0.10.1  v0.10.2  v0.10.3  v0.10.4  v0.11.0  v0.11.117.00+101

=-=- Version-specific details -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
version       v0.11.0
*...*
depends:      "ocaml" {>= "4.04.1"}
              "jbuilder" {build & >= "1.0+beta18.1"}
              "ppxlib" {>= "0.1.0"}
synopsis      Deprecated: use ppxlib instead
```

.merlin
```diff
@@ -1,11 +1,14 @@
 FLG -short-paths
 FLG -w +a-4-6-7-9-29-32..42-44-45-48-50-60
+FLG -safe-string
 
+PKG core
```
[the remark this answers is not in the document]
drvink: This might be a problem: I'll have to double check, but if I recall, some things have been moved to Core. :(

lib/bap_primus/bap_primus_lisp_program.ml
drvink: Are you sure you don't mean the opposite? OCaml 4.06 _requires_ eta expansion here--you ought to get an error without, explicitly rejecting the LHS without.

ivg: Yeah, the new let-rec checker rejects this, and although I'm fine with it, I've submitted an issue and notified upstream, as I have an impression from the PR comments that is not expected. But in any case, eta-expansion here is fine.

lib/bap_traces/bap_trace_meta.ml
drvink: Are you sure? `t.md5sum` is already a string. (The value `t` here is not of type `Md5_lib.t`.)

drvink:
I've addressed the Core portability issue; now we are back to only needing Core_kernel. However, the fix for this ends up bringing in `Unix.map_file`, which was introduced in OCaml 4.06. At this point we would need our own `mmap`/`MapViewOfFile` support in order to keep compatibility with older versions of OCaml, but realistically, it seems a lot easier to just bring the minimum version up to 4.06.

Since we'd be requiring OCaml 4.06, it would then probably make sense to just merge the `safe-string` related changes as well, even if it means we have to break the API (because at this point we're no longer being forced to change anything due to JSC, but rather OCaml itself, and people have had since 4.02.0--nearly four years--to adapt code to use `bytes` in preparation for the change to `safe-string` as the default mode.

ivg:
> However, the fix for this ends up bringing in Unix.map_file, which was introduced in OCaml 4.06. At this point we would need our own mmap/MapViewOfFile support in order to keep compatibility with older versions of OCaml, but realistically, it seems a lot easier to just bring the minimum version up to 4.06.

The `Unix.map_file` function was just moved from the `Bigarray` module (e.g., `Bigarray.Array1.map_file`) and it is in both places in 4.06, though marked as deprecated in Bigarray (and in the Bigarray it was for the last 30 years). I think that we should stick to the Bigarray version and pray that their deprecation process is long enough.

ivg:
> Since we'd be requiring OCaml 4.06

I was hoping for 4.04 and above. If core works from 4.04 and above, then I believe we can also be as compatible as they are.

lib/bap_traces/bap_trace_meta.ml
ivg: Yep, it is a `string`, but not a string of ASCII characters, but a string of bytes. The `string` type is not constrained only to printable characters, the only difference between strings and bytes is that the former is non-mutable. So in the trace `md5sum` is stored as 32 bytes of data, not in a textual and printable representation.

drvink:
Sorry for the wait. The branch now contains a JSC 0.11.x-compatible version that works with `-unsafe-string` (i.e. the `-safe-string` changes are gone--I still have them locally; I assume you'll want them eventually, so maybe we can make a separate PR for them and just leave it sitting around until you decide to merge it).

How should we handle vendoring `ppxlib`? As I said previously, the changes are trivial (we just re-add the plugin for `ocamlbuild` that the JSC folks removed), but it'll need a separate PR.

drvink:
@ivg ping :)

ivg:
@drvink, thanks, I actually missed the last update. I'm on it right now. So far, LGTM... but this `Map.set` drives me nuts<sup>1</sup>. 

> How should we handle vendoring ppxlib? As I said previously, the changes are trivial (we just re-add the plugin for ocamlbuild that the JSC folks removed), but it'll need a separate PR.

Yep, moreover it will need a separate repository and a separate package, as it will have a different license (Apache perhaps) and we can't pollute our license namespace. So, ideally, it should be a separate opam package, on which we should depend. 

So this is our roadmap:

- [ ] restore and package the ppxlib plugin
- [ ] fix the opam dependencies and pass through Travis
- [ ] update our opam-repository (@gitoleg)
- [ ] do something to reduce the impact of changes, or think about it (@ivg)
- [ ] update bap-plugins (@ivg)

-----------
<sup>1</sup>) Not only ugly it is, it will also have a severe impact on our users, as they would be forced to rewrite their code as well, so I'm considering different options to smooth the transition, e.g., providing a compatibility module that will support both `Map.add` and `Map.set` with the corresponding deprecation warning, provide a <strike>perl</strike> script that will do the fix, or, at least, provide guidelines. I will also make a few announcements beforehand in  gitter and discord. In fact, this is probably the biggest breaking change since the history of modern BAP and our dependency on core's interfaces sort of defeats the whole purpose of API stability. By the way, is `open Core_kernel.Std` not working anymore in v0.11.x?

drvink:
OK. Where should I make the PR for the vendored ppxlib?

I agree that the `Map.set` is ugly. It's necessary because they changed the API of `Map.add` to indicate whether a value was added to the map or if it was a duplicate. We could use `Map.add_exn` instead, but I figured that was worse than just burying the would-be exception by using `Map.set`. Your suggestion of a compatibility module with deprecation warnings sounds like a good one. By the way, where is the Discord? :)

`Core_kernel.Std` still works in 0.11.x, but you get lots of deprecation warnings, so I figured we may as well change it now.

lib/bap_dwarf/dwarf_fbi.mli
```diff
@@ -16,10 +16,10 @@ type fn [@@deriving bin_io, compare, sexp]
     - .debug_info   [Section.Info]
     - .debug_str    [Section.Str]
 *)
-val create : string Dwarf_data.t -> t Or_error.t
+val create : bytes Dwarf_data.t -> t Or_error.t
```
ivg: let's keep `string`s in the interface. Especially, since we're not going to mutate them, it is better to use the `string` type to denote this.

lib/bap_traces/bap_trace_meta.ml
```diff
@@ -41,7 +51,8 @@ module Binary = struct
        @\nmd5sum: %s\
        @\n@[<2>\
        args:   %a@]\
-       @\n@[<2>envp:@\n%a@]@]@\n}" t.path (Digest.to_hex t.md5sum) Args.pp t.args Envp.pp t.envp
+       @\n@[<2>envp:@\n%a@]@]@\n}"
+      t.path (hexstring_of_bytestring t.md5sum) Args.pp t.args Envp.pp t.envp
```
ivg: I'm wondering, why we can't use `Digest.to_hex` here? It is still in 4.06, and still has type `string -> string`.

lib/regular/regular.mli
```diff
@@ -580,7 +580,7 @@ module Std : sig
         ?dump  : (Out_channel.t -> 'a -> unit) ->
         ?pp    : (Format.formatter -> 'a -> unit) ->
         ?size  : ('a -> int) ->
-        ?blit_to_string:('a,string) copy ->
+        ?blit_to_bytes:('a,bytes) copy ->
```
ivg: we need to keep `blit_to_string` for compatibility, though adding `blit_to_bytes` is also a good idea.

lib/regular/regular.mli
```diff
@@ -605,11 +605,11 @@ module Std : sig
...
       val to_bigstring : 'a t -> 'a -> bigstring
 
-      (** [blit_to_string writeable buf value pos] copies a serialized
-          representation of the [value] into existing string [buf],
+      (** [blit_to_bytes writeable buf value pos] copies a serialized
+          representation of the [value] into existing bytes [buf],
           starting from the given position. It is undefined behavior,
           if the [value] doesn't fit into the string [buf] *)
-      val blit_to_string : 'a t -> string -> 'a -> int -> unit
+      val blit_to_bytes : 'a t -> bytes -> 'a -> int -> unit
```
ivg: can you please provide the `blit_to_string` function, for compatibility.

oasis/common
```diff
@@ -1,7 +1,7 @@
 OASISFormat: 0.4
 Name:        bap
 Version:     1.5.0-dev
-OCamlVersion: >= 4.03
+OCamlVersion: >= 4.06
```
ivg: why not 4.04.1 (as in core_kernel)?

plugins/map_terms/map_terms_main.ml
```diff
@@ -68,8 +68,12 @@ module Scheme = struct
...
     | _ -> Error {|expect "("<preds> <marks>")"|}
 
-  let sexp_error {Sexp.location; err_msg} =
-    Error (sprintf "Syntax error: %s - %s" location err_msg)
+  let sexp_error (err : Sexp.parse_error) =
+    let line, col = match err.parse_state with
+      | `Annot {parse_pos} -> parse_pos.text_line, parse_pos.text_char
+      | `Sexp {parse_pos} -> parse_pos.text_line, parse_pos.text_char
+    in
+    Error (sprintf "Syntax error: line %d, column %d - %s" line col err.err_msg)
```
ivg: :+1:

ivg:
> OK. Where should I make the PR for the vendored ppxlib?

To `ocaml/opam-repository`. You need to create a separate repository, first (you can then transfer the ownership to BinaryAnalysisPlatform if you want to). Then to package it as a separate opam package and submit it to opam-repository, once it got merged we can rely on it. 


> I agree that the Map.set is ugly. It's necessary because they changed the API of Map.add to indicate whether a value was added to the map or if it was a duplicate. We could use Map.add_exn instead, but I figured that was worse than just burying the would-be exception by using Map.set.

We can't use `Map.add_exn` as most of the uses of `Map.add` in our code base assume that the key may exist. So it will definitely raise. 

> By the way, where is the Discord? :)

https://discord.gg/bwJ3p7q

> Core_kernel.Std still works in 0.11.x, but you get lots of deprecation warnings, so I figured we may as well change it now.

great!

ivg:
> By the way, since optcomp is no longer maintained (see diml/optcomp#4), we might want to consider switching to ppx_optcomp or cppo. This was a -safe-string-related issue, so if people want to build bap on 4.06.x, they will need to pin and fix it themselves unless we either vendor it or use something else.

Honestly, I don't know why we have optcomp in the dependencies. Most likely it was a missed dependency somewhere upstream. What I'm sure is that we're not using it anywhere in bap. So we can consider dropping this dep at all (assuming that upstream has fixed it) or indeed switching to something more modern,

drvink:
> Honestly, I don't know why we have optcomp in the dependencies. Most likely it was a missed dependency somewhere upstream. What I'm sure is that we're not using it anywhere in bap. So we can consider dropping this dep at all (assuming that upstream has fixed it) or indeed switching to something more modern,

OK. I'll remove it and see if I can still build bap from a clean switch.

lib/bap_traces/bap_trace_meta.ml
drvink: Ah, I understand your comment now. `Caml.Digest.to_hex` will be fine, but `Core.Digest.to_hex` will not:

```
utop[3]# #show Core.Digest.to_hex;;
val to_hex : Core.Md5.t -> string
```

XVilka:
@drvink any updates on this?

drvink:
@XVilka you can try pulling from this PR and give it a try; it would definitely be nice to get feedback from users that everything works. Make sure to `opam pin -k git add ppxlib_ocamlbuild https://github.com/BinaryAnalysisPlatform/ppxlib_ocamlbuild.git` first (this will be submitted to opam before this PR is merged so that people won't need to do this), and then you should be able to build bap from this branch. Note that these changes bring the minimum OCaml version requirement to 4.04.1.

drvink:
Also, if you're on 4.06.x, note that `-unsafe-string` is being used for now in order to keep bap's API stable. I have a (working) bap branch for `-safe-string`, but since it substantially changes the API and also requires pinning a bunch of repos on which bap depends (see first post in this thread), it's not really suitable for normal use--also, I haven't optimized it at all, so there's probably significant unnecessary copying between `string` and `bytes` at the moment.

ivg:
I'm planning to merge this before 1.5 release. @gitoleg, can you please take care of this?

@drvink, if you have any new updates to this (due to the movement of the OCaml infrastructure) can you please push or at least show us them?

drvink:
@ivg Sorry for the delay in replying. I'll push a version rebased against bap master in a moment.

As you previously mentioned, we can use [ppx_driver_ocamlbuild](https://github.com/janestreet-deprecated/ppx_driver_ocamlbuild) now--it's basically the same as the [ppxlib_ocamlbuild](https://github.com/BinaryAnalysisPlatform/ppxlib_ocamlbuild) package I vendored. However, it doesn't seem to be in opam--how should we go about adding the dependency (i.e. since we can't just list it as an opam dependency, how shall we make the build work automatically)?

ivg:
So, previously this code had a restricting license which was preventing us from including (via plain copy-pasting) the code into our source code repository (i.e., in the bap repo directly), which is the simplest and clearest way to get ppx_driver back. Since now they had released it under a permissive, and more over the same, license, we finally can just pick that file, and store it in the bap repository and use it. 

TL;DR; just copy the code directly into the source code of bap (that is what vendoring is).

XVilka:
So I guess this one will not make it until 1.5.0, right?

ivg:
Until 1.6, yes. It will go into the next dev version.

drvink:
Sorry; I've just not had the time to try to integrate the ocamlbuild stuff, though I've been keeping the branch up to date against master so that it doesn't bitrot.

XVilka:
So it is time to merge now, as 1.5.0 was released.

drvink:
I think this is ready to go now thanks to @gitoleg. Sorry for not having looked into the ocamlbuild issue earlier (had I spent the moment to even check, this would have been ready for 1.5.0, since it seems the ppx stuff wasn't necessary at all anymore!)

drvink:
Also FYI, I've been maintaining a safe-string branch in parallel to this one, in `real-safe-string` under [drvink/bap](https://github.com/drvink/bap). This was up to date with master prior to gitoleg's changes; I'll bring it into sync again. Note that I haven't tried to optimize it to avoid needless copying between `string` and `bytes`.

lib/regular/regular.mli
```diff
@@ -817,17 +823,27 @@ module Std : sig
...
     (** bytes  *)
     type t = Bytes.t [@@deriving bin_io, compare, sexp]
 
     include Container.S0   with type t := t with type elt := char
     include Blit.S         with type t := t
     include Identifiable.S with type t := t
-    module To_string   : Blit.S_distinct with type src := t with type dst := string
     module From_string : Blit.S_distinct with type src := string with type dst := t
 
+    module To_string  : sig
```
ivg: I would suggest the following way to deprecated three functions out of this module:
```ocaml
module To_string : sig
    include Blit.S_distinct with type src := string and type dst := t
    val blit : (t, t) Blit.blit
    [@@deprecated "[since 2018-11] use the Bytes.blit instead"]

    (* etc *)
```

ivg:
LGTM as soon as we will fix all the external dependencies. Correct me if I'm wrong, but the only one left is FrontC? I have made a courtesy notification to the author that we're taking his package from him, no reply so far. But I think that he will be OK.

drvink:
FrontC is the only thing I have a local pin for that hasn't already been made part of this PR, so I think so.

XVilka:
Unrelated to FrontC vendoring, but seems there is also another option - https://github.com/jhjourdan/C11parser

XVilka:
Seems something wrong with CI? It still fails with error that supposed to be fixed already.

XVilka:
It seems that you need to upgrade also piqi & piqilib to the latest versions that support OCaml 4.06/4.07:
 - [piqi-v0.6.14](https://github.com/alavrik/piqi/releases/tag/v0.6.14)
 - [piqi-ocaml-v0.7.6](https://github.com/alavrik/piqi-ocaml/releases/tag/v0.7.6)

PR #821, opened 2018-04-10
Add an interface for binnable interval trees

DIJamner:
Add an interface and `Make_binable` functor for binable interval trees. This allows interval trees to be serialized and used as universal values (e.g. in term attributes).

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -953,9 +953,8 @@ module Std : sig
 
     (**  Binable Abstract Interval.
 
-         An extension of the Interval signature that supports the
-         necessary extensions to be serializable as a
-         {{!Value}universal value}.
+         An extension of the Interval signature with the
+         Binable interface.
     *)
     module type Interval_binable = sig
       type t [@@deriving bin_io, compare, sexp]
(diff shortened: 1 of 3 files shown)


lib/bap_types/bap_interval_tree.ml
```diff
@@ -315,3 +315,92 @@ module Make(Interval : Interval) = struct
...
+    key : key;
+    data : 'a;
+    height : int;
+    greatest : point;
+    least : point;
+  } [@@deriving fields, sexp, compare, bin_io]
+
+  type +'a t = 'a node option [@@deriving sexp, compare, bin_io]
+
+  let height = Base.height
```
ivg: it's ok to just include the `Base` module, though I'm fine with an explicit option.

lib/bap/bap.mli
```diff
@@ -950,6 +950,38 @@ module Std : sig
     module Make(Interval : Interval) : S
       with type key := Interval.t
        and type point := Interval.point
+
+    (**  Binable Abstract Interval.
+
+         An extension of the Interval signature that supports the
+         necessary extensions to be serializable as a
+         {{!Value}universal value}.
```
ivg: Well, in fact it has nothing to do with Values, so the comment is rather confusing. Can you fix it to "An extension of the Interval signature with the Binable interface"

lib/bap/bap.mli
```diff
@@ -950,6 +950,38 @@ module Std : sig
...
+        {{!Value}universal value}.
+    *)
+    module type S_binable = sig
+      type 'a t [@@deriving bin_io, compare, sexp]
+      include S with type 'a t := 'a t
+    end
+
+    (** [Make_binable(Interval)] create an abstract interval tree data type
+        that uses abstract [Interval] and can be serialized as a
+        {{!Value}universal value}.
```
ivg: the same, no values are involved.

ivg:
thanks!

PR #823, opened 2018-04-12
Add an `eval_cond` observation to allow observing non-taken conditionals

ethan42:
The `jumping` observation allows for observing the taken conditionals
as well as the jump destination. This extra observation allows
observing all conditions that were evaluated up until the taken branch
(including the false ones). Conditions in branches _after_ the taken
branch will not be observed.


ethan42:
@ivg: I think I did it, not sure if there is anything else that should be updated for the reflection.

PR #824, opened 2018-04-12
Rename Solution init value to default. Add Solution.derive

DIJamner:
Change the naming convention from `init` to `default` for default values in `Solution.t`s and expose this via a `Solution.default` getter function. Implement a `Solution.derive` function that allows for creating new `Solution.t`s from old ones.

### DIFF
--- a/lib/graphlib/graphlib_graph.ml
+++ b/lib/graphlib/graphlib_graph.ml
@@ -1094,27 +1094,38 @@ module Fixpoint = struct
   type ('n,'d) t = Solution : {
       steps : int option;
       iters : int;
-      init : 'd;
+      default : 'd;
       approx : ('n,'d,_) Map.t;
     } -> ('n,'d) t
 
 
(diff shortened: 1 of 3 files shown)


lib/graphlib/graphlib_graph.ml
```diff
@@ -1094,28 +1094,39 @@ module Fixpoint = struct
...
+    | None -> default
     | Some x -> x
 
   let is_fixpoint (Solution {steps; iters}) = match steps with
     | None -> iters > 0
     | Some steps -> iters < steps
 
+  let derive (Solution s) ~f default =
+    let f ~key ~data = f key data in
+    Solution {
```
ivg: We need to reset the step and iter counters, since this is a fresh solution, though based on an existing one. You may reuse the create function.

DIJamner: Got it. I wasn't sure which way it should go.

PR #829, opened 2018-04-26
splits Bap_helpers module 

gitoleg:
It becomes hard and inconvenient to support/edit/read `bap_helpers` module because it is about 1100+ lines length. 
So this PR splits `bap_helpers` module into few of less size:  `bap_eff`, `bap_types_infer`, `bap_normalize`, `bap_apply`, `bap_simpl` and `bap_helpers` itself


ivg:
by the in-person discussion, we decided to close this PR so far.

PR #830, opened 2018-04-26
adds const propagation

gitoleg:
This PR adds a const propagation, as well as some tests for it.


lib_test/bap_types/test_const_prop.ml
```diff
@@ -0,0 +1,154 @@
+open Core_kernel.Std
```
ivg: Your test only verifies that the propagation, i.e., substitution of a value occurs, it doesn't verify the most important property - that the semantics of the program is preserved. 

Your tests must also verify that your transformation is sound and complete, i.e., that for all programs `p`, `p'` such that `p' = f(p)` the following is true:

soundness: `P(p) -> P(p')`
completeness: `P(p') -> P(p)`

for all  predicates  `P` from a set of predicates that defines the semantics of BIL.

lib/bap_types/bap_const_propagation.ml
```diff
@@ -0,0 +1,93 @@
...
+
+(** [remove_forward_defs bil defs]
+    for every [x := exp] in [bil], where [exp] depends on some
+    [y1 .. yN] redefined later then [x],
+    removes both [x] and [y1 .. yN] from defs *)
+let remove_forward_defs bil defs =
+  (object
+    inherit [exp Var.Map.t] bil_visitor
+    method! enter_move var e defs =
+      let redefined = Set.filter (Exp_helpers.free_vars e)
```
ivg: It is likely that you are misunderstanding `free_vars`. For example, `free_vars (y := x + let x = z in x+ 1)` will yield `x,z`.

Also, why is a set of variables that are redefined by a move statement is a subset of a free variables of that statement s.t. they are not assigned after in any following statement? Either it is a wrong name or a wrong implementation.

lib/bap_types/bap_const_propagation.ml
```diff
@@ -0,0 +1,93 @@
...
+          ~f:(fun v -> is_assigned v succs) in
+      if Set.is_empty redefined then defs
+      else
+        Set.fold ~init:(Map.remove defs var) redefined
+          ~f:(fun defs v ->
+              if is_assigned v succs then Map.remove defs v
```
ivg: `redefined`  is defined above as a set of variables s.t. that for each `v` in this set `is_assigned v succs` is true. Thus this expression will be always true. So, what you're trying to say is that `Set.fold ~init:(Map.remove defs var) redefined ~f:Map.remove`. On in other words, it is `defs / v U redefined`.

I can understand this as a claim, that a statement `v := x` kills `c U {is_assigned x : FV(x)}` from the set of constants? 
This doesn't make any sense at all. This property has nothing to do with the FV property.

ivg:
Please, update this branch, split it if necessary and provide a thorough description of used algorithms, chosen approaches, and affected interfaces. The due date is before Friday so that I can start the reviewing process ASAP.

plugins/bil_pipe/bil_const_propagation.ml
```diff
@@ -0,0 +1,236 @@
...
+    method eval (bil : bil) : bil r =
+      SM.List.fold bil ~init:[] ~f:(fun acc s ->
+          self#eval_stmt s >>= fun s ->
+          SM.return (s :: acc)) >>= fun bil ->
+      SM.return (List.rev bil)
+
+    method eval_move v e : stmt r =
+      match e with
+      | Int w ->
+        self#update (Def w) v >>= fun () ->
```
ivg: this is correct only if this move dominates all possible uses of this variable.

plugins/bil_pipe/bil_const_propagation.ml
```diff
@@ -0,0 +1,236 @@
...
+
+    method eval_jmp e : stmt r =
+      self#eval_exp e >>= fun e ->
+      SM.return (Bil.jmp e)
+  end
+end
+
+module Dead_code = struct
+
+  let is_used var bil =
```
ivg: it is called free variables, and we already have `free_vars` for BIL.

plugins/bil_pipe/bil_const_propagation.ml
```diff
@@ -0,0 +1,236 @@
...
+          else used_in bil
+        | While (_, body) -> used_in body || used_in bil
+        | Move (var',_) when defined var' -> Some false
+        | s when is_in_free s -> Some true
+        | _ -> used_in bil in
+    match used_in bil with
+    | None -> false
+    | Some x -> x
+
+  let eliminate bil =
```
ivg: This is an extremely suboptimal implementation, for each statement you look all whether its variable is used in all consecutive statements, which is O(n^2). You can implement this in O(n), if you will first translate each stmt list, into `(stmt * live) list`.

plugins/bil_pipe/bil_const_propagation.ml
```diff
@@ -0,0 +1,236 @@
...
+    let rec loop acc scope = function
+      | [] -> List.rev acc
+      | st :: bil ->
+        let scope = decr scope in
+        match st with
+        | Move (v, Int _) when Var.is_virtual v ->
+	  if is_used v scope then loop (st :: acc) scope bil
+	  else loop acc scope bil
+        | If (cond, yes, no) ->
+	  let yes = loop [] (yes @ scope) yes in
```
ivg: also, very inefficient

plugins/bil_pipe/bil_const_propagation.ml
```diff
@@ -0,0 +1,236 @@
...
+    loop [] bil bil
+
+end
+
+module P = Propagate(Monad.State)
+
+let propagate_consts xs = Monad.State.eval ((new P.t)#eval xs) P.new_ctxt
+let eliminate_dead_code = Dead_code.eliminate
+
+let propagate_consts bil =
```
ivg: so, you're saying that "constant propagation is constant folding, constant propagation, and dead code elimination"?

plugins/bil_pipe/bil_pipe_main.ml
```diff
@@ -0,0 +1,62 @@
...
+let print fmt s =
+  Format.fprintf fmt "%s" @@
+  match s with
+  | true  -> "enable"
+  | false -> "disable"
+
+let switcher ?(default=true) =
+  Config.(param (converter parse print default) ~default)
+
+let simpl = Bil.fold_consts
```
ivg: so, simplification is actually the what you call propagate consts and is a fixed point of `FC . PC . DE`. 

I want to see the following options:

```
--bil-normalization=<level> -- defaults to 1 (with 0 keeps it denormalized)
--bil-optimization=<level> -- defaults to 1 (or 0, I don't know yet)
```

I'm not sure that we need to provide right now options between whether dead code elimination should be applied, or constant folding. I think we can start with the simple choice: optimization=0 - no simplifications are applied. optimization=1 -- all optimizations are applied.

plugins/bil_pipe/bil_pipe_main.ml
```diff
@@ -0,0 +1,62 @@
...
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "Applies analysises to a instruction bil code" ;
+      `Pre "
+The following bil analysises are in default pipeline:
+Constant Folding and Simplification
+Bil Normalization
+Constant Propagation
+Memoization
```
ivg: Though I don't see any memoizations here, you can actually help yourself with a hashtable that holds simplified BIL. You can compute a digest of a BIL code it should be faster than applying all your optimizations.

The main consideration here, is that we will leak a lot of memory in such memoization. Thus you shall implement some sort of cache eviction mechanism, e.g., the intuitive suggestion would be:
1) do not store trivial instructions that are not simplified
2) store a fixed number of instructions, say a hundred
3) once the limit is reached evict a random instruction

lib/bap/bap.mli
```diff
@@ -6593,6 +6593,11 @@ module Std : sig
       target.  *)
   val register_target : arch -> (module Target) -> unit
 
+  (** [provide_bil_transformation name f] register a new function [f] with a [name]
+      that will be applied to [bil] once instruction lifted and (optionaly)
+      after other trnasformations. *)
+  val provide_bil_transformation : string -> (bil -> bil) -> unit
```
ivg: We need to align this interface with the upcoming bap services and, at the same time, consistent with existing Project passes. 
1) move the function into the Bil module, and rename to `register_pass`
2) add a function `passes : unit -> pass list` 
3) add a function `select_passes : pass list -> unit`
4) add module `Bil.Pass` which implements the `Printable` interface.  

Documentation for the new functions (and the specifications)

```ocaml
(** [register_pass name pass] provides a pass to the BIL transformation pipeline. 
     
     The BIL transformation pipeline is applied after the lifting procedure, i.e., it is embedded into 
     each [lift] function of all Target modules. (You can selectively register_passes based on architecture
     by subscribing to the [Project.Info.arch] variable). All passes that  were in the selection provided to 
     the [select_passes] are applied in the order of the selection until the fixed point is reached or a loop 
     is detected. 
     By default, no passes are selected. The [bil] plugin provides a user interface for passes selection, 
     as well as some useful passes. 
*)
val register_pass : string -> (bil -> bil) -> unit

(** [select_passes passes] select the [passes] for the BIL transformation pipeline. 
     See {!register_pass} for more information about the BIL transformation pipeline *)
val select_passes : pass list -> unit

File: plugins/bil/bil_const_propagation.ml [hunk 12]
```diff
@@ -0,0 +1,184 @@
...
+    (object
+      inherit [var list] Stmt.visitor
+      method! enter_move v _ defs = v :: defs
+    end)#run bil []
+
+  let has_unconditional_jmps bil =
+    List.exists bil
+      ~f:(function | Jmp _ -> true | _ -> false)
+
+  class bili = object(self)
```
ivg: `bil_propagator`

plugins/bil/bil_const_propagation.ml
```diff
@@ -0,0 +1,184 @@
...
+
+  type const = Def of word | Top
+  type ctxt = const Var.Map.t
+  type 'a r = ('a, ctxt) SM.t
+
+  let new_ctxt = Var.Map.empty
+  let find = Map.find
+  let update = Map.add
+
+  class expi = object(self)
```
ivg: `exp_propagator`

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,77 @@
+open Core_kernel
+open Bap.Std
+open Regular.Std
+
+include Self ()
+
+(* variable is considred dead if it's a virtual and equal to const and
+   assigned only once. Assuming const propagation was ran before *)
```
ivg: why do we remove only constant assignments? What's wrong with removing a bogus `v = RAX`?

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,77 @@
...
+  provide_bil_transformation "internal" @@
+  fun bil -> apply bil [
+    norml norml_level;
+    simpl simpl_level;
+  ]
+
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "Applies analysises to a instruction bil code" ;
```
ivg: Selects the list of analyses to be applied during the lifing to BIL code.

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,77 @@
...
+      `Pre "
+The following bil analysises are in default pipeline:
+Constant Folding and Simplification
+Bil Normalization
+Constant Propagation
+";
+      `S "SEE ALSO";
+      `P "$(b,bap.mli)"
```
ivg: This is an incorrect link, see how they are made in the rest of manpages. Also, there is no need to reference the bap.std library at all.

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,77 @@
...
+  fun bil -> apply bil [
+    norml norml_level;
+    simpl simpl_level;
+  ]
+
+let () =
+  let () = Config.manpage [
+      `S "DESCRIPTION";
+      `P "Applies analysises to a instruction bil code" ;
+      `Pre "
```
ivg: This paragraph is not necessary, describe how passes are selected in the command line interface.

plugins/bil/bil_main.ml
ivg: Instead, provide some documentation from BIL here, i.e., explain what is the pipeline, how it is registered etc. Imagine, that you're a user that has no idea what it is about and how to use it. You should be able to provide and select new passes after the man page is read.

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,77 @@
...
+";
+      `S "SEE ALSO";
+      `P "$(b,bap.mli)"
+    ] in
+  let simpl =
+    let doc = "Applies expressions simplification." in
+    Config.(param int ~default:1 "simplification" ~doc) in
+  let norm =
+    let doc = "Produces a normalized BIL program" in
+    Config.(param int ~default:1 "normalization" ~doc) in
```
ivg: add `passes` option, that if defaults to an empty list. Build the transformation pipeline as a concatenation of normalization, optimization, and passes.

plugins/bil/bil_memo.ml
```diff
@@ -0,0 +1,20 @@
+open Core_kernel
```
ivg: provide an mli file

plugins/bil/bil_main.ml
```diff
@@ -34,44 +33,46 @@ let eliminate_dead_code bil =
...
+    (fun bil -> Bil.fold_consts bil |> Bil_const_propagation.run |> eliminate_dead_code)
 
-let run norml_level simpl_level =
-  provide_bil_transformation "internal" @@
-  fun bil -> apply bil [
-    norml norml_level;
-    simpl simpl_level;
-  ]
+let run passes =
+  let passes = ["normalization"; "optimization"] @ passes in
```
ivg: nay, you shall keep the  normalization and optimization parameters. The selected set of passes should be 
```ocaml
norm n @ optim m @ passes
```
where `norm n` is the list of normalizations for level `n`, `optim m` is the list of optimizations on the optimization level `m`, and `passes` are the passes that are selected by a user.

This interface allows disabling both optimization and normalization, as well as an option to define arbitrary pipeline, e.g., `--bil-normalization=0 --bil-optimization=0 --bil-passes=my-normalization,my-optimization`

plugins/bil/bil_main.ml
```diff
@@ -34,44 +33,46 @@ let eliminate_dead_code bil =
...
+  let is_selected p =
```
ivg: your passes parameter should have type `Bil.pass list Term.t` and whatever the parameter returns should be used as a user selection.

plugins/bil/bil_main.ml
```diff
@@ -34,44 +33,46 @@ let eliminate_dead_code bil =
...
-      `P "Applies analysises to a instruction bil code" ;
-      `Pre "
-The following bil analysises are in default pipeline:
-Constant Folding and Simplification
-Bil Normalization
-Constant Propagation
-";
-      `S "SEE ALSO";
-      `P "$(b,bap.mli)"
+      `P "Establishes a pipeline of BIL analysis";
```
ivg: Use the following manpage:

Controls the BIL transformation pipeline. 

The BIL transformation pipeline is used to transform the BIL code of each machine instruction and is applied to the output of a lifter. Each piece of transformation is called $(i,a pass). This plugin enables full control over the pass selection as well as provides some predefined passes. See the $(b,--bil-list-passes) for a comprehensive list of
available transformations. The pipeline is formed as a concatenation of all normalization passes that correspond to the selected level of normalization (setting it to zero will disable normalizations), all optimization passes applicable to the selected level of optimization (again the zero level corresponds to an empty list of optimizations), and all passes selected via the `$(b,--bil-passes)` command line argument. All passes are applied until a fixed point is reached (i.e., until the outcome of the pipeline is equal to the income) or until a loop is detected. Thus to prevent infinite looping each registered pass should be a pure function of its inputs (same input shall result in the same output). 

A new pass could registered via the $(b,Bil.register_pass) function. It won't be automatically selected, so it should be specified explicitly via the $(b,--bil-passes) command line argument.

plugins/bil/bil_main.ml
```diff
@@ -34,44 +33,46 @@ let eliminate_dead_code bil =
...
+  Bil.register_pass "normalization" @@
+  Stmt.normalize ~keep_ites:false ~normalize_exp:false
 
-let simpl level =
-  if level = 0 then ident
-  else
-    Bil.fixpoint
-      (fun bil -> Bil.fold_consts bil |> Bil_const_propagation.run |> eliminate_dead_code)
+let () =
+  Bil.register_pass "optimization" @@
```
ivg: same as above, do not namesquash abstract names, optimization is an abstract name for a family of transformations, while your transformation is just a member equal among others. (This is the same as to use `list`, `set`, `int` as variable names)

plugins/bil/bil_main.ml
```diff
@@ -34,44 +33,46 @@ let eliminate_dead_code bil =
...
+  Bil.fixpoint
```
ivg: remove fixed point from here. You shall rely on the framework's fixed point instead.

plugins/bil/bil_live.ml
```diff
@@ -0,0 +1,28 @@
...
+          | Jmp e -> s :: bil, free e ++ live
+          | If (cond, yes, no) ->
+            let yes, live' = loop live yes in
+            let no, live'' = loop live no in
+            let live = free cond ++ live' ++ live'' in
+            Bil.if_ cond yes no :: bil, live
+          | Move (var,e) ->
+            if is_dead var live then bil, live
+            else s :: bil, free e ++ (live -- var)
+          | While (cond, body) ->
```
ivg: in the body of a while loop the liveness property is propagated in both directions, since the body has to outcoming edges, one that is connected with the rest of the program (and brings liveness from it) and another is from the loop body itself, that brings liveness from all statements of loop body. In your algorithm you effectively ignoring this information, thus you will remove variables whose liveness are guaranteed by preceding statements in the loop body. 
E.g.,

```c
x := 0;
y := 0;
while (i < 2) {
    y = x + 1;
    x = 1;
    i = i + 1;
}
```

assuming that `x` is not live and that `y` is live in the rest of the program, then your analysis will transform the program into
```c
x := 0;
// y := 0; -- killed since `y` is not live here anymore
while (i < 2) {
    y = x + 1;
    // x = 1; --- killed since `x` is dead
    i = i + 1; // this would be also removed, since you didn't treat cond right. 
}
```

To fix, an easy solution would be to stop killing _and_ removing once you enter a while loop. (Remember that this property is transitive to all contents of the while loop, e.g., in `while (c) {x; y; if (d) {p} else {q}; z}`, statements `p` and `q`, as well as `x`, `y`, `z` should be preserved).

lib_test/bap_types/test_optimizations.ml
```diff
@@ -0,0 +1,655 @@
...
+            ] [
+              will_survive := one;
+            ]
+          ]
+        ]
+      ] [];
+    ] in
+  bil ==> bil
+
+let contains_conditions =
```
ivg: I think you meant `includes_conditions`

lib_test/bap_types/test_optimizations.ml
```diff
@@ -0,0 +1,655 @@
...
+let x42 = int 42
+
+let error_of_bil input got =
+  sprintf "something went wrong for\n%s\ngot\n%s\n"
+    (Bil.to_string input) (Bil.to_string got)
+
+let alive_anyway = physical "alive_anyway"
+let will_survive = virtual_ "will_survive"
+let will_be_dead = virtual_ "will_be_dead"
+let could_be_dead = will_be_dead
```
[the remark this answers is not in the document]
ivg: and in any case, please don't give different names to the same bil variables, that only confuses

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,104 @@
...
+    "2", [bnf2]; ]
+
+let fold_consts =
+  Bil.register_pass "constant-folding" ~desc:"constant-folding" Bil.fold_consts
+
+let consts_propagation =
+  Bil.register_pass "constant-propagation" ~desc:"constant-propagation" Bil.propagate_consts
+
+let prune_dead =
+  Bil.register_pass "prune-dead"
```
ivg: prune-dead-virtuals

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,104 @@
...
+        function. It won't be automatically selected, so it should be
+        specified explicitly via the $(b,--bil-passes) command line
+        argument."
+    ] in
+  let norml =
+    let doc = "Specifies the BIL normalization level.
+      The normalization process doesn't change the semantics of
+      a BIL program, but applies some transformations to simplify it.
+      There are two BIL normal forms (bnf): bnf1 and bnf2, both
+      of which are described in details in $(b,bap.mli).
```
ivg: It should be a reference to documentation not an mli file, i.e., `$(b,bap)(3)`, also add the SEE ALSO section with the reference to BAP, and bapdoc will make it clickable.

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,104 @@
...
+      Briefly, $(b,bnf1) produce the BIL code  with load expressions
+      that applied to a memory only. And $(b,bnf2) also adds some
+      more restrictions like absence of let-expressions and makes
+      load/store operations sizes equal to one byte.
+      So, there are next possible complementations for
```
ivg: s/complementations/options

plugins/bil/bil_main.ml
```diff
@@ -0,0 +1,104 @@
...
+      normalization level:
+      $(b,0) - disables normalization; $(b,1) - produce Bil in bnf1;
+      $(b,2) - produce Bil in bnf2" in
+    Config.(param (enum normalizations) ~default:[bnf1] ~doc "normalization") in
+  let optim =
+    let doc = "Specifies a level of optimization.\n
```
ivg: the optimization level

PR #833, opened 2018-05-04
Adding support for CDQ, CDQE, CQO, CWD, CWDE, CBW opcodes

ethan42:


ivg:
Thanks!

Assigning @gitoleg as he is currently on the lifters stuff. Oleg, can you also check if we need to enable any checks in the veri suite, that will cover the new functionality?

ethan42:
@gitoleg : just ran ocp-indent on them. let me know if you have a specific config you'd like me to use.

re: opcodes: I actually had to add the instruction because of 0x99 opcodes (they seem to be common in some of the binaries I'm looking at)

gitoleg:
@ethan42  so my faith that they are seldom was wrong : )

PR #834, opened 2018-05-09
fixes bug in reconstructor

gitoleg:
This PR fixes a BUG in reconstructor, where we actually ignore lots of calls


ivg:
Seriously, such kind of issues should be fixed as soon as possible. Don't stockpile fixes in your own branch!

PR #836, opened 2018-05-22
Fix packed x86

gitoleg:
This PR brings a lot of changes in x86 packed instructions. The main goal is to make a bil code closer to the BIL normal form, i.e. to reduce an amount of an additional code injected during the normalization process as well a time spent for lifting.
As a result, `bap` works better with files that previously took approximately infinite time.

Affected instructions are:
- different cases of `pcmp`, `pmin`, `pmax`;
- instructions like `psrl`, `psll`, `pavg`, `xorp`;
- instructions that compare strings: `pcmpistri` 

Just few numbers to compare: 
it was:
```
$: echo "\x66\x0f\xda\xc8" | bap-mc --show-bil | wc -l
327678
$: echo "\x66\x41\x0f\x3a\x63\x08\x4a" | bap-mc --show-bil | wc -l
Fatal error: out of memory.
```
it is:
```
$:echo "\x66\x0f\xda\xc8" | bap-mc --show-bil | wc -l
132
$: echo "\x66\x41\x0f\x3a\x63\x08\x4a" | bap-mc --show-bil | wc -l
139
```

Also, This PR adds few instructions to x86-lifter:
`pmaxsb`, `pmaxsw`, `pmaxuw`, `pmaxud`, `pmaxsd` and tests for them.


lib_test/x86/test_pcmp.ml
```diff
@@ -0,0 +1,216 @@
...
+   3) pmaxub xmm1, xmm0
+      expected: all bytes in xmm1 should be equal
+      to corresponded bytes in xmm0, except first and last
+      bytes, which stay unchanchged
+   4) pmaxsb xmm1, xmm0
+      expected: xmm1 = xmm0
+
+*)
+
+open Core_kernel.Std
```
ivg: try to use `open Core_kernel` in the new code, the `Std` module is deprecated

PR #844, opened 2018-06-19
Remove special handling of jumps/cpuexns during linearization.

ethan42:
This is necessary for correctly handling control flow in BIL
containing the above patterns. One example explaining why this
is necessary is the following:

```
if false {
  CPUExn 10
}
x := 5
```

The above would be translated to:

```
    goto %takeoff
landing:
    goto %next
takeoff:
    when false cpuexn N return %landing
next:
     x := 5
```

Note that the `takeoff` basic block is missing the
possible control flow transition to the `next` block.

By removing these special cases jmps and cpuexn are handled as
normal statements within an `if` block.


ivg:
For the note, this also fixed another bug, (line 96), that will linearize the following BIL program
```
if (c) {
  jmp d
}
x := y
```
to
```
I (when c jmp d)
I (x := y)
```

that will result in a totally wrong block:
```
2: x := y
1: when c jmp d
```

PR #845, opened 2018-06-23
Adding support for XMM/YMM on x86 and enabling the user to map `reg -> Var.t`

ethan42:
This MR adds:

* XMM/YMM to the set of register types returned by the LLVM disassembler and the
  related utility functions.

* Adds an interface under `x86_llvm_env` that enables the user to fetch the variables
  (physical CPU registers) that a register operand of an instruction refers to.


ethan42:
Starting the discussion on how to enable users to get variables out of register operands for x86. @ivg: let me know if this is along the lines you were thinking.

lib/x86_cpu/x86_env.ml
```diff
@@ -160,6 +162,61 @@ module R32 = struct
...
   let ymms = Array.sub ymms 0 8
 
+  (* mapping registers to arch variables *)
+  let of_reg reg =
+    let internal_reg = X86_asm_reg.decode reg in
+    match internal_reg with
+    | None -> None
+    | Some reg ->
+      (match reg with
+       | `AL | `AH | `AX | `EAX -> Some rax
```
ivg: shouldn't we map to `exp` instead and preserve the width of the register? With such representation, we have that `bitwidth reg <> bitwidth var`.

lib/x86_cpu/x86_env.ml
```diff
@@ -123,6 +124,7 @@ module type ModeVars = sig
   val r : var array
   val nums : var array
   val ymms : var array
+  val of_reg : reg -> var option
```
ivg: Changing the module type is a breaking change since we're changing an abstraction, which is contravariant, as module types are used as parameters to functors. 

Moreover, the `of_reg` function is LLVM specific, and we are hiding this. For example, it relies on a `decode` function, that presumes the llvm encoding, and fails with the `Failure` exception if it is not the case. So my suggestion would be to factor it out in a separate module, which has an explicit dependency on llvm. (We don't really need the build time dependency, as I'm envisioning cases, when we need to analyze LLVM output, without actually using LLVM library, e.g., when a serialized version of a project is passed to another machine). But we still need to make it clear, both in documentation and in the choice of names, that this mapping is LLVM specific, e.g., `bap_x86_llvm_cpu_env.ml`.

lib/x86_cpu/x86_env.ml
ivg: Basically, we can have `to_var : reg -> var option` that will be a one-to-one match, e.g., 
```ocaml
to_var `AL = None
to_var `EAX = if arch = `x86 then Some rax else None
to_var `RAX = if arch = `x86_64 then Some rax else None
```

and we can also have a `to_base_var` function, that will for the given register return its base vairant, e.g., RAX for AL in x86-64 mode.

lib/x86_cpu/x86_env.ml
```diff
@@ -160,6 +162,61 @@ module R32 = struct
...
+       | `RDI | `RSI | `RBP | `RSP
+       | `R8 | `R9 | `R10 | `R11
+       | `R12 | `R13 | `R14 | `R15
+       | `RIP
+       | `XMM8 | `YMM8 | `XMM9 | `YMM9
+       | `XMM10 | `YMM10 | `XMM11 | `YMM11
+       | `XMM12 | `YMM12 | `XMM13 | `YMM13
+       | `XMM14 | `YMM14 | `XMM15 | `YMM15
+       ->
+         assert false (* incorrect architecture *)
```
ivg: there is no static guarantee that this function wouldn't be called for a wrong architecture. So we either need to raise `invalid_argument` here and specify this clearly, that this function is defined only for x86, or use `result` as the return type, to indicate that the function is not total.

lib/x86_cpu/x86_env.ml
```diff
@@ -197,6 +254,58 @@ module R64 = struct
...
   let ymms = ymms
 
+  (* mapping registers to arch variables *)
+  let of_reg reg =
+    let internal_reg = X86_asm_reg.decode reg in
+    match internal_reg with
+    | None -> None
+    | Some reg ->
+      (match reg with
+       | `AL | `AH | `AX | `EAX | `RAX -> Some rax
```
ivg: I think that this code could be reused from the x86 representation.

ethan42: The `rax`, etc reference different objects. I can introduce sharing of course, but the function would have to be parameterized by the architecture. is that preferable?

lib/x86_cpu/x86_env.ml
ethan42: `invalid_argument` sounds good, will switch it

lib/x86_cpu/x86_env.ml
ethan42: happy to change it. however, this doesn't break semantic versioning does it? the previous interface remains intact I thought.

lib/x86_cpu/x86_env.ml
ivg: yep, I don't think that it would be to heavy to parametrize it with the modvars module. It could be hidden internally or made public, if preferred.

lib/x86_cpu/x86_env.ml
ethan42: This would be equivalent to the RR.get function from x86_tools. Variables are sometimes a bit more convenient to work with compared to exp. In any case, I'll rename to `to_base_var` to avoid confusion. I'll see if I can move the RR.get functionality in here as well. Let me know what you think.

lib/x86_cpu/x86_env.ml
ivg: it breaks, it changes the module type of ModeVars, thus any function/functor that were taking ModeVars as a parameter will break. 

Basically, semantic version treats the subtyping of a library interface as a non-breaking change. So we can make the interface more specific by adding new things. But as always with subtyping 
we need to distinguish between covariant and contravariant types. So we may add new stuff to covariant types, and erase fields from contravariant. We can't touch invariant types. In our case, ModeVars is invariant, so we can't add or remove anything to it without breaking code.

lib/x86_cpu/x86_env.ml
ivg: sounds good, as long as we make it explicit and giving a user the options to choose from.

ethan42:
@ivg: what about now? I was thinking about possibly merging some of the X86_asm* modules, but not sure if that's a good idea or not.

ivg:
@ethan42, are you sure that the patch is right? Looks like you've missed some files... like I don't see an addition of the new registers in x86_asm_reg_types.ml only the deletion.

lib/x86_cpu/x86_llvm_env.ml
```diff
@@ -0,0 +1,76 @@
...
+
+
+let base_var mode reg =
+  let module CPU = (val (X86_env.vars_of_mode mode)) in
+  let internal_reg = X86_asm_reg.decode reg in
+  match internal_reg with
+  | None -> None
+  | Some reg ->
+    if mode = X86_types.X86 then (
+      match reg with
```
ivg: could we simplify this check by using polymorphic variants abbreviation patterns, e.g.,

```ocaml
match reg with
| #r64 | #r128 | #r256 | `RIP when mode = X86 -> 
   invalid_argf "register %s is not valid in the x86 mode" (string_of_reg reg) ()
| ...
```

ethan42: partially. the problem is some of the smaller variants are also infeasible (e.g., the lower part of r8)

ethan42:
oops, you are correct (I blame force-pushing). let me try again.

lib/x86_cpu/x86_llvm_env.ml
ethan42: pushed a slightly improved version

ivg:
LGTM, but can you double check that the code is ocp-indented?

ethan42:
@ivg: done

Issue #846, opened 2018-06-26
Bil Store Eval doesn't infer types properly


DIJamner:
The call to `infer` in this example returns `Ok`, but `eval` fails with a type error:
```
open Bap.Std;;
open Bil;;
let addr = int (Word.zero 64);;
let m = (unknown "mem" (Type.mem `r64 `r8));;
let m' = store ~mem:m ~addr (int (Word.zero 8)) LittleEndian `r8;;
Type.infer m';;
Exp.eval m';;
```

This is likely due to one of the type inference functions used in `Exp.eval`, probably [bap_helpers.ml#679](https://github.com/BinaryAnalysisPlatform/bap/blob/b4ecb65eafb318e885541be0c3602e73f1e6519e/lib/bap_types/bap_helpers.ml#L679)

gitoleg:
fixed in https://github.com/BinaryAnalysisPlatform/bap/pull/847

Issue #843, opened 2018-06-14
BIL Store Typechecker too lenient


DIJamner:
The call to `infer` in this example returns `Ok`, but `eval` fails with a type error:
```
Open Bap.Std;;
let m = Bil.store ~mem:(Bil.unknown "mem" (Type.mem `r64 `r8))
          ~addr:(Bil.int (Word.one 8)) (Bil.int (Word.one 8)) 
          LittleEndian `r8;;
Type.infer m;;
Exp.eval m;;
```

DIJamner:
The infer function at [`bap_helpers.ml#L226`](https://github.com/BinaryAnalysisPlatform/bap/blob/a691e22626e1541e9ad07e4f49baf9c1f982145d/lib/bap_types/bap_helpers.ml#L226) only checks that the memory's address width is at least the width of the input address. Is there a case where BAP should allow it to be greater than or should it always be equal?

ivg:
gosh, what a typo! it should be `<>` of course, not `<`

gitoleg:
fixed in https://github.com/BinaryAnalysisPlatform/bap/pull/847

PR #847, opened 2018-06-26
Fix mem types

DIJamner:
Fixes issues [#846](https://github.com/BinaryAnalysisPlatform/bap/issues/846) and [#843](https://github.com/BinaryAnalysisPlatform/bap/issues/843) by using the right comparison and adding a necessary case to a simplified type inference function.


ivg:
@gitoleg, please review and merge when ready.

PR #851, opened 2018-08-02
updates travis env

gitoleg:
After last changes in `ocaml-ci-scripts`, a default version of `opam` became `2.0`, that breaks our builds. The simplest fix would be just to set an environment variable OPAM_VERSION=1.2.2`, but it doesn't help. The reason is in the order of setting env variables and installation of packages in the deep of those scripts.
And we don't want to wait neither for fix nor for next breaking changes from their side, so we will use just a couple of their old scripts that works fine for us.

### DIFF
--- a/.travis-ocaml.sh
+++ b/.travis-ocaml.sh
@@ -1,0 +1,179 @@
+## basic OCaml and opam installation
+
+full_apt_version () {
+  package=$1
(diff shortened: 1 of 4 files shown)


ivg:
welcome back `.travis-opam.sh`! We missed you.

Issue #720, opened 2017-10-04
The IDA service should cleanup minidump before 


ivg:
When IDA crashes it creates a minidump file in the `/tmp/ida` folder. Every time it is started afterwards it looks for the minidump and if it is present an annoying window popups.It will continue showing until the file is removed. Thus it is the task for the IDA service to look at the `/tmp/ida` folder, and if it contains any minidumps, remove them while issuing a warning into the log.

ivg:
FYI, the file looks like `<UUID>.dmp`, e.g.,
```
ls /tmp/ida/
7db70f7b-9c37-ffbd-1484485b-51bb481e.dmp
```

just unlinking this file fixes IDA and it could be run without any problems. You can also delete the whole directory if you find it more convenient.

ivg:
it is in the worklist since October. Please, do it finally.

PR #853, opened 2018-08-08
cleans /tmp/ida every time ida is called

gitoleg:
fix https://github.com/BinaryAnalysisPlatform/bap/issues/720

Just remove dump files from `/tmp/ida` before any `ida` launch

### DIFF
--- a/plugins/ida/bap_ida_service.ml
+++ b/plugins/ida/bap_ida_service.ml
@@ -79,8 +79,17 @@ let setup_headless_env path =
       if old_path <> "" then Unix.putenv var old_path
     with _ -> ()
 
+let cleanup_minidump () =
+  let dump_path = "/tmp/ida" in
+  if FileUtil.(test Exists dump_path) then
+    let ext = FilePath.DefaultPath.extension_of_string "dmp" in
+    let is_dmp = FileUtil.(test (Has_extension ext)) in
+    FileUtil.ls dump_path  |>
+    List.iter ~f:(fun f -> if is_dmp f then FileUtil.rm [f])
+
 (* ida works fine only if everything is in the same folder  *)
 let run (t:ida) cmd =
+  cleanup_minidump ();
   let cwd = Unix.getcwd () in
   let clean = match t.curses with
     | Some path -> setup_headless_env path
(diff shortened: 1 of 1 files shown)


ivg:
Please, look into `plugins/cache/cache_main.ml` for the inspiration on file locks.

plugins/ida/bap_ida_service.ml
```diff
@@ -79,8 +81,24 @@ let setup_headless_env path =
       if old_path <> "" then Unix.putenv var old_path
     with _ -> ()
 
+let cleanup_minidump () =
+  let is_dump x = String.equal (Filename.extension x) ".dmp" in
+  let dump_path = "/tmp/ida" in
+  if Sys.file_exists dump_path then
```
ivg: it will fail if `dump_path` is a file, add `Sys.is_directory`

plugins/ida/bap_ida_service.ml
```diff
@@ -79,8 +81,24 @@ let setup_headless_env path =
       if old_path <> "" then Unix.putenv var old_path
     with _ -> ()
 
+let cleanup_minidump () =
+  let is_dump x = String.equal (Filename.extension x) ".dmp" in
```
ivg: use `Filename.check_suffix n ".dmp"`
in general use the most specialized function available. It conveys you intentions better and is usually more efficient and straightforward

plugins/ida/bap_ida_service.ml
```diff
@@ -79,8 +81,24 @@ let setup_headless_env path =
...
+  let dump_path = "/tmp/ida" in
+  if Sys.file_exists dump_path then
+    FileUtil.ls dump_path  |>
+    List.filter ~f:is_dump |> function
+    | [] -> ()
+    | files ->
+      info "ida minidump is not empty";
+      let lock = sprintf "%s/lock" dump_path in
```
ivg: protect this code with finally, to ensure that the file is always unlocked.

plugins/ida/bap_ida_service.ml
```diff
@@ -79,8 +81,24 @@ let setup_headless_env path =
...
+      let lock = Unix.openfile lock Unix.[O_RDWR; O_CREAT] 0o640 in
```
ivg: since the lockfile is global it should be readable (and thus shareable) by all users, so provide corresponding rights.

Consider the scenario when BAP is run by two different users, the first user will create a lock file, while the second will not be able to graph the lock as she will fail on `open`.

Issue #764, opened 2018-02-07
`deprecated` parameter doesn't emit a warning when used with flags


ivg:
If a command line parameter is specified as a flag, either directly using the `flag` function or indirectly, using the `as_flag` parameter of the `param` or `param_all`,  the deprecation warning is not emitted when a command line parameter is specified on the command line. 

This is because the converter is never called in case of flags.

PR #854, opened 2018-08-09
Fix deprecated args

gitoleg:
also warns if a deprecated parameter is used as a flag
fix https://github.com/BinaryAnalysisPlatform/bap/issues/764

### DIFF
--- a/lib/bap/bap_self.ml
+++ b/lib/bap/bap_self.ml
@@ -17,7 +17,7 @@ module Create() = struct
 
   let manifest =
     try Bundle.manifest bundle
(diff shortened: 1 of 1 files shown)


lib/bap/bap_self.ml
```diff
@@ -102,9 +102,15 @@ module Create() = struct
     type 'a parser = string -> [ `Ok of 'a | `Error of string ]
     type 'a printer = formatter -> 'a -> unit
 
+    let warn_if_deprecated msg name is_used = match msg with
```
ivg: it should be just `warn_if` ...

PR #855, opened 2018-08-09
optimizes reconstructor, symtab, and brancher performance

gitoleg:
This PR rectifies reconstructor, symtab and brancher with a respect to performance, without addition of a new behavior or breaking of an existing one.

Changes:
### reconstructor
Rewrote it and reduced the number of iterations over nodes/edges of cfg.
The main point is how we build the `filtered` cfg. Previously, we did like the following: take all nodes, add them ony-by-one to an empty cfg. So it was an iteration over all nodes of cfg. Then iterate over all edges, and add each of them if their's destination block is not root. 
Also, we did one more iteration over all nodes in `find_calls` function.

So what I'm proposing here is just to iterate once over the whole cfg to find all roots. and then build the `filtered` cfg by iterating over the roots and removing their input edges.

### symtab
There was a bit inefficient implementation of `add_symbol` function, so every addition of a symbol called filter on the whole table, although there were enough info to reduce such calls.
Concerning details. Every time we add symbol to the symtab we also remember it's name and address of an entry block. Also we filter `memmap` from previously added entries,
that have same name and entry block. But, as we remembered the address of the
entry block and the name of the symbol, we can check, if `memmap` needed to be
filtered without event touching it! 

### bracnher
Just a small fix that checks if an instruction has jumps at all before the subsequent call of `fold_consts` that could be heavy for some of the instructions

### DIFF
--- a/lib/bap_disasm/bap_disasm_reconstructor.ml
+++ b/lib/bap_disasm/bap_disasm_reconstructor.ml
@@ -19,51 +19,47 @@ type reconstructor = t
 let create f = Reconstructor f
 let run (Reconstructor f) = f
 
-let find_calls name roots cfg =
-  let starts = Addr.Table.create () in
-  List.iter roots ~f:(fun addr ->
-      Hashtbl.set starts ~key:addr ~data:(name addr));
-  Cfg.nodes cfg |> Seq.iter ~f:(fun blk ->
-      let () =
-        if Seq.is_empty (Cfg.Node.inputs blk cfg) then
-          let addr = Block.addr blk in
-          Hashtbl.set starts ~key:addr ~data:(name addr) in
-      let term = Block.terminator blk in
-      if Insn.(is call) term then
-        Seq.iter (Cfg.Node.outputs blk cfg)
-          ~f:(fun e ->
-              if Cfg.Edge.label e <> `Fall then
-                let w = Block.addr (Cfg.Edge.dst e) in
-                Hashtbl.set starts ~key:w ~data:(name w)));
-  starts
+let roots_of_blk roots cfg blk =
+  let addr = Block.addr blk in
+  let term = Block.terminator blk in
+  let init =
+    if Set.mem roots addr || Seq.is_empty (Cfg.Node.inputs blk cfg)
+    then [blk]
+    else [] in
+  if Insn.(is call) term then
+    Seq.fold ~init (Cfg.Node.outputs blk cfg)
+      ~f:(fun rs e ->
+          if Cfg.Edge.label e <> `Fall then Cfg.Edge.dst e :: rs
+          else rs)
+  else init
+
+let find_calls cfg roots =
+  let roots = List.fold ~init:Addr.Set.empty ~f:Set.add roots in
+  Graphlib.depth_first_search (module Cfg)
+    cfg ~init:Block.Set.empty
+    ~enter_node:(fun _ blk all ->
+        roots_of_blk roots cfg blk |>
+        List.fold ~init:all ~f:Set.add)
 
 let reconstruct name roots cfg =
-  let roots = find_calls name roots cfg in
-  let init =
-    Cfg.nodes cfg |> Seq.fold ~init:Cfg.empty ~f:(fun cfg n ->
-        Cfg.Node.insert n cfg) in
-  let filtered =
-    Cfg.edges cfg |> Seq.fold ~init ~f:(fun cfg e ->
-        if Hashtbl.mem roots (Block.addr (Cfg.Edge.dst e)) then cfg
-        else Cfg.Edge.insert e cfg) in
-  let find_block addr =
-    Cfg.nodes cfg |> Seq.find ~f:(fun blk ->
-        Addr.equal addr (Block.addr blk)) in
-  Hashtbl.fold roots ~init:Symtab.empty
-    ~f:(fun ~key:entry ~data:name syms ->
-        match find_block entry with
-        | None -> syms
-        | Some entry ->
-          let cfg : cfg =
-            with_return (fun {return} ->
-                Graphlib.depth_first_search (module Cfg)
-                  filtered ~start:entry ~init:Cfg.empty
-                  ~enter_edge:(fun _ -> Cfg.Edge.insert)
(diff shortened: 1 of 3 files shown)


lib/bap_disasm/bap_disasm_symtab.ml
```diff
@@ -52,19 +52,27 @@ let merge m1 m2 =
...
 
+let filter t ((name,entry,_ ) as fn) =
+  if Map.mem t.names name || Map.mem t.addrs (Block.addr entry) then
+    remove t fn
+  else t
+
 let add_symbol t (name,entry,cfg) : t =
   let data = name,entry,cfg in
-  let t = remove t data in
+  let t = filter t data in
```
ivg: can you please justify why we can use `filter` instead of `remove` here? It's not quite obvious to me.

lib/bap_disasm/bap_disasm_symtab.ml
```diff
@@ -52,19 +52,27 @@ let merge m1 m2 =
...
+
 let remove t (name,entry,_) : t = {
   names = Map.remove t.names name;
   addrs = Map.remove t.addrs (Block.addr entry);
-  memory = Memmap.filter t.memory ~f:(fun (n,e,_) ->
-      not(String.(name = n) || Block.(entry = e)))
+  memory = filter_mem t.memory name entry;
 }
 
+let filter t ((name,entry,_ ) as fn) =
```
ivg: OK, so from the IRL discussion, the remove function is inefficient since it filters the memory mapping even if the symbol is not in the symtab, that fact that could be checked very fast with Set membership operation.

However, this is the problem of the `remove` function, that should be fixed on its side. So, please remove the filter function and add the corresponding check to the `remove` function (in fact a single test for inclusion in `addrs` would be sufficient.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,51 +19,45 @@ type reconstructor = t
...
-          Hashtbl.set starts ~key:addr ~data:(name addr) in
-      let term = Block.terminator blk in
-      if Insn.(is call) term then
-        Seq.iter (Cfg.Node.outputs blk cfg)
-          ~f:(fun e ->
-              if Cfg.Edge.label e <> `Fall then
-                let w = Block.addr (Cfg.Edge.dst e) in
-                Hashtbl.set starts ~key:w ~data:(name w)));
-  starts
+let callees_of_block cfg roots blk =
```
ivg: a bad choice of name, since you're adding the block itself to the resulting set it is no longer a callees_of_block. Suggestion, use `entries_of_block` since you're effectively collecting the entry nodes. 

Also, instead of doing lots of set unions you can pass the set along and use `Set.add`. It is more efficient.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,51 +19,45 @@ type reconstructor = t
...
+      ~f:(fun cls e ->
+          if Cfg.Edge.label e <> `Fall then
+            Set.add cls (Cfg.Edge.dst e)
+          else cls)
+  else init
+
+let update_callees cfg roots callees blk =
+  Set.union callees (callees_of_block cfg roots blk)
+
+let find_callees cfg roots =
```
ivg: rename to `collect_entries`

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,51 +19,45 @@ type reconstructor = t
...
-                  ~enter_edge:(fun _ -> Cfg.Edge.insert)
-                  ~start_tree:(fun n t ->
-                      if Block.equal n entry
-                      then Cfg.Node.insert n t
-                      else return t)) in
-          Symtab.add_symbol syms (name,entry,cfg))
+  let callees = find_callees cfg roots in
+  let is_call e = Set.mem callees (Cfg.Edge.dst e) in
+  let rec traverse fng node =
+    let fng = Cfg.Node.insert node fng in
```
ivg: Please, don't use abbreviations that are not common, suggestions,

`graph` or `g` or `prog` for the whole program graph
`cfg` for the CFG. 

Also, not `edg`, either `edge` or `e`.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,51 +19,45 @@ type reconstructor = t
...
-                  filtered ~start:entry ~init:Cfg.empty
-                  ~enter_edge:(fun _ -> Cfg.Edge.insert)
-                  ~start_tree:(fun n t ->
-                      if Block.equal n entry
-                      then Cfg.Node.insert n t
-                      else return t)) in
-          Symtab.add_symbol syms (name,entry,cfg))
+  let callees = find_callees cfg roots in
+  let is_call e = Set.mem callees (Cfg.Edge.dst e) in
+  let rec traverse fng node =
```
ivg: it is not `traverse` as this function builds a new graph. The better name `reachable` or `add_reachable` or just `add`.

lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,51 +19,45 @@ type reconstructor = t
...
+    let fng = Cfg.Node.insert node fng in
+    Seq.fold (Cfg.Node.outputs node cfg) ~init:fng ~f:(fun fng edg ->
+        if is_call edg then fng
+        else
+          let dst = Cfg.Edge.dst edg in
+          let visited = Cfg.Node.mem dst fng in
```
ivg: I think that the following is easier to read
```ocaml
if is_call edge then cfg
else 
  let cfg' = Cfg.Edge.insert edge cfg in
  if Cfg.Node.mem (Cfg.Edge.dst edge) cfg then cfg'
  else add cfg' node
```

gitoleg:
so, what about time - the new reconstructor on my machine with our favorite `dropbear` works 25 seconds less than the old one.

ivg:
relative timing would be much more interesting. 25 seconds out of 3 hours is much different from 25 seconds out of 30 seconds.

PR #856, opened 2018-08-14
fixes the bug that was producing unreachable blocks

gitoleg:
The root of the problem was the rooter, which was producing roots that were pointing in the middle of an instruction. Since the disassembler was cutting a block on the start of each root this led to an incorrect CFG with non-connected components. 

It's fine for a rooter to produce such garbage, however, the disassembler shall not take rooter's output as granted. The new implementation keeps track of the set of addresses that start valid instructions and once this set is fully discovered removes those roots that do not belong to this set.

### DIFF
--- a/lib/bap_disasm/bap_disasm_rec.ml
+++ b/lib/bap_disasm/bap_disasm_rec.ml
@@ -174,9 +174,10 @@ type blk_dest = [
 type stage1 = {
   base : mem;
   addr : addr;
-  visited : Span.t;
+  visited  : Span.t;
+  insns : Addr.Set.t;
   roots : addr list;
-  inits : addr list;
+  inits : Addr.Set.t;
   dests : dests Addr.Table.t;
   errors : (addr * error) list;
   lift : lifter;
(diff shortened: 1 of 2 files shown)


lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -174,9 +174,10 @@ type blk_dest = [
 type stage1 = {
   base : mem;
   addr : addr;
-  visited : Span.t;
+  visited  : Span.t;
+  insns : Addr.Set.t;
```
ivg: misleading name, let's name it `valid` instead

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -240,7 +243,8 @@ let next dis s =
     | r :: roots when not(Memory.contains s.base r) ->
       loop {s with roots}
     | r :: roots when Span.mem s.visited r ->
-      loop {s with roots}
+      if Set.mem s.insns r then loop {s with roots}
+      else loop {s with roots; inits = Set.remove s.inits r}
```
ivg: This code will remove an invalid root only in case if we have discovered that it is invalid before we extract it from the worklist. Since the worklist is an unordered sequence, this won't work in general.

gitoleg:
speaking about several segments case, this removal plays nice, just because the roots from other segment are not taking in account even without these changes.

PR #857, opened 2018-08-15
implements deadcode-elimination pass optimization 

gitoleg:
This PR implements the following optimizations of the `dead-code-elimination` plugin. 
1) adds caching of the plugin result;
2) memoizes sequences of computations;
3) computes SSA only once, instead of computing it on each step (this is the point of discussion)


plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,63 @@ let propagate_consts sub =
...
+let compute_dead free (_,ssa) = compute_dead free ssa
+let is_alive dead t = not (Set.mem dead (Term.tid t))
+let live_def dead blk = Term.filter def_t ~f:(is_alive dead) blk
+let live_phi dead blk = Term.filter phi_t ~f:(is_alive dead) blk
+
+let clean dead (sub,ssa) =
+  Term.map blk_t sub ~f:(live_def dead),
+  Term.map blk_t ssa ~f:(fun b -> live_def dead b |> live_phi dead)
+
+let rec run subs =
```
ivg: Translate the `run` to a function, that takes a _list_ of functions in the SSA form and the initial set of dead terms, and returns the final (maximal) set of dead terms. E.g.,

```ocaml

let rec run subs dead = 
   let free = ... in
   let dead' = .. in
   if Set.length dead' > Set.length dead 
   then run (clean dead' subs) dead'
   else dead
```

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,63 @@ let propagate_consts sub =
...
-  report_progress ~note:"updating" ();
-  if Set.is_empty dead then proj
-  else
-    Term.map sub_t prog ~f:clean |>
-    Project.with_program proj |>
-    process
+  if Set.is_empty dead then subs
+  else run (Seq.map subs ~f:(clean dead))
+
+let process prog =
```
ivg: obtain a set of dead terms and use `Term.map` to clean dead terms from each subroutine.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,63 @@ let propagate_consts sub =
...
+    method! enter_jmp t dst = Digest.add dst "%a" Jmp.pp t
+  end)#run
+    (Project.program proj)
+    (Digest.create ~namespace:"dead_code_elimination")
+
+
+let run proj =
+  let digest = digest proj in
+  let p =
+    match Program.Cache.load digest with
```
ivg: Never store the program (or project) in the cache, only the results of the computation. And then reapply them to the program. Otherwise, your pass will delete results from the upstream passes, which are otherwise transparent to you (e.g., their attributes)

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -1,10 +1,11 @@
 open Core_kernel.Std
 open Bap.Std
+open Regular.Std
 open Format
 include Self()
 
 let union ~init ~f =
-  Seq.fold ~init ~f:(fun acc x ->
+  List.fold ~init ~f:(fun acc x ->
```
ivg: FYI, there is `Set.union_list` (we needed this `union` helper just because there is no `union_sequence` function in core.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -24,8 +25,9 @@ let computed_def_use sub =
   def_use_collector#visit_sub sub (Var.Set.empty,Var.Set.empty)
 
 let sub_args sub =
-  Term.enum arg_t sub |>
+  Term.enum arg_t sub |> Seq.to_list |>
```
ivg: there is no need to transform it to `list`. I was asking only to use a list instead of a memoized sequence for the list of subroutines. Not like use list everywhere.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,72 @@ let propagate_consts sub =
...
+
+let apply prog deads =
+  Term.map sub_t prog ~f:(fun s ->
+      Term.map blk_t s ~f:(fun b ->
+          Term.filter def_t b ~f:(is_alive deads)))
+
+let process prog =
+  let subs = Term.enum sub_t prog |>
+             Seq.map ~f:Sub.ssa |>
+             Seq.to_list in
```
ivg: use `Set.to_list_rev` as the order doesn't matter.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,72 @@ let propagate_consts sub =
...
+    method! enter_arg t dst = Digest.add dst "%a" Arg.pp t
+    method! enter_def t dst = Digest.add dst "%a" Def.pp t
+    method! enter_jmp t dst = Digest.add dst "%a" Jmp.pp t
+  end)#run
+    (Project.program proj)
+    (Digest.create ~namespace:"dead_code_elimination")
+
+let run proj =
+  let digest = digest proj in
+  let prog = Term.map sub_t (Project.program proj) ~f:propagate_consts in
```
ivg: why do you do `propagate_consts` before you load something from a cache?

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -73,27 +75,72 @@ let propagate_consts sub =
...
 
-let rec process proj =
-  let prog = Project.program proj in
-  report_progress ~note:"ssa" ();
-  let subs = Term.enum sub_t prog |> Seq.map ~f:Sub.ssa in
+let is_alive dead t = not (Set.mem dead (Term.tid t))
+let live_def dead blk = Term.filter def_t ~f:(is_alive dead) blk
+let live_phi dead blk = Term.filter phi_t ~f:(is_alive dead) blk
+
+let clean dead ssa =
```
ivg: the clean procedure was doing `propagate_consts` before, which actually affects the dead code elimination procedure,  for example:

```
x := 12
y := x + 1
```
after constant propagation became

```
x := 12
y := 12 + 1
```

That makes `x` dead, so it could be removed.

plugins/dead_code_elimination/dead_code_diff.ml
```diff
@@ -0,0 +1,113 @@
+open Core_kernel
```
ivg: Ok, we need to do it better. Currently we have the following problems:
- [ ] too complex - I want you to target to 60 lines of code or so
- [ ] ineffective - the implementation is incredibly hungry for memory and is doing too much unnecessary work
- [ ] dirty - use hashtables.

The right approach, is start from bottom to top. Define small and easy to understand functions,

e.g.,

```ocaml
let changes_in_def d1 d2 = 
   let different = phys_equal d1 d2 || Exp.(Def.rhs d1 <> Def.rhs d2) in
   if different Tid.Map.singleton (Term.tid d2) (`NewRhs (Def.rhs d2)) else Tid.Map.empty
```
Then implement `changes_in_jmps` which will return 
```
[> `NewCond of exp | `NewDest of exp] Tid.Map.t
```

and 


```ocaml
let diff_of_blk b1 b2 = 
    let (++) = Map.merge  ~f:merge_changes in
    let was = elt ~take:Blk.elts b1 and now = elts ~take:Blk.elts b2 in
    Map.fold2 was now ~init:[] ~f:(fun ~key:_ ~data diffs -> match data with
      | `Both (`Def d1, `Def d2) -> changes_in_defs d1 d2 ++ acc
      | `Both (`Jmp b1, `Jmp b2) -> changes_in_jmp d1 d2 ++ acc
      | _ -> acc)
```

where the `elts` function creates a map of a term subterms, e.g.,

```ocaml
let elts ~take t =  
  Seq.fold ~init:Tid.Map.empty (take t) ~f:(fun elts key data  -> Map.add elts ~key ~data)
```

Now use the same approach to compare two subroutines, i.e.,

```ocaml
let diff_of_subs s1 s2 = 
    let was = elts ~take:(Term.enum blk_t) s1 and now = elts ~take:(Term.enum blk_t) s2 in
    Map.fold2 ...
```

and... you guess, the same on the program level.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -23,95 +38,155 @@ end
...
+    inherit Term.mapper as super
+    method! map_var v =
+      match Map.find vars v with
+      | None -> Bil.var v
+      | Some e -> e
+    method! map_exp e = Exp.fold_consts (super#map_exp e)
+  end)#map_sub sub
+
+(* A simple constant propagation. Note, the input is required to be in SSA. *)
+let propagate_consts checked sub =
```
ivg: this function does more work than needed, the term visitor will go through jmps and phi-nodes, maybe it is better to rewrite it as a fold over `Term.enum def_t blk`

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -23,95 +38,155 @@ end
...
-let rec process proj =
-  let prog = Project.program proj in
-  report_progress ~note:"ssa" ();
-  let subs = Term.enum sub_t prog |> Seq.map ~f:Sub.ssa in
+  substitute sub vars
+
+let clean checked dead sub =
+  Term.map blk_t sub ~f:(fun b -> live_def checked dead b |> live_phi dead)
+
+let return_from_ssa sub =
```
ivg: we don't need to return from ssa, just drop the optimized program after you compute the optimization data.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -23,95 +38,155 @@ end
...
+  Seq.fold ~init:Var.Set.empty ~f:(fun free sub ->
+      let free = Set.union free (Sub.free_vars sub) in
+      Seq.fold ~init:free (Term.enum arg_t sub)
+        ~f:(fun free arg ->
+            Set.union free (Exp.free_vars (Arg.rhs arg))))
+
+let union ~init ~f = List.fold ~init ~f:(fun xs x -> Set.union xs (f x))
+
+let process arch prog levels =
+  let checked = check_level (is_flag arch) levels in
```
ivg: checked? by whom... please rename.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -23,95 +38,155 @@ end
...
+  let free = free_vars prog in
+  let rec run subs =
```
ivg: apply optimization on each subroutine independently (except the protected set which should be computed interprocedurally). Then store the optimization for each subroutine in the cache (do not forget to include the digest of the protected set into the digest of a subroutine). 

In general always think about Google Chrome. The current implementation will require you to load in the memory and store in a list 3 google chromes if not more. While we can't even allow to store one. 

Basically, try to implement you analysis in a such way that it keeps in memory as small data as possible, ideally, no more than one subroutine. Also try to drop unused data as soon as possible, i.e., once you have a subroutine ready push it back to the project, store the cache, and keep going with the next one).

ivg:
Please, reimplement in a such way that in no point of time your algorithm requires to hold in memory more than one subroutine. It should be O(1) in terms of memory consumption wrt to the size of a program in number of subroutines.

plugins/dead_code_elimination/dead_code_elimination_main.ml
```diff
@@ -125,9 +168,19 @@ Config.manpage [
...
-]
+    `S "DEPENDENCIES";
+    `P "$(b,bap-plugin-api)(1)";
+
+    `S "SEE ALSO";
+    `P "$(b,bap-plugin-api)(1), $(b,bap-plugin-ssa)(1)";
+  ];
+  let level =
+    let doc =
+      "An integer that a level of optimization, i.e. what variables
```
ivg: Specifies the optimization level. The higher the value the more aggressive (and less safe) optimizations are applied. 

On level 1, we optimize only the synthetic code that was generated by the lifter. Since such code can't leave a scope of instruction it is not affected by the imprecision of a control flow graph. On level 2, we also move and optimize processor flags. This removes a significant amount of code and simplifies the program and is a fair compromise between safety and performance. (Since flags are rarely used non-locally). Finally, on level 3 we extend our analysis to all variables.

plugins/ida/bap_ida_service.ml
```diff
@@ -95,7 +95,7 @@ let cleanup_minidump () =
       Unix.lockf lock Unix.F_LOCK 0;
       protect ~f:(fun () ->
           List.iter files ~f:Sys.remove)
-        ~finally:(fun () -> Unix.lockf lock Unix.F_ULOCK 0)
+        ~finally:(fun () -> Unix.lockf lock Unix.F_ULOCK 0; Unix.close lock)
```
ivg: :+1:

plugins/dead_code_elimination/dead_code_diff.ml
```diff
@@ -0,0 +1,92 @@
...
+
+let changes_in_def d1 d2 =
+  let different = not (phys_equal d1 d2) && Exp.(Def.rhs d1 <> Def.rhs d2) in
+  if different then
+    let d2 = Def.map_exp ~f:drop_index d2 in
+    Some (Term.tid d1, Update (Rhs (Def.rhs d2)))
+  else None
+
+let changes_in_jmp j1 j2 =
+  let different = not (phys_equal j1 j2) && Jmp.(j1 <> j2) in
```
ivg: do not compare terms directly, this will involve the comparison of all attributes which could be very heavy. Compare only those parts that the algorithm could change.

PR #863, opened 2018-08-29
fixes calls search in reconstructor

gitoleg:
Actually, this PR just restores the previous behavior, that was lost due to lack of tests and comments.

So this PR:
1) extends a list of roots with addresses of such CFG nodes, that don't have any input edges. The effect is noticeable in a case of pure code input, since the list of roots is empty in such case, and therefore the reconstructor will create just an empty `symtab`.
2) adds comments in the code to avoid losing such behavior in the future.
3) updated documentation for reconstructor in `bap.mli`, since it looks quite outdated

### DIFF
--- a/lib/bap_disasm/bap_disasm_reconstructor.ml
+++ b/lib/bap_disasm/bap_disasm_reconstructor.ml
@@ -19,9 +19,18 @@ type reconstructor = t
 let create f = Reconstructor f
 let run (Reconstructor f) = f
 
+(* [is_start roots cfg blk] returns true if [blk] is a function start.
+   Block can be a function start in the next cases:
+   - block address is in a set of [roots]
+   - block doesn't have any input edges in [cfg].
+   The later is extremly important in case of pure code input. *)
+let is_start roots cfg blk =
+  Set.mem roots (Block.addr blk) ||
+  Cfg.Node.degree ~dir:`In blk cfg = 0
+
 let entries_of_block cfg roots entries blk =
   let entries =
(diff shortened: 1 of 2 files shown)


lib/bap_disasm/bap_disasm_reconstructor.ml
```diff
@@ -19,9 +19,18 @@ type reconstructor = t
 let create f = Reconstructor f
 let run (Reconstructor f) = f
 
+(* [is_start roots cfg blk] returns true if [blk] is a function start.
+   Block can be a function start in the next cases:
+   - block address is in a set of [roots]
+   - block doesn't have any input edges in [cfg].
+   The later is extremly important in case of pure code input. *)
```
ivg: Let's rephrase it as I'm afraid that we won't be able to decipher this later... today for example :)
```
A block is considered to be a function start if one of the following is true:
   - the block address belongs to the provided set of roots;
   - the block has no incoming edges

In general, all blocks should be reachable from the set of roots, thus a block without incoming edges should belong to the set of roots, so it might be tempting to say that the second clause is redundant. 

However, our implementation of the recursive descent disassembler can generate blocks that are not reachable from the initial set of roots. This can happen only if the set of roots is empty, which is treated as a special case, that instructs the disassembler to treat the first byte of the provided input as a root. 

Beyond being a questionable design decision, this behavior is actually just an example of a more general problem. The reconstructor and disassemblers could be called with different sets of roots. So, in general, we can't rely on the set of roots passed to the reconstructor for functions start classification. 
```

PR #865, opened 2018-08-30
saves cache index in bin_prot

gitoleg:
Previously we used `sexp` for storing index, but it turned out that it is very slow. And in this PR we save index with `bin_io`, that solves the problem.
Also, this PR fixes the issue with closing of lock files.

### DIFF
--- a/plugins/cache/cache_main.ml
+++ b/plugins/cache/cache_main.ml
@@ -80,13 +80,44 @@ module Index = struct
     then clean (evict_entry idx)
     else idx
 
-  let remove_entry e =
-    Sys.remove e.path
+  let remove_entry e = Sys.remove e.path
 
   let remove_files old_index new_index =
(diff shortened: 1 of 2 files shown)


plugins/cache/cache_main.ml
```diff
@@ -80,31 +80,71 @@ module Index = struct
...
+      let tmp =
+        Filename.temp_file ~temp_dir:(cache_dir ()) "tmp" "index" in
+      let fd = Unix.(openfile tmp [O_RDWR; O_CREAT] 0o600) in
+      let size = T.bin_size_t data in
+      let () =
+        try
+          let buf = Bigstring.map_file ~shared:true fd size in
+          let _ = T.bin_write_t buf ~pos:0 data in
+          Unix.close fd
+        with e -> Unix.close fd; raise e in
```
ivg: add `Sys.remove tmp` to cleanup after yourself in case of a problem.

Issue #861, opened 2018-08-21
implement IDA brancher


ivg:
The IDA part of the brancher has been implemented a long time ago in #550 and since then [was added][1] to the master branch of the bap-ida-python integration package. 

The BAP part though is still not there. What is left, is to implement a parser that will parse the brancher's service output and implement a brancher interface. And debug it. 

[1]: https://github.com/BinaryAnalysisPlatform/bap-ida-python/blob/master/plugins/bap/utils/ida.py#L86-L92

XVilka:
@ivg so this can be closed, right?

ivg:
Indeed! Finally :)

PR #868, opened 2018-09-05
Add IDA Pro brancher.

abhcs:
These changes allow a user to specify the argument '--brancher=ida' to ask BAP to utilize IDA's brancher information. These changes have been adapted from pull request #550 for #861. Pull request #550 was also concerned with kernel symbol recovery; only the changes specific to brancher information from IDA have been used in the current changes.

### DIFF
--- a/plugins/ida/ida_main.ml
+++ b/plugins/ida/ida_main.ml
@@ -136,6 +136,95 @@ let loader path =
             else code, Memmap.add data mem sec) in
   Project.Input.create arch path ~code ~data
 
+let int64_to_word arch = 
+  let addr_size = Arch.addr_size arch in 
+  (Word.of_int64 ~width:(Size.in_bits addr_size)) 
+
+(* copied from lib/bap_disasm/bap_disasm_brancher.ml *)
+let kind_of_dests = function 
+  | xs when List.for_all xs ~f:(fun (_,x) -> x = `Fall) -> `Fall 
+  | xs -> if List.exists  xs ~f:(fun (_,x) -> x = `Jump) 
+    then `Jump 
+    else `Cond 
+
+let print_addr_with_dests addr dests = 
+  Format.printf "%x: %a\n" 
+    (Int64.to_int_exn addr) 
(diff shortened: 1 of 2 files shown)


plugins/ida/ida_main.ml
```diff
@@ -136,13 +136,103 @@ let loader path =
...
+    let default_edge = ListLabels.assoc_opt 
+        (Some !dest_addr) 
+        default_dests 
+    in 
+    match default_edge with 
+    (* If dest_addr is also found by the default BIL brancher, 
+     * we reuse the edge type *) 
+    | Some kind -> (Some !dest_addr, kind)::acc 
+    (* Else we use a heuristic to guess the edge type *) 
+    | None -> (Some !dest_addr, (kind_of_dests default_dests))::acc
```
abhcs: Note that I use a different heuristic from #550, which assumed that the unknown type was a jump.

plugins/ida/ida_main.ml
```diff
@@ -136,13 +136,103 @@ let loader path =
             else code, Memmap.add data mem sec) in
   Project.Input.create arch path ~code ~data
 
+let int64_to_word arch =
```
abhcs: A constant width (32) was used in #550: I thought that this was more appropriate.

plugins/ida/ida_main.ml
ivg: I don't think that we need any heuristic here, as soon as we have a `Fall` edge, it shouldn't really matter whether a jump is conditional or not, or do you experience any problems when you've set all to `Jump` kind?

If indeed it has negative impact, then you can use `Insn.(is_conditional insn)` and mark all edges with other flow as `Cond` if it is true. (Note, this predicate will use instruction semantics (i.e., BIL) if available, and will fallback to the semantic information available from the disassembler backend).

abhcs: Setting all to jump works, but its just a little weird that the Brancher module expects more detailed information i.e. jump types but does not use it.

PR #871, opened 2018-09-13
provides information sensitivity analysis

ivg:
The analysis is based on the Primus Taint Engine and is fully
implemented in Primus Lisp and consists of two modules:

- taint-sources: is a module that automatically detects untrusted
user inputs and correspondingly introduces taint into the system;

- sensitive-sinks: monitors functions that are considered sensitive
to untrusted/unsanitized user inputs, and reports policy violations
if such functions are called with tainted arguments.

It is possible to specify arbitrary sanitization procedures (and in
general extend the policy with your own rules).

The code is thoroughly documented, so it should be more or less
obvious on how to use it.


ivg:
Note, the PR is fully implemented in Primus Lisp, so it is not checked or run by CI (unfortunately).

gitoleg:
looks like that evetryhing is just fine: these files successfully installed to my `~/share/primus/site-lisp`

PR #881, opened 2018-10-04
handles correctly format change in the cache module

gitoleg:
`bap` fails in case entry in cache index doesn't exist. This PR fixes it.

### DIFF
--- a/plugins/cache/cache_main.ml
+++ b/plugins/cache/cache_main.ml
@@ -80,7 +80,10 @@ module Index = struct
     then clean (evict_entry idx)
     else idx
 
-  let remove_entry e = Sys.remove e.path
+  let remove_entry e =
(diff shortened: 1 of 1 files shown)


plugins/cache/cache_main.ml
```diff
@@ -167,6 +189,31 @@ module Index = struct
...
+        index_to_file (cache_dir () / index_file) old;
+      | x ->
+        warning
+          "can't update index version from %d to %d" x index_version in
+    Sys.remove file
+
+  let get_index () =
+    FileUtil.ls (cache_dir ()) |> List.find ~f:is_index
+
+  let verify () =
```
ivg: the name is misleading, you're not verifying anything, you are actually upgrading it, if needed.

Issue #890, opened 2018-10-29
Failure to run a pass on a ARM EABI binary


ebtaleb:
The incriminating binary is a ARM7 EABI binary compiled with IAR, and can be found [there](https://www50.zippyshare.com/v/4BcFQeLv/file.html).
I am using BAP master with bap-signatures 1.4.0.

When running the command `bap ./myE.elf  --no-byteweight `, the following error is raised:

>Failed to create a project: (((Failure "Sexplib.Parser: failed to parse line 1 char 838402: sexp")
>  "Raised at file \"pervasives.ml\", line 32, characters 17-33\
>  \nCalled from file \"parsing.ml\", line 142, characters 39-75\
>  \nCalled from file \"parsing.ml\", line 164, characters 4-28\
>  \nRe-raised at file \"parsing.ml\", line 183, characters 8-17\
>  \nCalled from file \"lib/ogre/ogre.ml\", line 383, characters 8-64\
>  \nCalled from file \"src/or_error.ml\", line 62, characters 9-15\
> \n")
>  "Raised at file \"src/import0.ml\" (inlined), line 234, characters 22-32\
> \nCalled from file \"src/error.ml\" (inlined), line 9, characters 14-30\
> \nCalled from file \"src/or_error.ml\", line 70, characters 17-32\
> \nCalled from file \"lib/bap/bap_project.ml\", line 264, characters 47-54\
> \nCalled from file \"src/or_error.ml\", line 62, characters 9-15\
> \n")

[Here is the log](https://pastebin.com/vY4FtuXX) from `~/.local/state/bap/log `.

PR #891, opened 2018-10-29
escapes strings before generating OGRE document

gitoleg:
fixes https://github.com/BinaryAnalysisPlatform/bap/issues/890

Bap fails with a very nasty exception if the llvm backend produces an ogre document that can't be parsed due to an unexpected symbol sequence, e.g.  parsing of the
```
(symbol-entry "?<Constant "Disconnecting; %d\r\n">" 0x9258 0x14 0x4a2d8) 
```
failed just because of `;` symbol, which introduces a comment. 

Previously, we tended to quote strings only if there are some suspicious characters inside.
Starting from this PR we will quote any string in ogre document and replace any suspicious character with its code. And suspicious is everything but letters and digits:
```
(symbol-entry "\x3f\x3cConstant\x20\x22Disconnecting\x3b\x20\x25d\x5cr\x5cn\x22\x3e" 0x9258 0x14 0x4a2d8)
```

Note, these changes will not affect an ogre dump readability: 
```
(symbol-entry .text_45 46438 0 312806)
(symbol-entry .text_46 46484 0 312852)
(symbol-entry "?<Constant \"Disconnecting; %d\\r\\n\">" 37464 20 303832)
(symbol-entry "?<Constant \"mac.c\">" 37440 8 303808)
(symbol-entry "?<Constant \"macCtrl\">" 37448 8 303816)
(symbol-entry "?<Constant \"macMon\">" 37456 8 303824)
....
```

### DIFF
--- a/lib/bap_llvm/llvm_loader_utils.hpp
+++ b/lib/bap_llvm/llvm_loader_utils.hpp
@@ -12,19 +12,20 @@ namespace loader {
 
 using namespace llvm;
 
-bool str_mem(const std::string &s, char x) {
-    return (s.find(x) != std::string::npos);
+std::string escape(const std::string &src) {
+    std::stringstream dst;
+    for (auto it = src.begin(); it != src.end (); ++it) {
+        if (isalpha(*it) || isdigit(*it)) dst << *it;
+        else
+            dst << "\\x" << std::hex << int(*it) ;
+    }
+    return dst.str();
 }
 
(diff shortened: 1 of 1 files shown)


ivg:
The change is necessary but not sufficient. It is still possible that an OGRE document passed to BAP would not be well-formed, so on the ML side we shall handle all errors and instead of failing produce `Error`s.

PR #906, opened 2018-12-11
makes dead code elimination less conservative

### LINKED ISSUE #905: Problem with optimizations
So as requested in #904. It seems to doesn't work correctly, although I'm not sure is it unwanted behavior. It's optimizng some othere instructions like 000000da but cannot get rid of obvious double assignment like 00000116.

```
ubuntu@DESKTOP-UG9LBOD:/mnt/c/Users/root/Desktop$ bap file.out -d > file.bap_dump
ubuntu@DESKTOP-UG9LBOD:/mnt/c/Users/root/Desktop$ bap file.out --optimization-level=3 -d > file.bap_dump_opt
```
[files.zip](https://github.com/BinaryAnalysisPlatform/bap/files/2667656/files.zip)

ivg:
As was noted in #905 the optimization passes sometimes misses an
opportunity to remove obvious assignments. The underlying reason
is that we compute a very conservative set of variables which are
used for interprocedural communication. And if a variable is a
member of this set then we do not delete it (under assumption, that
it could be used in other subroutines).

What was overconservative is that all versions of the same variable
were protected, while only the last definition one could be used for
interprocedural communication. The propsed change relaxes the
protected set, and removes only the last version of a variable from
the set of the dead variables.

Fixes #905

### DIFF
--- a/plugins/optimization/optimization_main.ml
+++ b/plugins/optimization/optimization_main.ml
@@ -38,14 +38,23 @@ end
 let computed_def_use sub =
   def_use_collector#visit_sub sub (Var.Set.empty,Var.Set.empty)
 
-let compute_dead can_touch protected sub =
+let protect dead protected =
+  Set.to_sequence dead ~order:`Decreasing |>
+  Seq.fold ~init:(Var.Set.empty,protected)
+    ~f:(fun (dead,protectors) var ->
+        if Set.mem protectors (Var.base var)
+        then (dead,Set.remove protectors (Var.base var))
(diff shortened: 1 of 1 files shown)


plugins/optimization/optimization_main.ml
```diff
@@ -38,17 +38,30 @@ end
 let computed_def_use sub =
   def_use_collector#visit_sub sub (Var.Set.empty,Var.Set.empty)
 
+let protect dead protected =
```
gitoleg: dead code elimination requested)

PR #911, opened 2019-01-18
Fix ir graph's missing edges.

abhcs:
Earlier the IR graph was created from just the IR. Now we store the block successors with each block term as an attribute "succs". These successors are determined from the cfg. The attribute "succs" is used when creating the IR graph. This may not be the best way to fix the graph but it seems like simplest approach to me. The IR graph is fixed for the two test cases I tried.

This is not really a pull request, more like a review request (although you can accept this if it looks good to you).

### DIFF
--- a/lib/bap_types/bap_ir_graph.ml
+++ b/lib/bap_types/bap_ir_graph.ml
@@ -1,6 +1,7 @@
 open Core_kernel.Std
 open Regular.Std
 open Graphlib.Std
+open Bap_types.Std
 open Option.Monad_infix
 
 open Bap_bil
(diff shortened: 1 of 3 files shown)


lib/bap_types/bap_ir_graph.ml
```diff
@@ -143,13 +204,22 @@ module Node = struct
...
-              | Some tid -> Some {src; pos; dst}))
+              | Some _ -> Some {src; pos; dst}))
 
-  let outputs src t : edge seq =
+  let outputs_from_jmps src t : edge seq =
     Term.enum jmp_t src |> Seq.filter_mapi ~f:(fun pos jmp ->
         succ_of_jmp t.sub jmp >>= fun dst ->
         if mem src t && mem dst t then Some {src; pos; dst} else None)
 
+  let outputs src t : edge seq =
```
abhcs: This is the part of the commit that actually bothers me. The successors from the attribute "succs" is used only if no successors are found in the jmp terms. The position is just marked as 0. Although this works in the few cases I am concerned with, this is not a general solution.

abhcs:
Thanks for the review. I have not tried @gitoleg 's implementation yet. Can you give me a timeline when the implementation will enter the master branch? I will have to act accordingly.

gitoleg:
@abhcs  I think in nearest few days, say today or tomorrow

abhcs:
Thank you!

Issue #899, opened 2018-11-15
Incoming edges of switch blocks are missing in the cfg when using the ida brancher 


abhcs:
I had tested the changes in #861 against a large binary that I had, and found that I was able to disassemble all the switch blocks in that particular binary. However, recently I have needed to work with the cfg of one of the functions in the binary, and I found that the switch blocks had no predecessors, i.e. the incoming edges that were correctly reported by IDA (as verified by me in IDA) were lost somehow.

To reproduce the issue at a smaller scale, I created a similar file fib.c:
```
#include <stdint.h>

typedef struct {
  uint8_t magic1;
  uint8_t magic2;
} TEST;

#define MAGIC1 0x1
#define MAGIC2 0x2
#define MAGIC3 0x3
#define MAGIC4 0x4
#define MAGIC5 0x5
#define MAGIC6 0x6
#define MAGIC7 0x7
#define MAGIC8 0x8
#define MAGIC9 0x9

int fib (TEST *test, int n)
{
  int a = 0, b = 1, i = 0;
  for (; i < n; i++) {
    switch (test->magic2) {
    case MAGIC1:
      b = a + 1;
      a = a - 1;
      break;
    case MAGIC2:
      b = a + 2;
      a = a - 2;
      break;
    case MAGIC3:
      b = a + 3;
      a = a - 3;
      break;
    case MAGIC4:
      b = a + 4;
      a = a - 4;
      break;
    case MAGIC5:
      b = a + 5;
      a = a - 5;
      break;
    case MAGIC6:
      b = a + 6;
      a = a - 6;
      break;
    case MAGIC7:
      b = a + 7;
      a = a - 7;
      break;
    case MAGIC8:
      b = a + 8;
      a = a - 8;
      break;
    case MAGIC9:
      b = a + 9;
      a = a - 9;
      break;
    default:
      b = a + b;
      a = b;
    }
  }
  return b;
}

int main (void) {
  TEST t;
  int n = 42;
  fib (&t, n);
}
```
and compiled it with `gcc -o exe fib.c`. The switch blocks are disassembled correctly, but their incoming edges are missing from the cfg, which is generated using `bap exe -dcfg --brancher=ida --print-symbol=fib`. I have verified the cfg to be correct in IDA.

PR #914, opened 2019-01-24
preserves brancher information in the BIL code of an instruction

gitoleg:
# Introduction
During the program reconstruction we rely on several sources of information, such as lifter and brancher. However, the information that we obtained from the brancher is not stored that leads to a mismatch later. This PR preserves the information provided by the brancher in the BIL code of an instruction. Ideally, the brancher should operate on the BIL level and fix the BIL itself, but we have what we have. 

# Design

We have two sources of information, `D` the set of destinations from the lifter and `D'` the set of destinations form the brancher, which are equally credible for us, but may provide conflicting information. Our task is to generate the BIL code that will have a set of destinations `R` which preserves the existing destinations (completeness) and do not introduce any false destinations (soundness), or formally `R = D + D'`. This requirement is not very strong, as it doesn't quantify over branch predicates, so this transformation might introduce extra (and sometimes unnecessary) non-determinism. However, we're trying to limit it as much as possible without making the code to complicate, in particular, we're ensuring that if `D'` is a subset of `D` (i.e., there is no new information from the brancher) then we don't touch the code. 

# Implementation

If we have conflicting information between the brancher and the lifter, we resolve it using a conditional branch predicated with the `unknown` expression,

```
if (unknown) jmp d1;
if (unknown) jmp d2;
...
if (unknown) jmp dm;
jmp d;
```

If there is no conflict, for example, when lifter tells us to jump to an indirect destination, while brancher gives some concrete destinations `[d1,d2,...,dm]`, we spill the following code:
```  
  if (RAX = d1) jmp d1;
  if (RAX = d2) jmp d2;
  ...
  if (RAX = dm) jmp dm;
  jmp RAX
```
# References

Resolves: #899
Previous attempt: #911

### DIFF
--- a/lib/bap_disasm/bap_disasm_rec.ml
+++ b/lib/bap_disasm/bap_disasm_rec.ml
@@ -327,16 +327,18 @@ let add_destinations bil = function
   | [] -> bil
   | dests ->
      let dests = Addr.Set.of_list dests in
-     let dests_of_bil = dests_of_bil bil in
-     let jmps_are_differ = not (Set.is_subset dests ~of_:dests_of_bil) in
-     let all = Set.union dests dests_of_bil in
-     if jmps_are_differ then
-       (object inherit Stmt.mapper
-          method! map_jmp = function
-            | Int _    -> join_destinations all
-            | indirect -> make_switch indirect all
+     if has_jump bil then
+       let dests_of_bil = dests_of_bil bil in
+       let jmps_are_differ = not (Set.is_subset dests ~of_:dests_of_bil) in
+       let all = Set.union dests dests_of_bil in
+       if jmps_are_differ then
+         (object inherit Stmt.mapper
+            method! map_jmp = function
+              | Int _    -> join_destinations all
+              | indirect -> make_switch indirect all
           end)#run bil
-     else bil @ join_destinations all
+       else bil
+     else bil @ join_destinations dests
 
 let stage2 dis stage1 =
   let stage1 = filter_valid stage1 in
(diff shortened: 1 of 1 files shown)


lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -353,8 +393,16 @@ let stage2 dis stage1 =
             Dis.stop s (Dis.insns s)) |>
       List.map ~f:(function
           | mem, None -> mem,(None,None)
-          | mem, (Some ins as insn) -> match stage1.lift mem ins with
-            | Ok bil -> mem,(insn,Some bil)
+          | mem, (Some ins as insn) ->
+            match stage1.lift mem ins with
+            | Ok bil ->
+              let bil =
+                if has_jump bil then
```
ivg: it should be required for an instruction to have a jump, so we should remove this check and ensure that the BIL code always contains the sum of the destinations returned by the lifter and by the brancher.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -299,6 +299,46 @@ let create_indexes (dests : dests Addr.Table.t) =
 
 let filter_valid s = {s with inits = Set.inter s.inits s.valid}
 
+let join_destinations dests =
```
ivg: I think that we can stick to a more simple spraying destination in case of the nondeterminism and adopt the following scheme:
```
if (unk) jmp d1;
...
if (unk) jmp dm;
jmp d;
```

where `d` is the original BIL destination and `d1,dm` is the set of destinations provided by the brancher (alternatively you can just pick an arbitrary `d` from the whole set, it doesn't really matter).

This code will produce graphs that are simple (ideally, it should map to a single block with multiple `when` guarded jumps), while your version requires some sophistication during graph generation and with naive approach will spill a graph that is quadratic with respect to the number of destinations (which in case if we will retarget VSA here, could be a very big number).

PR #920, opened 2019-02-01
the i64 import error

### LINKED ISSUE #919: i64 import error
Cause `Filename.quote ida` in **plugins/ida/bap_ida_info.ml**, the `Filename.check_suffix ida "64"` won’t work. Therefore, it would not use the i64 for 64-bit bins.

If you added any new symbols or renamed existing in a 64-bit bin, it won't pass the information to BAP for the code error.

neoni:
fix #919


ivg:
@gitoleg, can you check that it doesn't break compatibility with 6.x and 7.x versions of IDA? Otherwise, LGTM. Thanks, @neoni!

Issue #901, opened 2018-11-20
`addr32 callq` instruction is dropped


tobast:
It seems that BAP does not handle [address size overrides](https://wiki.osdev.org/X86-64_Instruction_Encoding#Operand-size_and_address-size_override_prefix), eg. `67 e8 fd 4b 00 00` (which, according to `objdump`, translates to `addr32 callq 14c00`, `0x67` being the `addr32` prefix). Upon encountering such an instruction, the instruction is simply dropped, resulting in a disconnected control flow graph.

A program containing such an instruction can be found [here](https://tobast.fr/files/bap_addr32_test.bin), its source code being [this](https://paste.tobast.fr/?c5b9073ab0116a58#cfKv9VmbHz2/kDkPR4JtAj5sweJIEmfS3h7scslVQCw=).

ivg:
Note, this code occurs in C-runtime generated by modern compilers, e.g., this compiler is known to generate it:
```
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /build/gcc/src/gcc/configure --prefix=/usr --libdir=/usr/lib --libexecdir=/usr/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=https://bugs.archlinux.org/ --enable-languages=c,c++,ada,fortran,go,lto,objc,obj-c++ --enable-shared --enable-threads=posix --enable-libmpx --with-system-zlib --with-isl --enable-__cxa_atexit --disable-libunwind-exceptions --enable-clocale=gnu --disable-libstdcxx-pch --disable-libssp --enable-gnu-unique-object --enable-linker-build-id --enable-lto --enable-plugin --enable-install-libiberty --with-linker-hash-style=gnu --T --enable-multilib --disable-werror --enable-checking=release --enable-default-pie --enable-default-ssp --enable-cet=auto
Thread model: posix
gcc version 8.2.1 20180831 (GCC)
```

PR #922, opened 2019-02-15
few x86 enhancements

gitoleg:
this PR adds a couple of x86 instructions:

1) adds `endbr` instructions
   Despite that we implement these instructions as `noop`, they still are very important for cfg integrity, 
   since they usually appear in the beginning of functions.
 
2) adds call instructions with operand override prefix.  
    Previously, we were lifting `call` instructions only if no prefixes present. But `x86_disasm` does all the 
    job needed for prefix comprehension (at least in  part of operand overriding).  

Few examples:
```   
$ echo "f3 0f 1e fa" | bap-mc --show-bil --show-insn
ENDBR64()
{

}
$ echo "f3 0f 1e fb" | bap-mc --show-bil --show-insn
ENDBR32()
{

}
$ echo "67 e8 00 fe ff ff" | bap-mc --show-bil --show-insn --addr=0x400500
CALL64pcrel32(-0x200)
{
  RSP := RSP - 8
  mem := mem with [RSP, el]:u64 <- 0x400506
  jmp 0x400306
}
```
fix https://github.com/BinaryAnalysisPlatform/bap/issues/901

### DIFF
--- a/plugins/x86/x86_endbr.ml
+++ b/plugins/x86/x86_endbr.ml
@@ -1,0 +1,15 @@
+open Core_kernel
+open Bap.Std
+open X86_asm.Reg
+
+module IA32  = X86_backend.IA32
+module AMD64 = X86_backend.AMD64
+
+type endbr = [ `ENDBR32 | `ENDBR64 ] [@@deriving bin_io, sexp, compare, enumerate]
+
+let lift _mem _insn = Ok [ ]
+
+let () =
+  let name op = sexp_of_endbr op |> Sexp.to_string in
+  List.iter all_of_endbr ~f:(fun op -> IA32.register (name op) lift);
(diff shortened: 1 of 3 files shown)


plugins/x86/x86_endbr.ml
```diff
@@ -0,0 +1,15 @@
+open Core_kernel
+open Bap.Std
+open X86_asm.Reg
+
+module IA32  = X86_backend.IA32
+module AMD64 = X86_backend.AMD64
+
+type endbr = [ `ENDBR32 | `ENDBR64 ] [@@deriving bin_io, sexp, compare, enumerate]
```
ivg: Shouldn't we use special here? :/

ivg:
you can merge as soon as Travis approves :)

PR #923, opened 2019-02-18
SSE XMM0 ABI 

turnersr:
The following code allows ymm registers to be cast to smaller register. This new casting operation is used to extract xmm registers from ymm registers in order to support monomorphic functions in sse. The SysV ABI is extended to receive more arguments and can now recognize functions that read and write to xmm0.

### DIFF
--- a/plugins/x86/x86_abi.ml
+++ b/plugins/x86/x86_abi.ml
@@ -16,7 +16,7 @@ module type abi = sig
   val arch : Arch.x86
   val name : string
   val size : C.Size.base
-  val arg  : pos -> exp
+  val arg  : C.Type.t -> pos -> exp
   val demangle : string -> string
   val autodetect : project -> bool
 end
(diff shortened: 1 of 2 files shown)


plugins/x86/x86_abi.ml
```diff
@@ -29,7 +29,16 @@ module SysV = struct
...
   let stack n = Stack.create arch n
-  let arg = function
+
+
+  let xmm r =  Bil.(cast low 64 (var ymms.(r)))
+
+  let flt = function
+    | Ret_0 -> xmm 0
+    | Ret_1 -> xmm 1
+    | Arg n -> xmm n
```
ivg: This function will blow up with an exception if we will try to pass more than 16 floats, and will be just incorrect if we will try to pass more than 8,  as according to SysV AMD64 ABI only xmms0-xmms7 are used for arguments passing, while the rest XMM8-XMM15 are used as scratch registers. So it should be something like this:
```ocaml
| Arg n -> if n < 8 then xmm n else stack Int.(n-8)
```

Note, this is still an under-approximation, as it will only work for monomorphic functions, but better than terminating analysis with an exception :)

plugins/x86/x86_abi.ml
```diff
@@ -40,6 +49,10 @@ module SysV = struct
     | Arg 5 -> var r.(1)
     | Arg n -> stack Int.(n-6)
 
+  let arg = function
+    | `Basic {C.Type.Spec.t=`double} -> flt
```
ivg: if I'm reading SysV AMD64 correctly, then `float` is passed in the same way as `double`. So, if we will extend our `flt` with an extra parameter for `width`, instead of hardcoding `64`, we can easily handle `float` as well.

ivg: forgot to add the link to the standard, here it is, see pages 17-22
https://www.uclibc.org/docs/psABI-x86_64.pdf

PR #932, opened 2019-03-12
enables memory sharing between instructions

gitoleg:
Given that `byteweight` can emit roots that point somewhere to the middle of a real instruction, we can end up with two possible scenarios:
1) Bytes from the pointed address can't be disassembled. It leads to malformed blocks, and that's the issue we already fixed by filtering only disassembleable code.
2) Bytes can be disassembled, and it's a more interesting and tough case.
   
Example:
```
8dd820:       45 0f be 07             movsbl (%r15),%r8d
8dd824:       83 c3 01                add    $0x1,%ebx
```
Byteweight gives us a root at `0x8dd821`, and instruction `0f be 07` is perfectly disassembled too. And in general, we can't decide which one is correct.
Our solution is to process both of the instructions: `0x8dd820` and `0x8dd821`, and create 3 blocks here: the first one is a block that terminates at instruction with bounds `0x8dd820 - 0x8dd823`, the second one is a block that contains the only instruction `0x8dd821 - 0x8dd823`, and finally a block at `0x8dd824` with fall-through edges from the previous two. Note, that we preserve an invariant that there is only one block at some address.

### Details of implementation:

We allow instructions to share bytes, with the only exception that a first byte, i.e. the address of an instruction must be unique and unambiguously define an instruction. That's why the last byte can't be a key in all previously used mappings.  Also, we can't iterate over the memory by incrementing addresses, as we did it before to find block bounds. That's why we use instruction bounds and addresses of start and end of any block correspond to a first and to the last bytes of some instructions.

We got rid of `Span` module. And now we remember all memory chunks where a successfully disassembled instruction occurred. For the same reason we don't need to check our input roots anymore: any call of `Set.mem inits` will be done for an address which can be only the first byte of an instruction and nothing else.

We consider all instructions that share max address as a barrier, and we unconditionally update destinations for them to get a behavior we described in the example above.

### DIFF
--- a/lib/bap_disasm/bap_disasm_rec.ml
+++ b/lib/bap_disasm/bap_disasm_rec.ml
@@ -148,7 +148,7 @@ let need_barrier s ((addr,max_addr) as insn) =
 let find_intersections s =
   let update_barriers xs init =
     List.fold xs ~init ~f:(fun bs x ->
-        if need_barrier s x then fst x :: bs else bs) in
+        if need_barrier s x then Set.add bs (fst x) else bs) in
   let find_intersected (addr, max_addr) =
     collect ~dir:Visited.next s addr
       ~stop_on:(fun (a,_) -> Addr.(a > max_addr)) in
@@ -165,11 +165,11 @@ let find_intersections s =
       let barriers = update_barriers insns barriers in
       let barriers =
         if List.exists insns ~f:(is_shared_max insn)
-        then addr :: barriers
+        then Set.add barriers addr
         else barriers in
       let data = List.fold insns ~init:data ~f:add in
       loop barriers data (next max_addr) in
-  loop [] Addr.Map.empty (Visited.min s.visited)
+  loop Addr.Set.empty Addr.Map.empty (Visited.min s.visited)
 
 let rec has_jump = function
   | [] -> false
@@ -216,21 +216,23 @@ let stop_on = [`Valid]
 (* find intersections, i.e. instructions which share bytes, but have different address,
    and update destinations for those of them, which also share max address to
    split a block in such place *)
-let update_intersections disasm brancher mem s =
-  let roots,intersected = find_intersections s in
-  let init = {s with intersected; roots} in
+let update_intersections disasm brancher s =
   let next dis s = match s.roots with
     | [] -> Dis.stop dis s
     | addr :: roots ->
-      let mem = Memory.view ~from:addr s.base in
-      let mem = Result.map_error mem ~f:(fun err -> Error.tag err "next_root") in
-      mem >>= fun mem ->
+      Memory.view ~from:addr s.base >>= fun mem ->
       Dis.jump dis mem {s with roots; addr} in
-  Dis.run disasm mem ~stop_on ~return ~init
-    ~hit:(fun d mem insn s ->
-        let () = update_dests s (brancher mem insn) mem in
-        next d s)
-    ~stopped:next
+  let roots,intersected = find_intersections s in
+  let s = {s with intersected; roots=Set.to_list roots} in
+  match Set.min_elt roots with
+  | None -> Ok s
+  | Some addr ->
+    Memory.view ~from:addr s.base >>= fun mem ->
+    Dis.run disasm mem ~stop_on ~return ~init:s
+      ~hit:(fun d mem insn s ->
+          let () = update_dests s (brancher mem insn) mem in
+          next d s)
+      ~stopped:next
 
 let stage1 ?(rooter=Rooter.empty) lift brancher disasm base =
   let roots =
@@ -247,7 +249,7 @@ let stage1 ?(rooter=Rooter.empty) lift brancher disasm base =
     ~hit:(fun d mem insn s ->
         next d (update s mem insn (brancher mem insn)))
     ~invalid:(fun d mem s -> next d (errored s (`Failed_to_disasm mem)))
-    ~stopped:next >>= update_intersections disasm brancher mem
+    ~stopped:next >>= update_intersections disasm brancher
 
 (* performs the initial markup.
 
--- a/lib/bap_disasm/bap_disasm_rec.ml
+++ b/lib/bap_disasm/bap_disasm_rec.ml
@@ -114,52 +114,17 @@ type t = stage3
 
 let errored s ty = {s with errors = (s.addr,ty) :: s.errors}
 
-let update_dests base dests mem ds =
-  let ds = List.map ds ~f:(fun d -> match d with
+let filter_dests base ds =
+  List.map ds ~f:(fun d -> match d with
       | Some addr,_ when Memory.contains base addr -> d
-      | _,kind -> None, kind) in
+      | _,kind -> None, kind)
+
+let update_dests s mem ds =
   let key = Memory.min_addr mem in
-  match ds with
-  | [] -> Addr.Table.add_exn dests ~key ~data:[]
+  match filter_dests s.base ds with
+  | [] -> Addr.Table.add_exn s.dests ~key ~data:[]
   | ds -> List.iter ds ~f:(fun data ->
-      Addr.Table.add_multi dests ~key ~data)
-
-let collect ~dir ~stop_on ?cond s start =
-  let add = match cond with
-    | None -> fun x xs -> x :: xs
-    | Some cond -> fun x xs -> if cond x then x :: xs else xs in
-  let rec loop addr data =
-    match dir s.visited addr with
-    | None -> data
-    | Some x when stop_on x -> data
-    | Some x -> loop (fst x) (add x data) in
-  loop start []
-
-let exists_visited ~dir ~stop_on ~cond s start  =
-  not (List.is_empty @@ collect ~dir ~stop_on ~cond s start)
-
-let is_shared_max (_,a) (_,a') = Addr.equal a a'
-
-let find_intersections s =
-  let find_intersected (addr, max_addr) =
-    collect ~dir:Visited.next s addr
-      ~stop_on:(fun (a,_) -> Addr.(a > max_addr)) in
-  let next max_addr =
-    let next_start = Addr.succ max_addr in
-    match Visited.find s.visited next_start with
-    | None -> Visited.next s.visited max_addr
-    | Some max' -> Some (next_start, max') in
-  let add data (addr,max) = Map.add_multi data max addr in
-  let rec loop data = function
-    | None -> data
-    | Some ((_, max_addr) as insn) ->
-      let insns = find_intersected insn in
-      let data = match insns with
-        | [] -> data
-        | insns ->
-           List.fold insns ~init:(add data insn) ~f:add in
-      loop data (next max_addr) in
-  loop Addr.Map.empty (Visited.min s.visited)
+      Addr.Table.add_multi s.dests ~key ~data)
 
 let rec has_jump = function
   | [] -> false
@@ -179,7 +144,7 @@ let is_barrier s mem insn =
 let update s mem insn dests : stage1 =
   let s = { s with visited = Visited.add s.visited mem } in
   if is_barrier s mem insn then
-    let () = update_dests s.base s.dests mem dests in
+    let () = update_dests s mem dests in
     let roots = List.(filter_map ~f:fst dests |> rev_append s.roots) in
     { s with roots }
   else {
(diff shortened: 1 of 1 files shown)


ivg:
> We consider all instructions that share max address as a barrier

What is the formal basis for this?

gitoleg:
If instructions share last bytes, then they also can share destinations, e.g. fall edges like in example in comment above. i.e. it could be a case when two fall edges point to the same next instruction. so to make this instruction reachable from both places we need to make it a start of a new block. and terminates previous two.

ivg:
> If instructions share last bytes, then they also can share destinations

E.g., suppose we have a sequence of bytes `xyyy` where `xyyy` and `yyy` are valid instructions, and we have two roots, pointing correspondingly to `xyyy` and `yyy`. where `yyy` is treated as `incr R0` and `xyyy` is treated as `jmp R0`. They do not share edges, and `yyy` should be a fall-through, while `xyyy` is a barrier.

ivg:
In other words, this assumption is wrong. Here is my reasoning. A leader of a basic block is either a root or an instruction that has more than one incoming edge. What you're saying is that if two instructions that belong to two different basic blocks share the same bytes, then they _may_ also share the same destinations, therefore we have a block which have more than one incoming edge and we should generate a terminator at this point. What's wrong in your reasoning, is you assume that both instructions a fall-through instructions, hence they share the _same_ destination. However, they could be absolute jumps, which do not share edges, or it could be a jump and a fall-through, etc. So it is perfectly possible for them to have non-intersecting sets of destinations. 

If we will go back to the crux of the problem, then the following lemma is easily proved. 
Lemma 1.
If `q` is a fallthrough instruction, and there exists an instruction `p <> q` with the same address of the last byte, which is also a fallthrough instruction, then `q` is a barrier. 

Proof.
Since `p` and `q` are different instructions and both are fallthrough, then we have `max(p)+1=max(q)+1=fallthrough`, therefore `fallthrough` has two incoming edges from different blocks. 

Lemma 2. `p` is also a barrier.
Proof. By reflexivity.

The Lemma 2 actually implies, that your implementation have to update all candidates. Make sure you have it. 

Note, this lemmas are specific, and in general you can go without them. As the disassembler still must adhere to the definition of a basic block. I'm not sure why you need this assumptions at all. Probably to find a synchronization point and stop re-disassembling the same subgrahps?

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -66,121 +63,38 @@ module Cfg = Graphlib.Make(Node)(Edge)
...
-  type t [@@deriving sexp_of]
-
-  val empty : t
-  val add : t -> range -> t
-  val mem : t -> addr -> bool
-  val upper_bound : t -> addr -> addr option
-  val min : t -> addr option
-  val max : t -> addr option
-  val pp : Format.formatter -> t -> unit
+module Visited = struct
```
ivg: Why is visited an addr->addr mapping, not just a set? What is the payload? 
Why you're not using `Bap.Std.Memmap` or `Bap.Std.Table`?

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -195,23 +109,73 @@ type stage3 = {
...
+    | Some x when stop_on x -> data
+    | Some x -> loop (fst x) (add x data) in
+  loop start []
+
+let exists_visited ~dir ~stop_on ~cond s start  =
+  not (List.is_empty @@ collect ~dir ~stop_on ~cond s start)
+
+let is_shared_max (_,a) (_,a') = Addr.equal a a'
+
+let need_barrier s ((addr,max_addr) as insn) =
```
ivg: it is not a barrier, but a _terminator_ (barrier is basically an instruction that is followed by data). 

could you please explain all the clauses? 

Also, it is totally inefficient to construct the list,  just to prove that it is not empty. Please, get rid of it.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -241,34 +202,54 @@ let next dis s =
...
+      mem >>= fun mem ->
+      Dis.jump dis mem {s with roots; addr} in
+  loop s
 
 let stop_on = [`Valid]
 
+(* find intersections, i.e. instructions which share bytes, but have different address,
+   and update destinations for those of them, which also share max address to
+   split a block in such place *)
+let update_intersections disasm brancher s =
```
ivg: As we discussed this is not sound, please rewrite it.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -241,34 +202,54 @@ let next dis s =
...
+  let next dis s = match s.roots with
+    | [] -> Dis.stop dis s
+    | addr :: roots ->
+      Memory.view ~from:addr s.base >>= fun mem ->
+      Dis.jump dis mem {s with roots; addr} in
+  let roots,intersected = find_intersections s in
+  let s = {s with intersected; roots=Set.to_list roots} in
```
ivg: why `Set.to_list`? Is there any particular reason to change the representation? If not, then please keep it as Set.t

ivg:
> And now we remember all memory chunks where a successfully disassembled instruction occurred. For the same reason we don't need to check our input roots anymore: any call of Set.mem inits will be done for an address which can be only the first byte of an instruction and nothing else.

One of the reasons why we checked roots, was to identify starts of basic blocks (and correspondingly terminators). We have to treat every root as a basic block, even if we don't see any incoming edges, or see that there is only one. So I'm not sure whether your reasoning includes this. Please, clarify it.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -213,26 +203,75 @@ let next dis s =
...
+module Dest = struct
+  let rec compare_edge e e' =
+    if e = e' then 0
+    else
+      match e, e' with
+      | `Fall, _ -> -1
+      | `Jump, _ -> 1
+      | _ -> - (compare_edge e' e)
+
+  let compare (d,e) (d',e') =
```
ivg: can you, please, explain the reasoning behind this ordering? 

Both here, and leave the comment in the code.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -213,26 +203,75 @@ let next dis s =
...
+    match d, d' with
+    | None,None -> compare_edge e e'
+    | Some _, None -> 1
+    | None, Some _ -> -1
+    | Some d, Some d' ->
+       let x = Word.compare d d' in
+       if x = 0 then compare_edge e e'
+       else x
+
+  module Set = Set.Make(struct
```
ivg: Do not create Set modules in the new code. Derive comparable, and use polymorphic sets, e.g.,

```ocaml
module Dest = struct 
   type t = addr option * edge [@@deriving compare, sexp]
   include Base.Comparable.Make(struct type nonrec t = t end)
end
...
let dests = Set.empty (module Dests)
```

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -213,26 +203,75 @@ let next dis s =
...
+      type t = Brancher.dest [@@deriving sexp]
+      let compare = compare
+    end)
+
+end
+
+
+let equal_dest (d,e) (d',e') =
```
ivg: You already have `compare_dest` function, so `equal_dest` should be `compare_dest x y = 0`, or just use the derived equality from the `Base.Comparable` interface. 

If your `equal x y` is not defined as `compare x y = 0` mathematically, then this is just a wrong way of doing things.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -213,26 +203,75 @@ let next dis s =
...
+  match d, d' with
+  | None, None -> e = e'
+  | Some _, None | None, Some _ -> false
+  | Some d, Some d' -> Word.(d = d') && e = e'
+
 let update_intersections disasm brancher s =
-  let next dis s = match s.roots with
-    | [] -> Dis.stop dis s
+  let next dis ((roots,dests) as state) = match roots with
```
ivg: could please explain what is this function doing, and, most importantly, what are _intersections_ in your parlance?

lib/bap_disasm/bap_disasm_rec.ml
gitoleg: Set is not sufficient, because we need to know both first and last addresses of an instruction to figure out if it's a block terminator or not. `Bap.Std.Table` doesn't fit for this particular task because we have intersected memory regions. We can also use `Bap.Std.Memmap` in most of the cases, but it won't be so convenient as just the `addr->addr` mapping since we use addresses everywhere as a key for tables. And sometimes it will be even hard to use `mem` as a key when we need to infer the next key from a given one, which is easy to do in case of address by incrementing it, and not so easy for memory chunk.

lib/bap_disasm/bap_disasm_rec.ml
gitoleg: Set is not sufficient, because we need to know both first and last addresses of an instruction to figure out if it's a block terminator or not. `Bap.Std.Table` doesn't fit for this particular task because we have intersected memory regions. We can also use `Bap.Std.Memmap` in most of the cases, but it won't be so convenient as just the `addr->addr` mapping since we use addresses everywhere as a key for tables. And sometimes it will be even hard to use `mem` as a key when we need to infer the next key from a given one, which is easy to do in case of address by incrementing it, and not so easy for memory chunk.

lib/bap_disasm/bap_disasm_rec.ml
gitoleg: When I'm taking about intersections I mean instructions, which have intersected memory regions
And this function finds such instructions and updates the destinations table for those of them that have the same last byte and at least one common destination like it's described in details further. 
We save all the intersected instructions as a mapping `addr -> addr list`, where a key is the last address of instruction and data is a list of instructions addresses that share it. And if there is more then one instruction in a list then we need to find the destinations for each of them and update the destination table for those of the instructions which have at least one common destination.

gitoleg:
> One of the reasons why we checked roots, was to identify starts of basic blocks (and correspondingly terminators). We have to treat every root as the start of a basic block, even if we don't see any incoming edges, or see that there is only one. So I'm not sure whether your reasoning includes this. Please, clarify.

I think you confuse here our reconstructor and disassembler because we're not trying to figure out
anything by counting incoming edges here. We were checking the roots to get rid of such of them, which cause an error of a disassembler when it tried to process them. And it led to unexpected ends of blocks and undisassembled code. Now we don't need to do it, as we aren't trying to check every single address of the whole memory as we did it before, but only the known addresses, verified by the disassembler itself.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -241,34 +157,88 @@ let next dis s =
     | [] -> Dis.stop dis s
     | r :: roots when not(Memory.contains s.base r) ->
       loop {s with roots}
-    | r :: roots when Span.mem s.visited r ->
+    | r :: roots when Visited.is_insn s.visited r ->
```
ivg: Since now you visited set only records valid starts of instructions, not the set of addresses that were attempted by a disassembler, we may not now guarantee, that the recursive disassembler will terminate. A particular example (valid for the arm diassembler which prefers to skip through bytes, is the following):

inputs: `data=01234, root=[0]`
disassembler-output: `insn=1234, next=0`

Here we will mark `1234` as a valid instruction and visited, and continue with the same `root=[0]`, hence the infinite loop.

So we need to either check every time when we add a new root that it wasn't attempted, or check the same property every time we extract a root.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -223,12 +142,9 @@ let is_barrier s mem insn =
...
 
 let update s mem insn dests : stage1 =
-  let s = {s with valid = Set.add s.valid (Memory.min_addr mem)} in
+  let s = { s with visited = Visited.add s.visited mem } in
   if is_barrier s mem insn then
-    let dests = List.map dests ~f:(fun d -> match d with
-        | Some addr,kind when Memory.contains s.base addr -> d
-        | _,kind -> None, kind) in
-    update_dests s dests mem;
+    let () = update_dests s mem dests in
```
ivg: It also looks like that the implementation of the `is_barrier` function diverged from its name. Originally, we were classifying functions by using the is_barrier predicate provided by the disassembler, so that we were either pushing the fallthrough address or the address of an instruction. Many eons later, we basically started to classify them by whether they employ normal control flow with the fallthrough edge only, or it is a jump instruction, which may or may not have the fallthrough. 

Long story short, let's rename the `is_barrier` function to `is_jump`

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -241,34 +157,88 @@ let next dis s =
...
-  loop {s with visited}
+      let mem = Memory.view ~from:addr s.base in
+      let mem = Result.map_error mem ~f:(fun err -> Error.tag err "next_root") in
+      mem >>= fun mem ->
+      Dis.jump dis mem {s with roots; addr} in
+  loop s
 
 let stop_on = [`Valid]
 
+let find_intersections s =
```
ivg: This algorithm looks suboptimal and it computes more than returns, so it is a candidate for optimization. 
The algorithm is basically computing two groups of instructions:
1) those that share tails with any other instructions (represented with values that have more than one element), we need it because those instructions are barriers, and we need them in some efficient representation. This set is easily computed. 

2) a set of all instructions which intersect with some other instructions, but do not share tails (represented as values with singleton set). We need to in the later stage to ensure that all discovered instructions ended up in the CFG, and restart reconstruction on each. This set is hard to compute, however we can have it for free on stage 2, where we walk through the visited set and add instruction to CFG, at this step we should also remove the added instruction from the visited set. Therefore at the end we will have the sought  leftovers.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -336,72 +310,88 @@ let add_destinations bil = function
...
+  let fetch_blocks () =
+    let rec loop leftovers start =
+      match next_block leftovers start with
+      | Some (leftovers,start,(addr, max_addr)) ->
+         if Hashtbl.mem addrs start then
+           loop leftovers (next max_addr)
+         else
+           create_block start addr max_addr >>= fun () ->
+           loop leftovers (next max_addr)
+      | None when Map.is_empty leftovers -> Ok ()
```
ivg: it looks like that this case is not necessary, it will be handled by the case on line 389

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -66,116 +63,56 @@ module Cfg = Graphlib.Make(Node)(Edge)
...
-  let rec min = function
-    | Empty -> None
-    | Node (Empty,(a0,_),_,_) -> Some a0
-    | Node (lhs,_,_,_) -> min lhs
-
-  let max = function
-    | Empty -> None
-    | Node (_,_,m,_) -> Some m
-end
+  let visit t addr =
```
ivg: the name `touch` would be better, but this function has weird semantics. It behaves in the DWIM manner, and eventually, when `I` or `M` will change, it will bring lots of surprises. The right solution is to be as explicit as possible, and have two functions, named `add_insn` and `add_unknown` or parameterize `add` with the `entry` type. And the call explicitly in all handlers.

lib/bap_disasm/bap_disasm_rec.ml
```diff
@@ -241,34 +180,74 @@ let next dis s =
...
+      let mem = Result.map_error mem ~f:(fun err -> Error.tag err "next_root") in
+      mem >>= fun mem ->
+      Dis.jump dis mem {s with roots; addr} in
+  let s = {s with visited = Visited.visit s.visited s.addr } in
+  loop s
 
 let stop_on = [`Valid]
 
+let has_common_destination dests addrs a =
+  let find_destinations a = match Addrs.find dests a with
```
ivg: This is unnecessary inefficient:
1) you're creating sets every time from a list, when you can store sets in the beginning
2) you're creating sets just to make an intersection, and the use `is_empty` and drop the result. This will create a lot of pressure on the garbage collector. Consider doing `Set.exists xs ~f:(Set.mem ys)` to find if there exists an intersection between two sets.

Issue #691, opened 2017-08-17
Fix configuration files and merge 512


ivg:
Currently, the configuration file is broken, as we always looks at `<etc>/bap/config` and do not prefix parameters with the plugin name. It might be fixed in #512. We should revisit this PR and merge it, or abandon.

ivg:
forget about #512, just fix the configuration path.

PR #935, opened 2019-03-22
includes plugin name into the configuration path

gitoleg:
fixes https://github.com/BinaryAnalysisPlatform/bap/issues/691

`confdir` was already prefixed above as needed, but not used


ivg:
How did you test this? Could you please create a testcase in our testsuite that checks that parameters specified via the configuration file indeed work, and work as expected (i.e., are overridden by the command line arguments and so on)

Issue #722, opened 2017-10-05
Revise the logging infrastructure


ivg:
Our current system provides one log per run, and provides the log rotation mechanism, that will oldify previous logs before starting a new one, thus providing a history of the last 100 invocations.

The major limitation of this scheme is that it doesn't work fine with the multiple runs of bap in parallel. Once one bap instance starts it creates a new file named log and starts writing into it. If before the first instance finishes another is started, a new file will be created and the old one will be renamed to `log~1` (and `old~1` will be renamed to `old~2` and so on until the oldest file `old~99` will be reached, which will be just deleted). So at the end, the `log` file will correspond to an instance of bap that started last. That's not too bad, if we will ignore the fact that it is hard to figure out which log corresponds to which instance. However, on a high congestion system, it may happen that several instances of bap will enter the log-rotate sequence, that is in fact non-reenterant and is not protected. The ramifications are: (1) a log may be just unlinked - lost log or (2) two instances may open the same file - log corrupted. Basically, this is the classic TOCTOU class of bug. 

The latter two issues can be fixed by introducing some kind of locking mechanism (as it is done in the caching subsystem). However, it will not solve the usability issue that will make it hard to guess which log belongs to which instance. 

Basically, we have the following options:
1. Keep everything as it is (just fix the TOCTOU bug), possibly logging each job into its own directory.
2. Push data from all instances to the same log, prefixing each entry with the pid number. Enable log rotation based on the log size or make it daily
3. Create a unique log file for each instance, probably mixing in the PID number, and implement some sort of GC, that will remove too old logs. 
4. Something else?

## Fix it and keep it

Basically, the scheme works fine with multiple instances of BAP if each instance is using its own folder. This is how we use it with IDA Pro. If there are several chains of calls, then each chain may have its own logging folder, so the sequence of logs will be useful. We still need to fix the TOCTOU bug, as we may not expect a user to adhere to this scheme. We also need to provide a command line interface for logger directory location, as right now we are relying on the `BAP_LOG` variable. 

This is the solution of the least effort.

## Big mess of messages

This is a standard solution in the style of syslogger. All instances are writing to the same file, prefixing messages with the PID of a process. This approach has two main drawbacks:

1. It doesn't scale well. The log file becomes a congestion point very soon, since we will need to sync multiple instances of BAP on every event. The problem can be delayed by buffering, but it will be still a problem.
[...]

PR #937, opened 2019-03-28
fixes TOCTOU bug in bap log

gitoleg:
fixes https://github.com/BinaryAnalysisPlatform/bap/issues/722

### DIFF
--- a/lib/bap/bap_log.ml
+++ b/lib/bap/bap_log.ml
@@ -54,9 +54,9 @@ let print_message ppf msg =
   | Error -> eprintf "%s@\n%!" msg.message
   | _ -> ()
 
-let lock log_folder =
-  let lock = log_folder / "lock" in
-  let lock = Unix.openfile lock Unix.[O_RDWR; O_CREAT] 0o666 in
+let lock () =
+  let file = Filename.get_temp_dir_name () / "bap_lock" in
+  let lock = Unix.openfile file Unix.[O_RDWR; O_CREAT] 0o666 in
   Unix.lockf lock Unix.F_LOCK 0;
   lock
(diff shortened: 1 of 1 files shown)


lib/bap/bap_log.ml
```diff
@@ -54,13 +54,26 @@ let print_message ppf msg =
...
+  Unix.close lock
+
 let open_log_channel user_dir =
   try
     let log_folder = log_folder user_dir in
     mkdir log_folder;
     let file = log_folder / "log" in
-    if Sys.file_exists file
-    then rotate max_logs file;
+    let lock = lock log_folder in
```
ivg: In fact, it looks like that the `mkdir log_folder` expression is also experiencing the same TOCTOU issue, so could you please protect everything from the line 70.

lib/bap/bap_log.ml
```diff
@@ -54,13 +54,26 @@ let print_message ppf msg =
   | Error -> eprintf "%s@\n%!" msg.message
   | _ -> ()
 
+let lock () =
+  let file = Filename.get_temp_dir_name () / "bap_lock" in
```
ivg: 1) `bap.lock` is a much more conventional name 
2) this lock file would be global to all users, so my suggestion is to encode in the name of the lock file the path to the folder, that you're trying to create, salted with bap-prefix, the simplest idea is to take the digest of the path, e.g, 
```ocaml
let lock_filename logdir = 
    let digest = Md5.digest_string logdir in
    let name = sprintf "bap-%s.lock" (Md5.to_hex digest) in
    Filename.get_temp_dir_name () / name
```

Issue #740, opened 2017-12-21
the warn-unused-plugin is broken in 1.3


ivg:
Apparently the 817eac81 commit changed the semantics of taint introduction, so now, a place where taint is introduced also became a place where a taint is consumed. And although it makes some sense, it breaks WUR as the latter would consider all results as used at the callsite points. 

```
echo > main.c << EOF
#include <stdlib.h>
int main(int argc, char** argv) {
  char* buf = malloc(10);
  realloc(buf, 12);
  return 0;
}
EOF
gcc -O0 main.c -o main
$ bap ./main --warn-unused --print-bir-attr={visited,tainted-reg,tainted-regs} -d | grep 'call @realloc' -A8
00000128: call @realloc with return %00000129

.visited ()
00000129: 
.tainted-regs {RAX => [00000120, 00000128]}
.visited ()
.tainted-reg 00000128
0000023f: realloc_result := RAX
.tainted-regs {RDI => [00000128]}
```

Notice, that the synthetic `23f` term is now self proving, it both introduces and sanitizes the taint.

gitoleg:
fixed in https://github.com/BinaryAnalysisPlatform/bap/pull/939

PR #939, opened 2019-03-29
fixes the order of arguments in the callsites plugin

gitoleg:
Previously, the callsites plugin was inserting arguments in a such way that dual intent arguments were evaluated after the return arguments. This PR fixes the order, so that the in/our arguments are evaluated before the out arguments. It is also takes extra care and sorts the arguments before inserting, so that we do not rely too much on the order of arguments in the subroutine term.

This PR fixes the #740 in which a wrong order of evaluation led to the destruction of a def-use chain.

### DIFF
--- a/plugins/callsites/callsites_main.ml
+++ b/plugins/callsites/callsites_main.ml
@@ -54,8 +54,8 @@ let target intent sub blk call =
     | _ -> None
   else Some blk
 
-(* Note, that we intentionally invert compare for Both and Out
-   arguments, since they will be inserted in reverted order later. *)
+(* Note, that output arguments will be inserted in the reverse order, so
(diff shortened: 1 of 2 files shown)


plugins/callsites/callsites_main.ml
```diff
@@ -54,14 +54,29 @@ let target intent sub blk call =
     | _ -> None
   else Some blk
 
+(* Note, that we intentionally invert compare for Both and Out
+   arguments, since they will be inserted in reverted order later. *)
+let enum_args t =
+  let compare x y = match Arg.intent x, Arg.intent y with
```
ivg: looks like this order coincides with the `intent_order`, so it is easier and better just to use it.

PR #952, opened 2019-05-24
adds an apline based dockerfile

gitoleg:
This PR adds a docker file which builds the testing version of `bap` and is based on
`ocaml/opam2:alpine` image.

### DIFF
--- a/docker/alpine/Dockerfile
+++ b/docker/alpine/Dockerfile
@@ -3,8 +3,7 @@ FROM ocaml/opam2:alpine
 ENV HOME /home/opam
 WORKDIR $HOME
 
-RUN sudo apk update \
- && opam repo add bap git://github.com/BinaryAnalysisPlatform/opam-repository#testing \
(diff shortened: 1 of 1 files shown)


docker/alpine/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM ocaml/opam2:alpine as bap-testing
```
ivg: What's the purpose of the FROM stanza here?

docker/alpine/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+ENV HOME /home/opam
+WORKDIR $HOME
+
+RUN sudo apk update \
+ && sudo apk add \
```
ivg: Why do we need to to install them manually? Depext should do it for us. If not, then please update the opam-repository files, so that depext will handle it.

docker/alpine/Dockerfile
ivg: I mean, why do you need `as bap-testing` in the `FROM` stanza

Issue #961, opened 2019-07-30
BIL Lifting division with negative integers


icemonster:
Lifting a simple program like
```
int main(int argc, char const *argv[])
{
	int c = 10;
	int m = -3;
	int res = c / m;
	printf("%d\n", res);
	return 0;
}
```
Seems to yield incorrect BIL instructions for the division part. Specifically, you do this check right after computing a signed division:
``` If(EQ(HIGH(0x20, Var("divXXX", Imm(0x40))), Int(0x0, 0x20)), (), CpuExn(0x0)) ``` which is, I think, incorrect for negative numbers. This seems to be checking if the signed result (quotient) is too large for the destination but we always use numbers in two complement right? Values like Var("divXXX", Imm(0x40)) are represented as bit-vectors of size 64bits, so 0xFFFFFFFFFFFFFFFF will fail that check even though its a valid number (-1). 

Am I missing anything here? Does HIGH consider the high bits of Var to be zero if Var is -1?

ivg:
Let's simplify the program even more and consider only the division instruction, which is encoded as `f7 7d f8` in your bigger example,  here is the full bil output of this 

```fortran
$ bap-mc 'f7 7d f8' --show-insn=asm --show-bil
idivl -0x8(%rbp)
{
  if ((if high:1[mem[RBP - 8, el]:u32]
         then pad:64[mem[RBP - 8, el]:u32] | 0xFFFFFFFF00000000
         else pad:64[mem[RBP - 8, el]:u32]) = 0) {
    cpuexn (0)
  }
  #1 := (low:32[RDX].low:32[RAX]) /$ (if high:1[mem[RBP - 8, el]:u32]
                                        then pad:64[mem[RBP - 8, el]:u32] | 0xFFFFFFFF00000000
                                        else pad:64[mem[RBP - 8, el]:u32])
  #2 := (low:32[RDX].low:32[RAX]) %$ (if high:1[mem[RBP - 8, el]:u32]
                                        then pad:64[mem[RBP - 8, el]:u32] | 0xFFFFFFFF00000000
                                        else pad:64[mem[RBP - 8, el]:u32])
  if (high:32[#1] = 0) {
    
  }
  else {
    cpuexn (0)
  }
  #3 := low:32[#2].low:32[#1]
  RAX := pad:64[extract:31:0[#3]]
  RDX := pad:64[extract:63:32[#3]]
  CF := unknown[bits]:u1
  OF := unknown[bits]:u1
[...]

PR #962, opened 2019-07-31
fixes signed division and tweak both division representations

ivg:
The signed division operation was improperly encoded in BAP since long
time ago (probably since re-introduction of x86 in 0.9.x, back in
2014). As reported by @icemonster (#961), the overflow check was
incorrectly treating singed integers and any signed division by a
negative number was triggering the division error exception.

A small history research showed that in BAP 0.8 the overflow check
was commented out for signed division, with a corresponding (and
wrong) comment that it should be implemented and how it should be
implemented. It looks like later, during porting this code was
uncommented.

The new implementation implements a correct check as well as tweaks
the code so that it now produces a more compact and readable code, as
well as the lifting code is also now more readable and looks more like
a specification.

Here is the code generated for the signed division:
```

$ bap-mc 'f7 7d f8' --show-insn=asm --show-bil
idivl -0x8(%rbp)
{
  #1 := if high:1[mem[RBP - 8, el]:u32]
          then pad:64[mem[RBP - 8, el]:u32] | 0xFFFFFFFF00000000
          else pad:64[mem[RBP - 8, el]:u32]
  #2 := low:32[RDX].low:32[RAX]
  if (#1 = 0) {
    cpuexn (0)
  }
  else {
    #3 := #2 /$ #1
    #4 := #2 %$ #1
    if (0x7FFFFFFFFFFFFFFF <$ #3 | #3 <$ 0x8000000000000000) {
      cpuexn (0)
    }
    else {
      #5 := low:32[#4].low:32[#3]
      RAX := pad:64[extract:31:0[#5]]
      RDX := pad:64[extract:63:32[#5]]
    }
  }
  CF := unknown[bits]:u1
  OF := unknown[bits]:u1
  SF := unknown[bits]:u1
  ZF := unknown[bits]:u1
  AF := unknown[bits]:u1
  PF := unknown[bits]:u1
}
```

and for the unsigned one

```
bap-mc 'f7 75 f8' --show-insn=asm --show-bil
divl -0x8(%rbp)
{
  #1 := pad:64[mem[RBP - 8, el]:u32]
  #2 := low:32[RDX].low:32[RAX]
  if (#1 = 0) {
    cpuexn (0)
  }
  else {
    #3 := #2 / #1
    #4 := #2 % #1
    if (high:32[#3] = 0) {
      #5 := low:32[#4].low:32[#3]
      RAX := pad:64[extract:31:0[#5]]
      RDX := pad:64[extract:63:32[#5]]
    }
    else {
      cpuexn (0)
    }
  }
  CF := unknown[bits]:u1
  OF := unknown[bits]:u1
  SF := unknown[bits]:u1
  ZF := unknown[bits]:u1
  AF := unknown[bits]:u1
  PF := unknown[bits]:u1
}
```

a small smoke test was implemented as C program,
```c
 #include <stdio.h>
 #include <limits.h>

int stest(const char *msg, int dividend, int divisor, int expected) {
    int result = dividend / divisor;
    if (result == expected)
        puts("PASS");
    else
        puts(msg);
}

unsigned utest(const char *msg, unsigned dividend, unsigned divisor, unsigned expected) {
    unsigned result = dividend / divisor;
    if (result == expected)
        puts("PASS");
    else
        puts(msg);
}

int main(int argc, char const *argv[])
{
    stest("10 / -1 <> -10", 10,-1,-10);
    stest("10 / -2 <> -5", 10,-2,-5);
    stest("10 / -10 <> -1", 10,-10,-1);
    stest("10 / 1 <> 10", 10,1,10);
    stest("10 / 2 <> 5", 10,2,5);
    stest("10 / 10 <> 1", 10,10,1);
    utest("10 / 10 <> 1", 10,10,1);
    utest("10 / 1 <> 10", 10,1,10);
    utest("10 / 100 <> 0", 10,100,0);
    utest("10 / UINT_MAX <> 0", 10,UINT_MAX,0);
    stest("10 / 0 is #defined", 10,0,10);
    puts("FAIL");               /* this should not be reached */
}
```

which correctly yields all passes when run with `bap ./example --run`.

Note, this test doesn't really cover the overflow check when an actual
overflow happens, because I don't know how to force a C compiler to
generate code that will lead to an integer overflow from the
division. If anyone knows, please tell me :)

The file is attached. 

[example.zip](https://github.com/BinaryAnalysisPlatform/bap/files/3452402/example.zip)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: c
+dist: trusty
+
 
 sudo: required
 
--- a/plugins/x86/x86_lifter.ml
+++ b/plugins/x86/x86_lifter.ml
@@ -21,6 +21,11 @@ module ToIR = struct
 
   let size_of_typ s = Size.of_int_exn !!s (** doubts here  *)
 
+
+  let undefine =
+    List.map ~f:(fun r -> Bil.(r := unknown "#undefined" (Var.typ r)))
+
+
   (* stmt helpers *)
 
   let store_s mode s t a e =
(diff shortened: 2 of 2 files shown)


ivg:
please ignore Travis, our CI is experiencing networking issues since yesterday (it is unable to connect to the PPA server)

ivg:
OK, the CI issues is also resolved, they have switched the repo yesterday or so.

PR #964, opened 2019-08-06
Fixes primus self seeding generators

ivg:
Primus self seeding generators weren't storing their state and were recreating themselves on every call, which basically turned them into static zero generators.

This PR fixes them, so now we should observe random values read from uninitialized memory (note that our malloc allocator initializes memory with all zeros). 

This PR fixes another issue that became apparent after the generator fix - our memory component wasn't memoizing values read from dynamic and static layers, therefore two consequent reads were generating different values. This is akin to the problem in Env fixed in #958.

### DIFF
--- a/lib/bap_primus/bap_primus_generator.ml
+++ b/lib/bap_primus/bap_primus_generator.ml
@@ -17,37 +17,45 @@ let generators : Univ_map.t state = Bap_primus_machine.State.declare
     ~uuid:"7e81d5ae-46a2-42ff-918f-96c0c2dc95e3"
     (fun _ -> Univ_map.empty)
 
+module States = Univ_map
 
-module States = Univ_map.With_default
+module type Iter =
+  Iterator.Infinite.S with type dom = int
+
+type 'a iter = (module Iter with type t = 'a)
 
 type 'a gen = {
-  state : 'a States.Key.t;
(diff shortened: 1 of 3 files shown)


lib/bap_primus/bap_primus_generator.ml
```diff
@@ -17,37 +17,45 @@ let generators : Univ_map.t state = Bap_primus_machine.State.declare
...
+}
+
+type 'a key = 'a gen States.Key.t
+
+type 'a ready = {
+  key : 'a key;
+  gen : 'a gen;
+}
+
+type 'a wait = {
```
ivg: wait is a bad name here

We basically have two kinds of generators:
- context independent (Ready and Static)
- context dependent

Omitting static for its triviality, the context independent, the context independent has type `'a gen` while context dependent has type `int -> 'a gen`, where `int` denotes the context (it will be the hash of the machine identifier). Both a paired with the key, which is used to store them in the state. We store a value of type `'a gen` in the state. On the first call to the generator (if we don't have the generator stored in the state, then it is the first call), we either take the default state (which we always have) or we use the `init` function to get the context dependent initial value. Then we just apply the generator and store its in the state. On the next calls, if the generator is in the state we always use it, no matter if it is a context dependent or not.

PR #971, opened 2019-08-13
 limits the continuation at the call forkpoint in the promiscuous mode

gitoleg:
## Context
In the promiscuous mode we fork a new machine on each new call, so that if the function doesn't return for any reason, we can continue exploration of the code located after the call. 

## Problem
The stored continuation wasn't limited and was capturing the whole execution path, including the path in which the function was actually returning. That lead to quite funny false positives in our analysis. 

## Solution

The continuation is now explicitly delimited with `halt`.


ivg:
Since in 1.6 we do not test promiscuous mode at all (unlike 2.x where we do), I don't see any reason to wait for the tests to pass, so I'm merging this bypassing our CI.

PR #957, opened 2019-06-19
Try to build with OCaml 4.08 version

XVilka:
With the new OCaml 4.08 release there are lot of nice optimizations and features out of the box, even without changing the current code. Worth to make sure it builds. For now lets see if everything passes OK.

PR #972, opened 2019-08-15
Use 4.07.1 for Xenial docker

XVilka:


ivg:
LGTM, but this PR actually rises a question about this Dockerfile in general, it is a reminiscent of the older ages, when we had to build bap and opam from scratch, so honestly, I'm not sure why do we need it at all. The modern dockerfile for bap needs only one command,
```
FROM ocaml/opam2

RUN opam repo add bap git://github.com/BinaryAnalysisPlatform/opam-repository#testing \
 && opam update \
 && opam depext --install bap --yes
```

XVilka:
Yes, it makes sense to drop these files instead, drop trusty (who uses it anyway?) support for BAP 2.0 and use modern way to compile then 👍

XVilka:
Should I close this PR then? @gitoleg

gitoleg:
well, I would say that these files are old enough and we definitely need to update them. But it's up to you - if you don't have time for that - then just close and leave it t me, one day I need to do it anyway. Otherwise, you can use the next snippet for inspiration, I still didn't find time to test it.
```

XVilka:
It makes sense also to remove some old docker files as well.

XVilka:
We also can do a multistage build for all docker containers, so the size of resulting image will be significantly less.

XVilka:
Is irrelevant becase of https://github.com/BinaryAnalysisPlatform/bap/pull/957

PR #980, opened 2019-08-29
`break` doesn't work in `case` for shell

XVilka:
It works only in the loops. Spotted the warning with [shellcheck](https://github.com/koalaman/shellcheck).
See [SC2105](https://github.com/koalaman/shellcheck/wiki/SC2105) explanation at their wiki:
> `break` or `continue` was found outside a loop. These statements are only valid in loops. In particular, `break` is not required in `case` statements as there is no implicit fall-through.


ivg:
Thanks!

PR #982, opened 2019-09-03
Adding edge functions to fixpoint algorithm

Enkelmann:
This PR adds optional edge functions to the fixpoint algorithm. This is important for example in value set analysis, when one can gain knowledge on the current values depending on whether a conditional jump was taken or not.

Things that should be done before merging:
- [ ] Add tests
- [ ] Add documentation

Could you point me to the place to add both of these? For tests I suppose the *lib_tests* folder is right, but I don't know how one is supposed to compile and run these tests.

### DIFF
--- a/lib/graphlib/graphlib_graph.ml
+++ b/lib/graphlib/graphlib_graph.ml
@@ -1368,7 +1368,7 @@ module Fixpoint = struct
       | Some x -> x
       | None -> default in
     let get_edges ~start_node ~end_node =
-      let successors = if rev then G.Node.outputs start_node g else G.Node.inputs start_node g in
+      let successors = if rev then G.Node.inputs start_node g else G.Node.outputs start_node g in
       Seq.filter ~f:(fun edge -> (if rev then G.Edge.src edge else G.Edge.dst edge) = end_node) successors
     in
     let step visits works approx = match Set.min_elt works with
@@ -1383,7 +1383,7 @@ module Fixpoint = struct
               let ap' = match edge_fn with
                 | None -> merge out ap
                 | Some edge_function ->
-                  let edges = get_edges ~start_node:nodes.(current_node) ~end_node:nodes.(n) in (* TODO: Check whether I have to iterate over the nodes here instead *)
+                  let edges = get_edges ~start_node:nodes.(current_node) ~end_node:nodes.(n) in
                   Seq.fold ~f:(fun ap' edge -> merge ap' (edge_function edge out)) ~init:ap edges in
               let visits,ap' = user_step visits n ap ap' in
               if equal ap ap' then (visits,works,approx)
--- a/lib/graphlib/graphlib_graph.ml
+++ b/lib/graphlib/graphlib_graph.ml
@@ -1367,10 +1367,6 @@ module Fixpoint = struct
     let get approx n : d = match Map.find approx n with
       | Some x -> x
       | None -> default in
-    let get_edges ~start_node ~end_node =
-      let successors = if rev then G.Node.inputs start_node g else G.Node.outputs start_node g in
-      Seq.filter ~f:(fun edge -> (if rev then G.Edge.src edge else G.Edge.dst edge) = end_node) successors
-    in
     let step visits works approx = match Set.min_elt works with
       | None -> Done approx
       | Some current_node ->
@@ -1383,8 +1379,11 @@ module Fixpoint = struct
               let ap' = match edge_fn with
                 | None -> merge out ap
                 | Some edge_function ->
-                  let edges = get_edges ~start_node:nodes.(current_node) ~end_node:nodes.(n) in
-                  Seq.fold ~f:(fun ap' edge -> merge ap' (edge_function edge out)) ~init:ap edges in
+                  let edge = if rev then
+                      Option.value_exn (G.Node.edge nodes.(n) nodes.(current_node) g)
+                    else
+                      Option.value_exn (G.Node.edge nodes.(current_node) nodes.(n) g) in
+                  merge ap' (edge_function edge out) in
               let visits,ap' = user_step visits n ap ap' in
               if equal ap ap' then (visits,works,approx)
               else visits,
--- a/lib/graphlib/graphlib_graph.ml
+++ b/lib/graphlib/graphlib_graph.ml
@@ -1340,9 +1340,9 @@ module Fixpoint = struct
      the correct meet, f,  and initial approximation.
 
   *)
-  let compute (type g n d)
-      (module G : Graph with type t = g and type node = n)
-      ?steps ?start ?(rev=false) ?step
+  let compute (type g n e d)
+      (module G : Graph with type t = g and type node = n and type edge = e)
+      ?steps ?start ?(rev=false) ?step ?edge_fn
       ~init:(Solution {approx; iters; default}) ~equal ~merge ~f g : (n,d) t =
     let nodes =
       reverse_postorder_traverse (module G) ~rev ?start g |>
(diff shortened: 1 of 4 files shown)


ivg:
So far the graphlib tests reside in the `lib_test/bap_types/test_graph.ml` file. Just add a new case and add to the `suite` at the end of the file. To run tests, first enable them, e.g.,
```
./configure <other-options> --enable-tests
```
and then you can use the `make test` command to run them, or manually run them via the `./run_tests.native` binary located in the bap folder. (it lets you to define which tests to run).

FYI, when you make a PR tests are enabled and run during the CI, though it would be surprising if they won't :)

lib/graphlib/graphlib_graph.ml
```diff
@@ -1367,16 +1367,24 @@ module Fixpoint = struct
     let get approx n : d = match Map.find approx n with
       | Some x -> x
       | None -> default in
+    let get_edges ~start_node ~end_node =
+      let successors = if rev then G.Node.inputs start_node g else G.Node.outputs start_node g in
+      Seq.filter ~f:(fun edge -> (if rev then G.Edge.src edge else G.Edge.dst edge) = end_node) successors
```
ivg: 1. This predicate uses the polymorphic equality `(=)`, instead of `G.Node.(=)`, that will make equal nodes unequal, or vice versa, or terminate the program if nodes have functional values in them
2. Between two nodes there could be 0 or 1 edges, not arbitrary many of them, so `Sequence.t` is sort of unsuitable here.

And in general, if we want to iterate over edges, then we should do this using `G.Node.{inputs,outputs}` instead of iterating over nodes, and then asking for the first node outputs and then filtering for those that matches with destination. At least we can use the `G.edge` function that returns an edge between two nodes.

Enkelmann: 1. You're right, I will fix it (tomorrow, when I am in the office again).
2. In a general graph, there can be more than one edge between nodes. Think e.g. the callgraph when one function is called more than once by another one and the user wants to distinguish between those calls. I actually haven't tested it, but I hope graphlib allows the creation of such graphs.

ivg: > 2. In a general graph, there can be more than one edge between nodes ... but I hope graphlib allows the creation of such graphs.

Nope, Graphlib doesn't allow this kind of edges. An edge is totally defined by its two nodes. We follow the [definition][1] of the _simple directed graph with loops_ as a pair of two sets, `G = (V,E)` where a `E` is a set of ordered pairs of nodes. We specifically disallowed multigraphs and undirected graphs, as they could be always represented using an equivalent digraph, and therefore do not give any more expressivity, but their theory is less developed and the implementation is prone to be inefficient.

Multiple edges between two nodes could be always represented by either inserting pseudonodes or by extending the edge label with a proper annotation. For example, (actually using your example), a callgraph in BAP is represented as a simple digraph with edges labeled by a set of callsites. So no information is lost, but the representation is more compact and easier to analyze. 

And this brings us back to the PR, which I think is very necessary - since the edge labels bear important information, but ou solver doesn't provide access to them, basically assuming that all edges are unlabeled.

So, for starters, let's assume that we're designing the interface from the very beginning. The following is a dump of thoughts, a little bit unstructured, but I think it is better to share it here, instead of providing the end result without showing the whole train of thoughts.

The nodes, unfortunately, serve two purposes:
 - as variables, they denote program points
 - as basic blocks, they denote the operation of the path

consequently, a basic block also consists of two parts - one denoting the data transformation and the other the control. The latter part is reified into edges, thus a basic block is represented as a node plus a set of outcoming edge labels. 

The path transmittance, therefore is function of both node and the edge label, therefore the transfer function, that denotes the transmittance must have `Node.t -> Edge.label -> 'd`, so that we can define the transmittance based on both the node and the edge. An alternative definition would be `Edge.t -> 'd` but I'm afraid that this interface is very easy to abuse, given that the `Edge.t` is a triple `(src,label, dst) : Node.t * Edge.label * Node.t` I'm reluctant to give up the dst node as then the transmittance will become a function of both node and destination which will break the locality property and independence of variables. So, instead of exposing the whole edge, let's expose only the edge payload, following the principle the a good interface is hard to use incorrectly. 

Now it's time to revisit your suggestion, which is really nice in terms of minimality of changes (which I really appreciate). I think it has one major flaw, that will kick us in the future. It splits the transfer function into two functions, `node: Node.t -> 'd -> 'd` and `edge: Edge.label -> 'd -> 'd` so that the full transmittance could expressed as their composition `edge . node`. The main problem with this approach is that oftentimes the transmittance is a function of both node and edge, i.e., when the transmittance is not distributive. In that case a user will face the same problem as it has now, and will be forced to invent tricks, e.g., representing the `'d` type as some sort of stateful computation, like `type data = Wait_for_edge of node | Ready of data`, a solution which will yield problems with equality and provoke ugly anomalies (like having to deal with impossible states). 

Therefore, my opinion, is that instead of the sticking to a partial (but compromising) solution which will impose a burden of technical dept on us, we should implement a proper solution with the transfer function parametrized by both parties of the transmittance relation and provide the old interface (in which the transmittance function depends only on a node) for the backward compatibility (potentially attributing it with the deprecation message).

OCaml-wise, we can do the following, fix the `compute` function in the `Solution` module (which is in the implementation is known as `Fixpoint` - we should probably also rename it to Solution just for the consistency matters). 

Implement the current `fixpoint` function in terms of `Solution.compute`, which will just ignore the edge label. 


Concerning the implementation. We represent successors as a set of the rpost numbers. Having it a set was probably a mistake - it is already guaranteed by the definition of a graph that the all neighbours are distinct, so putting them into a set is not necessary. We also do not need a fast lookup, is we always just iterate over them. Therefore, we can just represent `succs` as a list of `(int * Edge.label)` elements.

The only thing to consider here is the order of visitation of the node successors. Right now we visit all neighbors of a node in the topological order. However the order shall not matter given a correct meet operator which should be commutative and associative, i.e., `x /\ y = y /\ x` and `x /\ (y /\ z) = (x /\ y) /\ z`. The truth is that many users do not provide correct meet operators. But bad input bad output, right? In any case, even given an incorrect meet operator there is no better or worse order of visiting neighbors (and previously it didn't matter at all, since edges weren't considered). To summarize, order doesn't matter and we can safely propagate data in an arbitrary order to successors. So we can store them in a list.

Now the last point, since now the transfer function is called for for each edge rather than for each node we accidentally introduce unnecessary computations in cases where the transmittance relation is idempotent to the edge (especially when the edge type is unit). To clarify, suppose a transfer function doesn't depend on the edge (e.g., a classical dataflow problem), but is quite expensive to compute, given the transfer function as we designed it so far, 

```ocaml
let transfer node _edge data = 
    compute_genkill node data
```
the `compute_genkill` function will be called twice with exactly the same inputs, which is unnecessary inefficient. I'm thinking of a couple solutions to this problem, but probably the easiest and less invasive solution would be to use partial applications. This will, however, require us to flip the order of arguments, so that our transfer function will now have type `'d -> node -> Edge.label -> 'd`, which naturally corresponds to the flow of signal through the graph (i.e., data first enters node, that transfers through edge), but most importantly will allow a user to reflect the fact that the transmittance is constant wrt to the edge label, e.g.,
```
let transfer data node =
   let data = compute_genkill node data in
   fun _ -> data
```

This is not an extremely ideal solution, I immediately can see a couple of problems (like that it is being implicit and that we will create an extra closure on each transfer), but to be honest other solutions that I have in mind suffer from even more problems. 

So, I'm eager to listen to you opinion, what do you think?
    


To summarize, that's the type of the `compute` function that I'm thinking of:
```
module Solution : sig
...
 val compute : (module Graph with type t = 'c
                              and type node = 'n
                              and type Edge.label = 'e) ->
  ?steps:int -> ?start:'n -> ?rev:bool ->
  ?step:(int -> 'n -> 'd -> 'd -> 'd) ->
  init:('n,'d) Fixpoint.t ->
  equal:('d -> 'd -> bool) ->
  merge:('d -> 'd -> 'd) -> f:('d -> 'n -> 'e -> 'd) -> 'c -> ('n,'d) Fixpoint.t

end
```


[1]: https://en.wikipedia.org/wiki/Directed_graph

ivg: I would also like to @fortunac, @codyroux, @jtpaasch, @ccasin, @DIJamner to the discussion, since they should also be interested in this PR.

Enkelmann: - Well, not allowing multigraphs is unfortunate for me as I am already aware of several cases in my use cases where I have to deal with them. Passing the responsibility of dealing with multigraphs to the user is too error-prone in these cases in my opinion. But that's my problem, I should have read the docs. I will simplify the code accordingly.
- Putting the node and edge transition functions together is probably the right idea from a conceptual standpoint. From a practical standpoint this will result in unnecessary computations which may hurt if both node and edge transitions are computationally expensive. That said, I didn't do it the right way in my PR either, I just wanted to present a minimal and backwards-compatible solution for now.
- Even with the edge functions the fixpoint algorithm will probably not be powerful enough for some of the algorithms I want to implement. I will probably develop a small zoo of fixpoint algorithms and then try to generalize a good interface out of them. But this will take some months. If you want, you can wait until then. But providing a minimal viable improvement now may be worth it if we are sure that this won't break something later. I do not assume that my end result will have a backwards-compatible interface, just that one can express the old fixpoint algorithm in terms of the new one. So I do not think this can hurt us, except if you do not assume any other improvements on the BAP-version of the fixpoint algorithm will happen after this one.
- I am also eager to hear the opinions of other users of the fixpoint function. :-)

ivg: > From a practical standpoint this will result in unnecessary computations which may hurt if both node and edge transitions are computationally expensive

Yep, but this problem is already addressed (starting from the paragraph "Now the last point, since now the transfer function is called for each edge...") , and the existing `fixpoint` function will be implemented so that the it will be called only once for each node (this will still involve a little bit of extra comparison and merges). An alternative solution would be to let users explicitly state which relation that would like to encode, i.e., to wrap the transfer function into a variant (names a tentative):
```ocaml
type transfer =[
   | `Node of Node.t -> 'd -> 'd
   | `Edge of Edge.label -> 'd -> 'd
   | `Both of Node .t -> Edge.label -> 'd -> 'd
]
```

ivg:
Closing as it is no longer, but it was a good idea, so if anyone is interested to continue, feel free to reopen.

PR #984, opened 2019-09-04
preserves falls only after 'real' calls

gitoleg:
Control flow should return only if an instruction is a real call, i.e. if `Insn.is call` is true. Otherwise, it's just a control transfer and should be done with no return.


ivg:
This is unsound, we can't assume that an instruction is not a call (or more precisely, that the instruction is a call without return) if there is no `is-call` flag.

ivg:
Instead of this kludge we should develop an analysis that will proper identify calls that do not return and erase false edges.

Issue #973, opened 2019-08-15
use ELF sections instead of segments to provide accurate attributes


ivg:
ELF files have two different views on their layout - segments and sections. BAP loaders are using the segment view, in which there is a big contiguous loadable and executable segment (usually named "2"), which contains many sections, including .plt, .text, and, even .rodata.

The loader, at least on Linux, is mapping the whole segment to memory, but then assigns different permission attributes based on _sections_, so despite the fact that .rodata belongs to an executable section it is still marked as non-executable by the kernel.

Therefore, we can end up disassembling the rodata section (and some other non-executable sections of executable segments). We can reach those sections either via byteweight's false positives or through our new speculative disassembler.

Usually, the disassembler is able to rebuke most if not all functions in the non-executable code sections by proving that all chains of executions terminate with data. This, however takes time, and is still not guaranteed.

Therefore, it sounds reasonable to use the available information, and do not disassemble non-executable sections. Thus we can disassemble faster and our analyses won't spend time analyzing functions that will never be executed.

An alternative [solution][1], would be to try to alleviate the problem downstream using our new knowledge base. But it is not really satisfying, since we now have conflicting views on the binary, some components such as byteweight still believe that the whole segment is executable, while others, not. Therefore, the problem should be fixed at the very beginning, on the loader level. 

Also, since the new disassembler works fine with several memory regions (even if the call each other) we could probably get rid of segments at all. We, probably, do not need them at all.

[1]: https://github.com/ivg/bap/commit/f897f68be08e62140e276ecddf900d4b00696946

PR #988, opened 2019-09-10
introduces a more fine-granular view on the image memory

gitoleg:
The current implementation is heavily influenced by the segmented view on the memory layout of a program. For historical reasons, the ELF representation of segments stopped to represent accurately the memory layout of the program, see #973 for more details. As a result, we ended up disassembling regions of memory which are clearly data (like `.rodata`) which reside in the executable segments. 

This PR introduces a new abstraction, called `code-region` since we don't want to overload the meaning of the segment/section anymore. All supported backends are implementing this abstraction in their specific manner (some are trivial, like COFF, some are not, like ELF)

The good thing about all the file types is that each of them has a section layout where sections have flags, that makes it possible to decide whether a section contains a code or not. Here is a short table for the different file types and sections flags that we are looking for.

File type  | Flags
-----------|--------
COFF       | IMAGE_SCN_CNT_CODE or IMAGE_SCN_MEM_EXECUTE
ELF        | SHF_EXECINSTR 
MACHO      | S_ATTR_PURE_INSTRUCTIONS or S_ATTR_SOME_INSTRUCTIONS

Given that, it's not that hard to define regions with an executable code.

### Bugs
There are few bugs that were fixed as well:
1) restores `-dasm` option for the coff object files.  
   COFF section header has two fields that denotates size: virtual size and raw data size.
   The virtual size is meaningful only for executable files and it's a size of the section when it's loaded
   into a memory.
   But for the object files this field doesn't have any sense and equal to zero. That's why we didn't mark memory in `Image` module correctly and didn't have any output for `-dasm` option. So we need to   use raw data size instead.

2) fixes wrong addresses for object files.
   Our llvm backend provides relative addresses for all the pieces of information (sections, symbols 
   etc.) and a default base address as well, so we can substitute a default if we need to.
   That works fine for executable files. But when we deal with the object files the problem becomes 
   visible, since we don't actually have any addresses at all and have to use offsets instead.
   Previously we were taking offsets as they are, and a first byte of code had an address equal to the 
   section offset where this byte belonged.
   This PR fixes this and use offsets relative to the offset of the first executable section, so the first byte 
   will have address 0,  as it's expected.
  

This PR fixes https://github.com/BinaryAnalysisPlatform/bap/issues/973

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -5218,11 +5218,6 @@ module Std : sig
     (** symbol  *)
     type symbol [@@deriving bin_io, compare, sexp]
 
-    (** code region  *)
-    type code_region [@@deriving bin_io, compare, sexp]
-    (** data region *)
-    type data_region [@@deriving bin_io, compare, sexp]
-
     type path = string
 
     (** {2 Constructing}  *)
@@ -5278,14 +5273,6 @@ module Std : sig
     (** [symbols image] returns a mapping from addresses to symbols *)
     val symbols : t -> symbol table
 
-    (** [code_regions image] returns a mapping from address to a
-        region with code *)
-    val code_regions : t -> code_region table
-
-    (** [data_regions image] returns a mapping from address to a
-        region with data *)
-    val data_regions : t -> data_region table
-
     (** {2 Tags}  *)
 
     (** tags a segment  *)
@@ -5298,10 +5285,7 @@ module Std : sig
     val section : string tag
 
     (** tags a code region  *)
-    val code_region  : code_region tag
-
-    (** tags a data region  *)
-    val data_region  : data_region tag
+    val code_region  : unit tag
 
     (** an image specification in OGRE  *)
     val specification : Ogre.doc tag
@@ -5513,10 +5497,6 @@ module Std : sig
 
       val code_region :
         (addr * size * off, (addr -> size -> off -> 'a) -> 'a) Ogre.attribute
-
-      val data_region :
-        (addr * size * off * bool, (addr -> size -> off -> bool -> 'a) -> 'a) Ogre.attribute
-
     end
   end
 
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -5218,6 +5218,11 @@ module Std : sig
     (** symbol  *)
     type symbol [@@deriving bin_io, compare, sexp]
 
+    (** code region  *)
+    type code_region [@@deriving bin_io, compare, sexp]
+    (** data region *)
+    type data_region [@@deriving bin_io, compare, sexp]
+
     type path = string
 
     (** {2 Constructing}  *)
@@ -5273,6 +5278,14 @@ module Std : sig
     (** [symbols image] returns a mapping from addresses to symbols *)
     val symbols : t -> symbol table
 
+    (** [code_regions image] returns a mapping from address to a
+        region with code *)
+    val code_regions : t -> code_region table
+
+    (** [data_regions image] returns a mapping from address to a
+        region with data *)
+    val data_regions : t -> data_region table
+
     (** {2 Tags}  *)
 
     (** tags a segment  *)
@@ -5283,6 +5296,12 @@ module Std : sig
 
     (** tags a section  *)
     val section : string tag
+
+    (** tags a code region  *)
+    val code_region  : code_region tag
+
+    (** tags a data region  *)
+    val data_region  : data_region tag
 
     (** an image specification in OGRE  *)
     val specification : Ogre.doc tag
@@ -5491,6 +5510,13 @@ module Std : sig
       (** [base_address addr] this is the base address of an image,
           i.e., an address of a first byte of the image.  *)
       val base_address : (addr, (addr -> 'a) -> 'a) Ogre.attribute
+
+      val code_region :
+        (addr * size * off, (addr -> size -> off -> 'a) -> 'a) Ogre.attribute
+
+      val data_region :
+        (addr * size * off * bool, (addr -> size -> off -> bool -> 'a) -> 'a) Ogre.attribute
+
     end
   end
 
--- a/lib/bap_image/bap_image.ml
+++ b/lib/bap_image/bap_image.ml
@@ -53,13 +53,15 @@ let pp_region fmt {name; locn={addr; size}} =
 
 let hash_region {locn={addr}} = Addr.hash addr
 
-module Segment = struct
+module Mapped_region(N : sig
+             val name : string
+           end) = struct
 
   module T = struct
     type t = mapped region [@@deriving bin_io, compare, sexp]
     let hash = hash_region
     let pp = pp_region
-    let module_name = Some "Bap.Std.Image.Segment"
+    let module_name = Some N.name
     let version = "2.0.0"
   end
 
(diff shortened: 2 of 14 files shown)


lib/bap/bap.mli
```diff
@@ -5218,6 +5218,11 @@ module Std : sig
     (** symbol  *)
     type symbol [@@deriving bin_io, compare, sexp]
 
+    (** code region  *)
```
ivg: we don't need this in the `Image` interface at all. Please remove `{code,data}_region` data types and functions.

lib/bap/bap.mli
```diff
@@ -5284,6 +5297,12 @@ module Std : sig
     (** tags a section  *)
     val section : string tag
 
+    (** tags a code region  *)
+    val code_region  : code_region tag
```
ivg: why not just `unit tag`?

lib/bap/bap.mli
```diff
@@ -5284,6 +5297,12 @@ module Std : sig
...
+
+    (** tags a data region  *)
+    val data_region  : data_region tag
```
ivg: We can also represent it with `unit tag` or `bool tag` (if we want to discriminate between writable and non-writable)

lib/bap_image/bap_image.ml
```diff
@@ -120,6 +131,8 @@ type t = {
   data : Bigstring.t;
   symbols : symbol table;
   segments : segment table;
+  data_regions : data_region table;
```
ivg: drop this, it is duplication of the `memory` data structure. Moreover, you already have them in the `spec` entry.

lib/bap_image/bap_image.ml
```diff
@@ -214,7 +243,8 @@ let make_table add base memory =
 
 let make_symtab = make_table add_sym
 let make_segtab = make_table add_segment
-
+let make_codtab = make_table add_code
```
ivg: hmm.... stop it :)

lib/bap_image/bap_image.ml
```diff
@@ -446,16 +489,44 @@ module Derive = struct
...
+    Fact.foreach Ogre.Query.(begin
+        select (from data_region $ named_region)
+          ~join:[[field addr];]
+      end) ~f:(fun (addr, size, off, write) {info=name} ->
+        location ~addr ~size >>= fun locn ->
+        int_of_int64 off  >>= fun off ->
+        int_of_int64 size >>| fun len ->
+        {locn; name; info={endian;off;len;r=true;w=write;x=false}}) >>=
+    Fact.Seq.all
+
```
ivg: Those two functions looks the same to me, consider generalizing

lib/bap_llvm/bap_llvm_ogre_coff.ml
```diff
@@ -37,8 +37,35 @@ module Make(Fact : Ogre.S) = struct
...
+                 [field size ~from:section_entry;
+                  field size ~from:data_entry]])
+      ~f:(fun _ (_,addr,vsize) (_,off,_,w) -> Int64.(base + addr),vsize,off,w) >>= fun s ->
+    Fact.Seq.iter s
+      ~f:(fun (addr, size, off, w) ->
+        Fact.provide data_region addr size off w)
+
+  let regions =
+    code >>= fun () ->
+    data
```
ivg: again, repetitive code

lib/bap_llvm/bap_llvm_ogre_coff.ml
```diff
@@ -69,9 +96,37 @@ module Relocatable = struct
...
+      Fact.require base_address >>= fun base ->
+      Fact.foreach Ogre.Query.(
+        select (from section_entry $ data_entry)
+          ~join:[[field name];
+                 [field size ~from:section_entry;
+                  field size ~from:data_entry]])
+        ~f:(fun _ (_,off,size,w) -> Int64.(base + off),size,off,w) >>= fun s ->
+      Fact.Seq.iter s
+        ~f:(fun (addr, size, off, w) ->
+          Fact.provide data_region addr size off w)
```
ivg: so those two (nearly identical, again) functions just map `code_entry -> code_region` and `data_entry -> data_region`, right?

lib/bap_llvm/bap_llvm_ogre_samples.ml
```diff
@@ -36,6 +37,38 @@ module Sections(Fact : Ogre.S) = struct
...
+    Fact.foreach Ogre.Query.(begin
+        select (from named_region $ data_entry)
+          ~join:[[field name];
+                 [field size ~from:named_region;
+                  field size ~from:data_entry]]
+    end)
+      ~f:(fun {addr; size;} (_,off,_,w) -> addr,size,off,w) >>= fun s ->
+      Fact.Seq.iter s ~f:(fun (addr,size,off,w) ->
+          Fact.provide data_region addr size off w)
+
```
ivg: I keep seeing the same code not only repeated twice, but repeated overall.

lib/bap_llvm/llvm_coff_loader.hpp
```diff
@@ -25,25 +25,32 @@ static const std::string coff_declarations =
...
     s.entry("section-entry") << sec.Name << sec.VirtualAddress << sec.SizeOfRawData << sec.PointerToRawData;
     s.entry("virtual-section-header") << sec.Name << sec.VirtualAddress << sec.VirtualSize;
     s.entry("section-flags") << sec.Name << r << w << x;
-    auto c = sec.Characteristics;
-    if ((c & COFF::IMAGE_SCN_CNT_CODE) ||
-        (c & COFF::IMAGE_SCN_CNT_INITIALIZED_DATA) ||
-        (c & COFF::IMAGE_SCN_CNT_UNINITIALIZED_DATA))
+    if (x)
         s.entry("code-entry") << sec.Name << sec.PointerToRawData << sec.SizeOfRawData;
+    else
```
ivg: are you sure that `not x` implies that the region is data?

lib/bap_llvm/llvm_elf_loader.hpp
```diff
@@ -137,6 +138,10 @@ void section_header(const T &hdr, const std::string &name, uint64_t base, ogre_d
     bool w = static_cast<bool>(hdr.sh_flags & ELF::SHF_WRITE);
     bool x = static_cast<bool>(hdr.sh_flags & ELF::SHF_EXECINSTR);
     s.entry("section-flags") << name << w << x;
+    if (x)
+        s.entry("code-entry") << name << hdr.sh_offset << hdr.sh_size;
+    else
```
ivg: again, `not SHF_EXECINSTR` does not imply that it is program data section, there are quite a few auxiliary and ancillary data structures used by the linker itself and various instrumentation tools, e.g., `.hash` section or `.dwarf`.

lib/bap_llvm/llvm_macho_loader.hpp
```diff
@@ -302,6 +304,34 @@ void relocations(const macho &obj, ogre_doc &s) {
...
 
+bool is_code_section(const macho &obj, const SectionRef &sec) {
+    auto flags = section_flags(obj,sec);
+    return static_cast<bool>
+        (flags & MachO::S_ATTR_PURE_INSTRUCTIONS ||
+         flags & MachO::S_ATTR_SOME_INSTRUCTIONS);
+}
+
+bool is_data_section(const macho &obj, const SectionRef &sec) {
+    return !(is_code_section(obj,sec));
```
ivg: the same issue `not (is_code) <> is_data`

PR #993, opened 2019-09-12
initializes PLT GOT in Primus

gitoleg:
In BAP 2.0 we have a more conservative disassembler. And while previously, only the first instruction of a PLT entry was disassembled, now we're getting the whole entry.

Unfortunately, the more precise disassembly hits analyses based on Primus. At least on x86/x86-64 targets. Previously, we got an unresolved call exception as soon as we enter a PLT entry, while now Primus follows to the whole chain of PLT resolution mechanism, like in the next example with call to free:
```
400560: pushq  0x200aa2(%rip)     # 601008 <_GLOBAL_OFFSET_TABLE_+0x8
400566: jmpq   *0x200aa4(%rip)    # 601010 <_GLOBAL_OFFSET_TABLE_+0x10>
40056c: nopl   0x0(%rax)<free@plt>:
400570: jmpq   *0x200aa2(%rip)        # 601018 <_GLOBAL_OFFSET_TABLE_+0x18>
400576: pushq  $0x040057b: jmpq   400560 <_init+0x28>>

```
which does the following
```
400576: increase the stack
40057b: jump to the very first plt entry
400560: increase the stack
400566: load the target from GOT table
```
Finally, at `400566` we're getting an unresolved call exception. And in the Primus promiscuous mode, we continue the execution.
Here comes the crux of the problem, the stack is now broken: stack was increased twice and there wasn't any `pop` or `ret` instruction, so the following execution is wrong. The underlying problem is that the runtime function that performs lazy resolution of the
PLT GOT entry is not stubbed and therefore the arguments that are passed to it via the stack are not popped.

Probably the right solution would be to stub this function, but this is quite and involved process and it is not immediately obvious how to link it correctly (we do not know neither name nor an address).

Another option, which is proposed in this PR, is to stop Primus from going that deep into the PLT entry by linking all GOT entries, by default, as unresolved. The ratification is simple -- if we got to the PLT entry, then the address wasn't linked, so there is no need to
continue and we can raise an unresolved function asap.

### DIFF
--- a/plugins/primus_x86/primus_x86_loader.ml
+++ b/plugins/primus_x86/primus_x86_loader.ml
@@ -99,8 +99,6 @@ module Component(Machine : Primus.Machine.S) = struct
       Value.of_int ~width addend >>= fun addend ->
       Primus.Linker.Trace.lisp_call_return >>> correct_sp sp addend
 
-  let seq = Machine.sequence
-
   let unresolve_plt_jumps p =
     Machine.Seq.iter (Plt.symbols p) ~f:Plt_entry.unresolve_jumps
 
--- a/plugins/primus_x86/primus_x86_loader.ml
+++ b/plugins/primus_x86/primus_x86_loader.ml
@@ -3,86 +3,74 @@ open Bap.Std
 open Bap_primus.Std
 open X86_cpu
 
-module Plt = struct
-  let is_plt_section v =
-    match Value.get Image.section v with
-    | Some x -> x = ".plt"
-    | _ -> false
+let is_section name v =
+  match Value.get Image.section v with
+  | Some x -> String.(x = name)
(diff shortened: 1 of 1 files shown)


plugins/primus_x86/primus_x86_loader.ml
```diff
@@ -3,11 +3,94 @@ open Bap.Std
...
+      method! enter_jmp c acc =
+        match find c with
+        | None -> acc
+        | Some a -> a :: acc
+    end)#visit_sub plt_sym []
+end
+
+module Make_unresolved(Machine : Primus.Machine.S) = struct
+  module Linker = Primus.Linker.Make(Machine)
+  open Machine.Syntax
```
ivg: you don't need this open.

PR #1005, opened 2019-10-29
introduces bap-main, unified frontend, new command line interface, etc

ivg:
# Table of Contents

- [Introduction](#introduction)
- [The Bap_main library](#the-bap_main-library)
- [The `bap` utility](#the-bap-utility)
- [The disassemble plugin](#the-disassemble-plugin)
- [The `raw` loader](#the-raw-loader)
- [The `mc` command](#the-mc-command)
- [A new `bap objdump` command](#a-new-bap-objdump-command)
- [The revamped Byteweight](#the-revamped-byteweight)
- [New  `baptop` toplevel](#new-baptop-toplevel)
- [The old `bap.top` library](#the-old-baptop-library)
- [A new bytecode frontend](#a-new-bytecode-frontend)
- [The recipe facility](#the-recipe-facility)
- [First-class readers and writers](#first-class-readers-and-writers)
- [Special thanks](#special-thanks)

# Introduction

This PR serves two purposes:
- to make it easier to use BAP as a library;
- to give users more control over BAP.

A supplement goal was to showcase the new interface, which ended up in a little bit bloated PR. We could probably split it, but it will defeat the main purpose and deliver weaker experience.

It has been observed for a long time, that BAP as a platform is not extremely friendly to other frameworks and it is hard to embed BAP in applications, that do not want to play by BAP rules and become a plugin. Embedding BAP in an application was notoriously hard, as not only BAP was interfering with the environment, but also because that significant pieces of functionality were implemented in the frontend, and wasn't available when the `bap` utility couldn't be used directly.

The `bap` utility, even when it could be used, had limitations on its own. It is built around the disassembly pipeline, letting the users inject their code as disassembly passes. It used to serve well and still works fine, but many applications just do not fit into this pipeline. Sometimes, the whole CFG is not needed, sometimes the application just has nothing to do with disassembling the binary. 

This PR addresses these two problems by
- moving a significant portion of bap frontend into the Bap_main library;
- letting the user choose the `main` procedure, which is now called _command_.

From the user perspective, it was reflected in a new command-line interface of the `bap` utility, which is now command-driven, the same as `git`, `docker`, or `opam`. For example, 

```
bap disassemble /bin/ls -d 
bap mc --show-insn=asm --show-bil -- 00f0
bap objdump /bin/ls 
```
No worries, you don't need to change your habits as the `disassemble` command is assumed by default, when the first argument is a file, the following will still work as expected:

```
bap /bin/ls -d
```

Moreover, the command name could be contracted to the smallest non-ambiguous prefix, e.g.,
```
bap dis /bin/ls 
```
The command-line arguments are now parsed using a global grammar, unlike previously, when each plugin was having its own grammar, and the whole command-line was never checked. It makes parsing more strict and, yes, now `bap --help` will print all possible command-line options. 

# The Bap_main library

The BAP Main library is the new entry point to BAP, both for embedding BAP in your applications and writing extensions to BAP (in other words, embedding your application into BAP). It provides a unified interface for writing BAP Extensions as well as makes it easy to initialize the platform from an arbitrary application. As usual, we reimplement the existing facilities using the new library, so there is no need to rewrite any existing code. 

## Command-line interface

The library provides a new API for specifying the command line interface of your plugin. The old interface (the `Bap_self().Config` module) is reimplemented in terms of Bap_main, so there is no need to change any existing code, but the new code is recommended to write using the new interface. The new interface is much easier to use than the old `Config` interface or the Cmdliner library itself (which used underneath the hood). It is also thoroughly documented (see the lib/bap_main/bap_main.mli file, the rendered documentation will be uploaded soon) with examples and even grammar specifications of each rule. 

## Single Configuration file

We abandoned the old style of configuration files (for those who didn't know it was always possible to pass BAP configuration parameters via a file) and switched to a new, single-file, style. In the old style, each plugin had its own configuration file in a separate folder. It turned to be unwieldy, so we decided to switch to a more common single file approach. All configuration parameters should be specified using their full and main name, e.g., `llvm-x86-syntax = intel`, no abbreviations, aliases, etc. 

## Recipes

The recipe system is supported directly by the Bap_main library and works as before. We're planning to extend the Recipe grammar with the `command` stanza, but for now, it works fine even without this extension. 

## Features and Requirements - purely functional caching

Beyond command line parsing and loading plugins, the library will take care of managing dependencies and computing cryptographic hashes of the input parameters, so that we can now build more reliable caching infrastructures. Any extension is invoked with the `ctxt` value which captures all environment and configuration parameters of this plugin and all other plugins on which the current plugin depends. The `Co# The `bap` utilityntext.digest` function returns the cryptographic hash of those parameters, that is guaranteed to change if any of the inputs change. The parameters are interpreted wrt to their semantics, e.g., parameters that are files are tracked by their contents rather by the name, directories are also handled specifically. To summarize, if you updated your byteweight signatures, then the caching system will automatically detect that the input was changed, even though the path didn't change, and the caching infrastructure will ensure that the disassembler is run again. The new features and requirements system plays an important part in this mechanism and enables expressing dependencies without mentioning implementations. It is described in detail in the Bap_main documentation.

## Markdown support

It is now possible to write man pages in markdown, which will be automatically rendered into the proper groff format. A small subset of markdown is supported, namely sections, itemized lists, and verbatim code. 

## The logging infrastructure

The logging infrastructure is also handled by the Bap Main library and it is now possible to specify a concrete log file or just log into a specified channel/formatter.

## Fewer dependencies -- better compilation times

As a side note, the `bap-main` library is having much fewer dependencies,  than the `bap` library is much smaller, and has a smaller signature, so depending on it instead of Bap.Std will make your compilation much faster. 

# The `bap` utility

The `bap` utility is rewritten from scratch (and even moved to another location - `src/bap_frontend.ml`). It is not doing any analysis and serves purely administrative needs. Beyond its main responsibility to run `Bap_main.init` it also provides the `list` command that enables introspection of different facilities (plugins, commands, recipes, features, tags) provided by BAP.

It also provides a default usage message in the style of docker or git, so just type `bap` on the command line and it will give you a comprehensive guide on bap, list all command and plugins, etc.

To hush this message use the `.` command, e.g.,

```
bap .
```

will print nothing. Useful, when you just writing a "hello, world" plugin and want to see the greeting :) 

# The disassemble plugin

The old contents of the bap_main.ml (the former frontend) moved to the plugins folder, namely the `disassemble` plugin. It is mostly the same code with the same set of features, except that the old `source-type` option is no longer provided. This functionality is now provided by a much more powerful `raw` loader.

# The `raw` loader 

This new loader is for parsing unstructured (raw) files or files whose structure is not known. It is a more versatile substitution of the old `source-type` option from the old `bap` frontend. Basically, now to load a pure code binary is as easy as
```
bap ./code --loader=raw
```

The `raw` plugin accepts a few command-line options, and it is possible to specify the architecture, the starting offset, the base address, and the size, that basically turns the `raw` loader into a swiss-knife manual file opener, with which you can specify all the necessary parameters of the file or even extract small portions of the files. 

# The `mc` command.

The former `bap-mc` tool is now also implemented as a plugin that provides the `mc` command as an extension. For the sake of the backward compatibility, the `bap-mc` binary is still provided, but it just execs the `bap mc` command. The `mc` command is also more versatile and easier to use than the previous implementation. The interface is mostly a superset of the old interface, the most notable change is that now it is possible to specify the binary string directly as command-line arguments, e.g.,
```
bap mc --show-insn=asm --show-bil -- 00 f0
```
# A new `bap objdump` command

The `mc` command young brother (provided by the same `mc` plugin) is the new `objdump` command, which, essentially, is the same `mc` command except that the input is a file, e.g.,
```
bap objdump --show-insn=asm /bin/ls
```

It shares the same command-line interface (modulo the input of course) as `mc`. It is especially nice to use it together with the new `raw` loader, e.g., to extract a portion of a file and disassemble it using the linear sweep disassembler,

```
bap objdump --show-insn=asm /bin/ls --raw-offset=640 --raw-base=0x400720 --raw-size=612
```

# The revamped Byteweight

Originally, we were intending to implement `bap-byteweight` as a command, the same as we did with `bap-mc`, but it turned out that it is better to implement it as a separate frontend. It is also a showcase of how it is easy to specify a new frontend and make it dependent only on stuff that it needs. Beyond the new look and fill, we implemented a new decision procedure, provided a richer interface to the Byteweight library itself, and Tries in general. 

The new decision procedure provides a more solid and mathematically kosher threshold. The main idea was that we are not using the prior information about the probability of a function start in a binary at all, even though this information is readily available for us. So we addressed this suboptimality by implementing the Bayes Factors decision procedure, that is a hypothesis testing method which compares the ratio of posterior probabilities of two competing hypotheses. Not only it gives now a more quantitative notion of the threshold but also makes the threshold itself adaptive, so that depending on individual statistics of each architecture/compiler the threshold will move accordingly. 

The old algorithm is also provided and could be chosen. (The new algorithm is enabled by default). We also provided the `--min-length` option that is the minimum length of a substring that we could use as a function start evidence. 

These two additions made the decision procedure much more accurate and drastically reduced the number of false positives and false negatives, at least on the tested samples.  We will return to Byteweight soon after the BAP 2.0 release to find out an optimal set of parameters, as well as to experiment with other decision procedures (there is still a lot of statistical knowledge about the binary and signatures that we're not employing). My original idea was to implement maximal a posteriori estimator, which will take into account individual probabilities of each prefix, but I didn't have enough time to fully finish this idea. 

# New  `baptop` toplevel

The `baptop` utility is no longer a simple bash script that runs utop, but instead is an OCaml program that uses UTop as a library and properly loads all dependencies. It is now much faster to start and is more portable. It could be seen as a yet another frontend, and originally the plan was to implement it as a subcommand to bap, e.g. `bap top`, but it turned to be impossible since `utop` and `compiler-libs` are using command-line arguments directly without giving us an option to override them. However, it was easy to embed BAP even into such an unfriendly environment. 
Remember, that `baptop` is basically the same as the `ocaml` utility, just a little bit enriched. So you can, for example, use it to run simple OCaml scripts that use `bap`, e.g.,

```
baptop script.ml
```

# The old `bap.top` library

For those of you who prefer the old `ocaml` toplevel, via emacs, or directly in your terminal (probably buffed with the `down` library), the good old `bap.top` library is still there so that `#require "bap.top"`  will load bap, initialize all plugins, and even install toplevel printers.

# A new bytecode frontend

The `bap` binary is an application compiled to the machine code, which is fast, but could be hard to debug, even though that OCaml's gdb support is getting better and better. So we decided to provide a `bap.byte` fronted, which is the same `bap` but compiled to the bytecode, which could be loaded into the OCaml debugger. 

# The recipe facility
We left the Recipe subsystem more or less intact. A lot of it was previously implemented in the bap frontend, so it was limited to `bap` utility only. Now, it is available as a standalone library `bap-recipe`  and the BAP Main library is capable of loading recipes, so it is now possible to use recipes in any frontend, and even when no command-line is available. 

The recipe plugin provides administration commands to manipulate recipes. We will later add more commands. To read the recipe documentation, do `bap recipe --help`. To list all recipes, do

```
bap print-recipes
```

# First-class readers and writers

The old frontends were manipulating Regular's readers and writers a lot, but the code wasn't really generic as it was hardcoding concrete types. So we decided to make the Regular library itself more flexible, and turn `Data.Read` and `Data.Write` type class instances into first-class values, which we can manipulate programmatically. 

# Special Thanks

I would like to thank Jay Bosamiya for starting this work back in #512, before even 1.0.0 was released :) This PR essentially finishes this work.


lib/bap_main/bap_main.mli
```diff
@@ -360,7 +360,7 @@ type ctxt
     it is not always an error, e.g., a user may have requested the
     help message, using the --help command).
 
-    If [init ()] terminates with any value other thatn [Ok ()] the
+    If [init ()] terminates with any value other that [Ok ()] the
```
ivg: it should be "than"

jaybosamiya:
Wow, this is _awesome_! This goes _well_ beyond anything I did for #512, so hats off for that :) Also, I love that things are now unified so nicely!

PR #1007, opened 2019-10-31
fixes issues in the oasis specification introduced by #1005

gitoleg:
There are many new libraries after #1005. And this PR fixes some of their's
oasis files in order to install bap from `opam`

### DIFF
--- a/oasis/recipe
+++ b/oasis/recipe
@@ -7,7 +7,7 @@ Library recipe
   Path: lib/bap_recipe
   FindlibName: bap-recipe
   CompiledObject: best
-  BuildDepends: base, stdio, parsexp, fileutils, camlzip, uuidm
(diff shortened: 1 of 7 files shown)


oasis/recipe
```diff
@@ -7,16 +7,6 @@ Library recipe
   Path: lib/bap_recipe
   FindlibName: bap-recipe
   CompiledObject: best
-  BuildDepends: base, stdio, parsexp, fileutils, camlzip, uuidm
+  BuildDepends: stdlib-shims, base, stdio, parsexp, fileutils, camlzip, uuidm
```
ivg: do we really need them?

ivg: yep, we need, I've used the `Stdlib` module accidentally, mostly thinking that it is overridden by Base, but apparently I was wrong. Since we already indirectly depend on those shims, let's keep them. However, after BAP 2.0 will be finally released, we will drop them along with the older versions of the compiler and core_kernel.

PR #1008, opened 2019-11-04
restores tags naming

gitoleg:
In BAP 1.x the function `Tag.name` returned the name that was provided as an argument to the `name` parameter to the `register` function, which is not guaranteed to be unique, but is more human-readable.

In BAP 2.0 after values were reimplemented in terms of the Knowledge value, the identifier returned by the `Tag.name` function became the fully-qualified name with namespace included, which is less readable and, more importantly, breaks the `map-terms` facility on which we rely a lot on our IDA integration.

This PR restores the old behavior.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -4197,10 +4197,6 @@ module Std : sig
           functions will be used to print, compare and serialize
           values.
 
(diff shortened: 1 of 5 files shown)


ivg:
subsumed by #1010

Issue #967, opened 2019-08-08
files are leaking in the bap cache


ivg:
When a cache entry is overwritten with new data the old file is not deleted and is no longer reachable from the index, therefore it will be never deleted with `--cache-clean` or evicted, that will lead to the hard drive memory leak.

ivg:
fixed in #1013

Issue #966, opened 2019-08-08
improve performance of the cache plugin


ivg:
The cache plugin is searching for the index file by enumerating all files in the cache (which usually is an order of thousands), given that the optimization plugin is accessing cache for each subroutine in a binary it may end up in a severe performance drop (which was reported by some of our users)

ivg:
This is still a big issue so that I even have to disable cache as in certain cases the cache significantly reduces BAP's performance, i.e., when I am enabling optimization and analyzing binaries with thousands of functions.

The hard requirement is that our cache should work (i.e., increase the performance) when there are thousands of entries in it (the upper bound should be no less than 100k entries). 

We need to benchmark it carefully, but so far the main culprit is that each cache access takes the linear time in the number of cache entries. This is because we perform cache garbage collection on each access (even worse, we do it twice). To mitigate it (after we confirm that this is the main issue with benchmarking), we will apply several strategies summarized below.

## Delay the Cache Garbage Collector

Delay the cache GC, so that it will run independently on the number of entries, this will essentially make cache insertion/query functions O(1). We should, of course, ensure that GC is not run after each update. The suggested criterium is the total size, once it reaches a certain threshold we shall run the GC and evict several entries at once (until we reach the desired level). Preliminary, I suggest using the `cache-size` parameter and add `cache-overhead` which is a percentage of the original size that we can leak. Therefore, the threshold will be `size + size * overhead` and each GC cycle shall shrink the cache below `size` (i.e., remove `size*overhead` worth of entries). 


## Add an option to disable GC

Add a separate command to run the cache GC as well as an option to disable it at all. A possible scenario for the latter would be running BAP with a cache located on some temporary partition with limited size and let the operating system to stop caching from work, once it is filled in. 

## Cache the cache

We don't really need to read the index file on each access as we can cache it and read it only if it was updated. This will save us time on locking/reading in exchange for one attribute reading. Of course, when we save an entry, we shall flush it immediately. It should be easy to implement and easy to benchmark. Will do merge this only if it will give us a significant improvement in time, e.g., if reading a thousand of optimization data entries with 100k entries in the cache, will be significantly (50-100%) faster than without caching. 

## Make the cache hierarchical

This approach should be investigated before implementation. It is not clear if it will give us any benefits. 

[...]

PR #1013, opened 2019-11-11
fixes the cache plugin

gitoleg:
This PR resolves a couple of long-standing issues with the cache plugin
and provides a more efficient procedure for finding the cache index
(see #966) and prevents the leakage of entries in the case when the same
entry is added twice (see #967).

### DIFF
--- a/plugins/cache/cache_main.ml
+++ b/plugins/cache/cache_main.ml
@@ -25,6 +25,8 @@ let (/) = Filename.concat
 
 module Index = struct
   let index_version = 2
+  let supported_versions = [2;1]
+
   let index_file = sprintf "index.%d" index_version
 
   let is_index path =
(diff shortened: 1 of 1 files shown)


plugins/cache/cache_main.ml
```diff
@@ -247,12 +256,15 @@ let create reader writer =
...
         Out_channel.close ch;
-        {
-          index with
-          entries = Map.set index.entries ~key:id ~data:{
-              size = size path; path; atime = ctime; ctime; hits = 1
-            }
-        }) in
+        let entry = {
+          size = size path; path; atime = ctime; ctime; hits = 1
+        } in
```
ivg: I don't like performing a removal as a side-effect of  a pure function. It is better to express it explicitly, e.g.,
```ocaml
if Map.mem index.entries id then begin
   warning "overwriting an entry";
   Index.remove_entry e;
end
```
Also, in general, this PR doesn't follow the spirit of the implementation, where the entries are removed all in one place. This is assumed in the code so I'm not sure whether this change breaks this assumptions anywhere.

PR #1020, opened 2019-11-14
Restore the Boeing dignity

XVilka:
I think you have to clarify which Korea was sponsoring the development. Otherwise you might come into the trouble ;)


ivg:
We also have a surprising test failure in ` ko/arm-can-bcm.ko` test. I wonder if it has anything to do with Korea) But jokes aside, we have to investigate it.

ivg:
superseded by #1021 

Thanks :)

PR #1028, opened 2019-12-04
a few hacks that improve the build time

ivg:
Building the whole BAP repository from scratch is now more than twice
as fast and takes about 8 minutes or even less with faster CPUs and
SSD drives.

What is more important, rebuilding BAP now takes less than a minute
even if something was changed deep in the BAP's core, like
disassembler or even regular or bap-types.

Below is the list of tricks that made this possible.

1) Disable -short-paths. This option was responsible for nearly x2
slowdown. Since nowadays we mostly do type-checking with merlin having
short paths doesn't really matter anymore. So screw them. Short paths
are also disabled in bapbuild.

2) Disable cross-module optimizations. When cross-module optimizations
are enabled (by default) every time a module changes all dependent
modules are recompiled. This behavior sort of defeats the purpose of
separate compilation and leads to 20 minutes of recompilation when
some module in disassembler is changed. Now it won't happen and when
any module is changed no other modules will be recompiled. If the
interface is changed only the modules that import this interface will
be recompiled. Theoretically, it may lead to less optimal builds. We
didn't notice any slowdown yet, but we may consider putting this
particular trick under a compilation flag and disable during releases.

3) Hotfix ocamlbuild linking rule. There is a bug in ocamlbuild that
prevents caching from working properly. The rule for linking library
mentions .so as its product, but it never builds it (and can't (and
shouldn't)). As a result, the caching system notices a missing object
and triggers relinking of all libraries. Ideally, it should be fixed
on the upstream (I will PR to them) but we don't want to wait for
upstream, especially since the fix is simple - we just override the
rule by our own, that didn't mention any .so files.

4) Disables compression when plugins are packed. Plugins are zip files
underneath the hood and when we pack and unpack them it takes time to
compress them. It goes much faster (especially for custom llvm builds
with debugging symbols) when the compression is disabled. This trick
is also a candidate to be made optional.

5) Packs a plugin in one step. Instead of doing 4 packings/repackings
we now do it only once.

6) Unlimits the number of jobs tried by ocamlbuild. This may break on
Travis, but let's try. We all know that ocamlbuild is not very good in
parallelizing tasks, but it tries its best. I didn't see more than 12
jobs at once, so it should be all right.


ivg:
For the reference, we can remove our own cmxs rule as soon as this issue [is resolved][1] in the upstream. 
[1]: https://github.com/ocaml/ocamlbuild/issues/304

PR #1032, opened 2020-01-22
fixes taint garbage collector

gitoleg:
The `finished` observation may not reach a receiver, because of unspecified order in which primus components are executed. E.g. greedy scheduler can intercept `finished` observation and kill a machine. As a result, no taint is ever finalized, which hits any analysis that depends on it. And this PR introduces taint finalization on the `halt` observation, which happens before `finished`.

### DIFF
--- a/lib/bap_taint/bap_taint.ml
+++ b/lib/bap_taint/bap_taint.ml
@@ -398,7 +398,7 @@ module Gc = struct
         ~f:(finish ~live:false) >>= fun () ->
       Machine.Local.put gc {old = cur}
 
-    let finalize _ =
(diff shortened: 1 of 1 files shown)


lib/bap_taint/bap_taint.ml
```diff
@@ -398,14 +398,14 @@ module Gc = struct
         ~f:(finish ~live:false) >>= fun () ->
       Machine.Local.put gc {old = cur}
 
-    let finalize () =
+    let finalize _ =
```
ivg: Revert this change.

PR #1033, opened 2020-01-22
adds bulk memory copying operation to Primus

gitoleg:
The current `memset` works quite straightforward and for every byte in the
requested memory it:
1) casts a value to the size of a byte
2) makes an observation before storing
3) stores the byte
4) makes an observation after storing.

The problem is that in the primus-promiscuous mode we often can't aford this number of primitive operations for every single byte as the length could be a random value, and a primus machine reaches its limits very soon that makes any block after such `memset` unreachable.

As a solution, we add a new function (namely `fill`) into the primus interpreter, that will interact with the underlying memory without intermediate interpreter's calls, meaning that the number of expressions
we spend on `memset` is equal to zero.

A couple of observations is added as well, like usually: `filling` and `filled`

### DIFF
--- a/lib/bap_primus/bap_primus.mli
+++ b/lib/bap_primus/bap_primus.mli
@@ -797,6 +797,17 @@ module Std : sig
           the address [addr] *)
       val stored : (value * value) observation
 
+      (** [filling (addr,len,value) happens before the memory [addr, addr + len)
+          is filled with [value]
+
+          @since 2.1.0 *)
+      val filling : (value * int * int) observation
+
+      (** [filled (addr,len,value) happens after the memory [addr, addr + len)
+          is filled with [value]
+
+          @since 2.1.0*)
+      val filled : (value * int * int) observation
 
       (** [reading x] happens before the variable [x] is read from the
           environment. *)
@@ -1071,6 +1082,18 @@ module Std : sig
             invoked and the operation is repeated. Otherwise the
             [Segmentation_fault] machine exception is raised.  *)
         val store : value -> value -> endian -> size -> unit m
+
+        (** [fill a len x] fills [len] bytes of the memory starting at
+            addres [a] with value [x]
+
+            If [a] is not mapped or not writable then the pagefault
+            trap is invoked. If the handler is provided, then it is
+            invoked and the operation is repeated. Otherwise the
+            [Segmentation_fault] machine exception is raised.
+
+            @since 2.1.0 *)
+        val fill : value -> len:int -> int -> unit m
+
       end
     end
 
--- a/lib/bap_primus/bap_primus_memory.ml
+++ b/lib/bap_primus/bap_primus_memory.ml
@@ -304,4 +304,18 @@ module Make(Machine : Machine) = struct
     find_layer addr layers |>
     function Some {perms={readonly}} -> not readonly
            | None -> false
+
+  let fill base ~len value =
+    let filter_out values region =
+      Map.filteri values
+        ~f:(fun ~key:addr ~data:_ ->  not (inside region addr)) in
+    get_curr >>= fun {layers;values} ->
+    is_writable base >>= fun is_writable ->
+    if is_writable then
+      let mem = { base; len; value = Generator.static value } in
+      let values = filter_out values mem in
+      let perms = {executable=false; readonly=false} in
+      put_curr {values; layers = {perms; mem = Dynamic mem} :: layers; }
+    else pagefault base
+
 end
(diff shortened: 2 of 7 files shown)


lib/bap_primus/bap_primus.mli
```diff
@@ -797,6 +797,17 @@ module Std : sig
           the address [addr] *)
       val stored : (value * value) observation
 
+      (** [filling (addr,len,value) happens before the memory [addr, addr + len)
+          is filled with [value]
+
+          @since 2.1.0 *)
+      val filling : (value * int * int) observation
```
ivg: it should be `value * value * value`

We need to track the origin of length and, even more important, the origin of the value which we fill (to propagate taint).

lib/bap_primus/bap_primus.mli
```diff
@@ -797,6 +797,17 @@ module Std : sig
...
+
+      (** [filled (addr,len,value) happens after the memory [addr, addr + len)
+          is filled with [value]
+
+          @since 2.1.0*)
+      val filled : (value * int * int) observation
```
ivg: same as with `fill`

lib/bap_primus/bap_primus.mli
```diff
@@ -1071,6 +1082,18 @@ module Std : sig
...
+        (** [fill a len x] fills [len] bytes of the memory starting at
+            addres [a] with value [x]
+
+            If [a] is not mapped or not writable then the pagefault
+            trap is invoked. If the handler is provided, then it is
+            invoked and the operation is repeated. Otherwise the
+            [Segmentation_fault] machine exception is raised.
+
+            @since 2.1.0 *)
+        val fill : value -> len:int -> int -> unit m
```
ivg: values

lib/bap_primus/bap_primus_memory.ml
```diff
@@ -304,4 +304,18 @@ module Make(Machine : Machine) = struct
     find_layer addr layers |>
     function Some {perms={readonly}} -> not readonly
            | None -> false
+
+  let fill base ~len value =
+    let filter_out values region =
```
ivg: The word `filter` already assumes `out`, as there is no `filter_in` :) If you have trouble remembering what "filter" is doing, as I am, then just pick a better name that doesn't involve filter, e.g., `remove_all_mapped`.

lib/bap_primus/bap_primus_memory.mli
```diff
@@ -47,4 +47,7 @@ module Make(Machine : Machine) : sig
   val is_mapped : addr -> bool Machine.t
 
   val is_writable : addr -> bool Machine.t
+
+  val fill : addr -> len:int -> int -> unit Machine.t
```
ivg: that signature is correct

plugins/primus_lisp/lisp/string.lisp
```diff
@@ -99,12 +99,9 @@
...
 (defun memset (p c n)
   (declare (external "memset"))
-  (let ((p p))
-    (while n
-      (memory-write p c)
-      (incr p)
-      (decr n)))
-  p)
+  (memory-fill p n c)
+  (+ p n))
```
ivg: `memset` returns the `p`

plugins/primus_lisp/primus_lisp_primitives.ml
```diff
@@ -92,6 +92,28 @@ module MemoryAllocate(Machine : Primus.Machine.S) = struct
...
+  let run = function
+    | [addr; len; value] ->
+       let len = Value.to_word len |> Word.to_int in
+       let value = Value.to_word value |> Word.to_int in
+       if Result.is_error len || Result.is_error value
+       then negone
+       else
+         let len = Or_error.ok_exn len in
+         let value = Or_error.ok_exn value in
+         Eval.fill addr len value >>= fun () ->
```
ivg: remove all this error handing, as we should use values here. 
Also, this error handling is undocumented and is not done correctly...

plugins/primus_lisp/primus_lisp_primitives.ml
```diff
@@ -300,6 +322,8 @@ module Primitives(Machine : Primus.Machine.S) = struct
         "(memory-read A) loads one byte from the address A";
       def "memory-write" (tuple [int; byte] @-> int) (module MemoryWrite)
         "(memory-write A X) stores by X to A";
+      def "memory-fill" (tuple [int; int; byte] @-> int) (module MemoryFill)
+        "(memory-fill A L X) fills every byte from A to A + L with X";
```
ivg: the type signature is made incorrect, as you return an 8-bit value (aka byte), not a word. But don't fix the signature, fix the implementation - let it return `A` and update the documentation to reflect this.

ivg:
closing this now as it clearly stalled.

PR #1034, opened 2020-01-22
fixes arguments attributes in the `callsites` plugin

gitoleg:
When attributes are assigned to a newly created synthetic argument term
in the `callsites` plugin, we don't take into consideration attributes
from the original argument term. This PR fixes this and now all the
attributes from origin argument term are assigned to the new one.

### DIFF
--- a/plugins/callsites/callsites_main.ml
+++ b/plugins/callsites/callsites_main.ml
@@ -30,11 +30,12 @@ let transfer_attr attr t1 t2 =
   | None -> t2
   | Some v -> Term.set_attr t2 attr v
 
-let transfer_attrs t1 t2 =
-  let t2 = Term.set_attr t2 Term.synthetic () in
-  Term.set_attr t2 Term.origin (Term.tid t1) |>
-  transfer_attr Disasm.insn t1 |>
-  transfer_attr address t1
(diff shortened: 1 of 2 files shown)


plugins/callsites/callsites_main.ml
```diff
@@ -30,11 +30,12 @@ let transfer_attr attr t1 t2 =
   | None -> t2
   | Some v -> Term.set_attr t2 attr v
 
-let transfer_attrs t1 t2 =
-  let t2 = Term.set_attr t2 Term.synthetic () in
-  Term.set_attr t2 Term.origin (Term.tid t1) |>
-  transfer_attr Disasm.insn t1 |>
-  transfer_attr address t1
+let transfer_attrs origin_arg call arg =
```
ivg: The `arg` argument is not an `arg` it is a `def`.

plugins/callsites/callsites_main.ml
```diff
@@ -30,11 +30,12 @@ let transfer_attr attr t1 t2 =
...
+  let arg = Term.with_attrs arg (Term.attrs origin_arg) in
```
ivg: move this code to the `def_of_arg` function, leave everything else intact.

PR #1035, opened 2020-01-22
adds a stub for realloc

gitoleg:
We add a stub for `realloc` - surprisingly many allocations are done by passing a null pointer to `realloc`, and an analysis that is based on a tracing of allocation/freeing memory won't notice such call.

We also dropped unresolved_call linking from the promiscuous mode as it was interfering with our stack handling routines (it was preventing them from being triggered).

Also, we add a new behavior of the `free` call processing in `memcheck-malloc`:  now, if the pointer being freed in unknown, i.e. there wasn't any memory allocation for it in the current primus machine, then we allocate one-byte length region and then release it. The justification is simple: by doing so, we are able to find more bugs like use-after-free or double-free, e.g. in some library, where any path we take won't necessarily contain the chain of calls like `malloc->free->memory-access/free` because such path exists only in the user code but not in the library itself.
ivg: however so far this feature produced only false positives, probably due to some unwanted aliasing between the pointer and pointee. Will keep in master for some time and see if it really a good idea.


plugins/primus_lisp/lisp/simple-memory-allocator.lisp
```diff
@@ -43,8 +43,18 @@
         (if failed 0
           (set brk (+ brk n))
           (malloc/fill-edges ptr n)
-          (+ ptr *malloc-guard-edges*))))))
-
+          (let ((ptr (+ ptr *malloc-guard-edges*)))
```
ivg: Dictionaries are not a part of the program state. We use dictionaries as storage for various analyses nothing more. Therefore, the semantics of a program shall depend only of the program state, which is fully described by `Env` and `Memory` components.

plugins/primus_lisp/lisp/simple-memory-allocator.lisp
```diff
@@ -29,22 +29,45 @@
 
 (defun memory/allocate (ptr len)
   (if *malloc-initialize-memory*
-      (memory-allocate ptr n *malloc-initial-value*)
-    (memory-allocate ptr n)))
+      (memory-allocate ptr len *malloc-initial-value*)
+    (memory-allocate ptr len)))
+
+(defun word-size () (/ (word-width) 8))
```
ivg: adding this function to the public interface will introduce confusion, between `word-size` and `word-width` so, please either remove it or mangle correspondingly.

plugins/primus_lisp/lisp/simple-memory-allocator.lisp
```diff
@@ -29,22 +29,45 @@
...
-          (malloc/fill-edges ptr n)
-          (+ ptr *malloc-guard-edges*))))))
-
+          (set brk (+ brk full))
+          (malloc/fill-edges ptr full)
+          (set ptr (+ ptr *malloc-guard-edges*))
+          (encode-memory-length ptr n)
+          (+ ptr width))))))
+
+(defun realloc (ptr new)
```
ivg: Please, rewrite this function to conform with [POSIX][1]. In particular,
1. when `ptr` is NULL it shall behave as malloc
2. when `len` is NULL it shall behave as free
3. when `len` is unchanged nothing should be done
4. when `len` is smaller the pointer shall not change, but see a not below.
5. only when `len` is greater than the `old-len` we may return a new pointer (and we don't have other choices with our current allocator).
6. if a new chunk is allocated by realloc the old chunk, pointed by `ptr` shall be deallocated.
The following code (untested) implements this semantics. Note, how `(or (and ..) (and ..) ...)` is used in place where you will be using pattern matching in OCaml. This is much easier to read than a bunch of nested if/then/elses. 


```lisp
(defun realloc (ptr len)
  (or (and (not ptr) (malloc ptr len))
      (and (not len) (free ptr))
      (let ((old-len (malloc/get-chunk-size ptr)))
        (if (<= old-len len)
            (prog (malloc/put-chunk-size ptr len) ptr)
          (free ptr)
          (let (new-ptr (malloc ptr len))
            (if new-ptr (memcpy new-ptr ptr old-len)
              new-ptr))))))
```

Note, if `realloc` is used to reduce the size of a chunk (e.g., for shrinking a dynamic vector), we shall return the same pointer, but mark the region `ptr+len,ptr+old-len` as invalid. Therefore, we need to update the region-analysis (where malloc is hooked into it, and track calls to realloc which shrink the regions and try to mark them as freed). If it would be too hard, we can ignore it right now and think about it later.

[1]:  https://pubs.opengroup.org/onlinepubs/009695399/functions/realloc.html

plugins/primus_test/lisp/memcheck-malloc.lisp
```diff
@@ -3,6 +3,9 @@
 (defmethod call (name ptr)
   (when (and ptr (= name 'free)
              (not (= ptr *malloc-zero-sentinel*)))
+    (when (and ptr
```
ivg: `ptr` is already non-NULL here, see the guard above. Also for `when (not cond)` there is <strike>`unless`</strike> `until cond`. Ooops, it should be, of course, `unless`, not `until` and not even `unit` (as it is documented) :)  Could you please add `unless` in addition to (the now deprecated) `until` to the `init.lisp`? And, please, update the docs correspondingly, and mark `until` as deprecated.

So the code will be now much more readable:
```lisp
(defmethod call (name ptr)
  (when (and ptr (= name 'free)
             (not (= ptr *malloc-zero-sentinel*)))
    (unless (memcheck-is-allocated 'malloc ptr)
      (memcheck-acquire 'malloc ptr 1))
    (memcheck-release 'malloc ptr)))

```

plugins/primus_test/lisp/memcheck.lisp
```diff
@@ -24,6 +24,12 @@
 (require incident)
 
 ;; Public Interface
+
+(defun memcheck-is-allocated (heap ptr)
```
ivg: I would name it `memcheck-is/was-allocated` or even better, `memcheck-is-tracked` to prevent future confusion.

plugins/primus_lisp/lisp/simple-memory-allocator.lisp
ivg: This `realloc` is also buggy, by the way, the trick `(or (and <cond> <body>) ...)` doesn't work here, since `<body>` in our case could also evaluate to false so the control flow will fall down, i.e., both `malloc` and `free` could return false, in fact `free` always returns false, so we will fall after free to realloc. It won't break a lot, i.e., we will now hit the `(<= old-len len)` guard, and just write `0` to the length field. But still it is more a bug than a feature.

plugins/primus_test/lisp/memcheck-malloc.lisp
ivg: Ok, I was really confused yesterday :) `unless` is not `until`. `unless` is a dual for `while`, where `until` is dual `when`. So here are the definitions that we should have,

```lisp
(defmacro when (cnd body)
  "(when CND BODY) if CND evaluates to true, the BODY is evaluated and
   the value of the last expression BODY becomes the value of the
   whole expression. Otherwise, if CND evaluates to false, nil is returned."
  (if cnd (prog body) ()))

(defmacro unless (cnd body)
  "(unless CND BODY) if CND is evaluates to true, then BODY is evaluated and
   the value of the last expression in BODY becomes the value of the
   whole expression. Otherwise, if CND evaluates to false, nil is returned."
  (if (not cnd) () body))

(defmacro until (c b)
  "(until COND BODY) if COND evaluates to false then the whole expression
    evaluates to nil and BODY is not evaluated. Otherwise, BODY is evaluated
    until COND evaluates to true and returns the value of the last evaluation
    of BODY."
  (while (not c) b))
```

ivg: Ok, I was really confused yesterday :) `unless` is not `until`. `unless` is a dual for `when`, where `until` is dual `while`. So here are the definitions that we should have,
```lisp
(defmacro when (cnd body)
  "(when CND BODY) if CND evaluates to true, then BODY is evaluated and
   the value of the last expression in BODY becomes the value of the
   whole expression. Otherwise, if CND evaluates to false, nil is returned."
  (if cnd (prog body) ()))

(defmacro unless (cnd body)
  "(unless CND BODY) if CND evaluates to false, then BODY is evaluated and
   the value of the last expression in BODY becomes the value of the
   whole expression. Otherwise, if CND evaluates to true, nil is returned."
  (if (not cnd) () body))

(defmacro until (c b)
  "(until COND BODY) if COND evaluates to true, then the whole expression
   evaluates to nil and BODY is not evaluated. Otherwise, if COND evaluates
   to false, then BODY is evaluated until COND evaluates to true and the value
   of the last evaluation of BODY becomes the value of the whole expression."
  (while (not c) b))
```

plugins/primus_lisp/lisp/simple-memory-allocator.lisp
```diff
@@ -29,22 +29,56 @@
...
-          (+ ptr *malloc-guard-edges*))))))
-
+          (set brk (+ brk chunk-size))
+          (malloc/fill-edges ptr chunk-size)
+          (set ptr (+ ptr *malloc-guard-edges*))
+          (malloc/put-chunk-size ptr n)
+          (+ ptr header-size))))))
+
+
+(defun realloc/update-chunk-size (ptr len)
```
ivg: rename it to `realloc/shrink-chunk`

plugins/primus_test/lisp/memcheck-malloc.lisp
```diff
@@ -3,6 +3,8 @@
 (defmethod call (name ptr)
   (when (and ptr (= name 'free)
              (not (= ptr *malloc-zero-sentinel*)))
+    (unless (memcheck-is-tracked 'malloc ptr)
```
ivg: For so far unknown to me reason this hack produces false positives. Maybe it is just my inputs, maybe this is something wrong with this idea (by the way, instead of `1` we shall use a parameter here). Anyway, let's merge this into master experiment more with it to see if really a good or bad idea.

PR #1036, opened 2020-01-22
adds the stub resolver abi pass

gitoleg:
The stub resolver pass redirects on the IR level calls to stub functions to the calls to real functions if such could be identified. A function is a stub if it has the stub attribute present in the knowledge base.  So far, we add only a knowledge provider for ELF stubs (that looks into the .plt section) others will be added later. The redirection algorithm establishes a one-to-one mapping between a stub and its prospective redirection. In case if a non-bijective mapping is detected (a stub has more than one implementation candidates or implementation has more than one stubs) no redirection is established.


plugins/primus_x86/primus_x86_loader.ml
```diff
@@ -50,6 +42,65 @@ module Plt_jumps(Machine : Primus.Machine.S) = struct
...
+         | Goto _ | Int _ | Ret _ -> jmp
+         | Call c ->
+            match Call.target c with
+            | Indirect _ -> jmp
+            | Direct tid ->
+               match Map.find twins tid with
+               | None -> jmp
+               | Some tid' ->
+                  Jmp.with_kind jmp (Call (Call.with_target c (Direct tid')))
+    end)#run prog
```
ivg: This function is too big, split it into several functions, each having a readable name and well-defined semantics. 

Next, the overall idea of the function is flawed. For your code to work the following assumptions should hold. 
1. all functions have different names (this is enforced by the Program module abstraction, so it is fine);
2. there is a demangler that uses `@` to mangle names that were originally having the same name (it is a hidden assumption that is implementaion defined, we can't rely on it);
3. The original names never had `@` in it. (Doesn't hold at all);
4. A function can have at most one other function that originally had the same name, aka twin in your parlance. (Not true - it could be a set of twins)
5. If a function has a twin then one of this pair must be a stub and another must not (doesn't hold).

So, please, revisit your implementation without relying on these assumptions. My suggestion would be to go through all stubs and try to build a partial mapping between stubs and non-stubs, like your twins mapping. This mapping shall be purely injective. In other words, for each stub there should be at most one implementation, and for each implementation there should be at most one stub. That means that if for a given stub you can select more than one implementation then neither this stub nor this implementation should be taken into the final mapping. The same is for the other side -- if implementation could be selected for more than one stub, then they should also be disregarded. Putting it simple, given a bipartile graph of stubs and implementations, you shall take only those stubs that have _exactly_ one edge. 





[1]: https://en.wikipedia.org/wiki/Partition_of_a_set

plugins/primus_x86/primus_x86_main.ml
```diff
@@ -3,7 +3,8 @@ open Bap_primus.Std
 include Self()
 
 let init _ =
-  Primus.Machine.add_component (module Primus_x86_loader.Component);;
+  Primus.Machine.add_component (module Primus_x86_loader.Component);
```
ivg: a component named component?

ivg:
> Unfortunately, the implementation was wrong: we linked an unresolved call handler to the address of a jump in a `plt` entry, so actually we
> just call a handler instead, and no actual unresolved call happens, no observation made, and no steps to restore SP taken.

That is not true :) Observations had happened by that time, therefore the code was correct. Weneed to reevaluate the problem and confirm that observations indeed do not happen and if they not identify why. My personal bet is that they do happen :)

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+  else
+    Project.with_program proj (relink (Project.program proj) links)
+
+let () =
+  let _man = Config.manpage [
+      `S "DESCRIPTION";
+
+      `P "This plugin provides an abi pass that transforms a program
+          by substituting calls to stubs with calls to real
+          subroutines.";
```
ivg: ... subroutines, when they are present in the binary.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+  let links = Plt_resolver.run proj in
+  if Map.is_empty links then proj
+  else
+    Project.with_program proj (relink (Project.program proj) links)
+
+let () =
+  let _man = Config.manpage [
+      `S "DESCRIPTION";
+
+      `P "This plugin provides an abi pass that transforms a program
```
ivg: <strike>This plugin</strike>

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+          by substituting calls to stubs with calls to real
+          subroutines.";
+
+      `P "Thus, for stubs in .plt section, the correspondence is
```
ivg: reprhase

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+            Jmp.with_kind jmp (Call (Call.with_target c (Direct tid')))
+  end)#run prog
+
+let run proj =
+  let links = Plt_resolver.run proj in
+  if Map.is_empty links then proj
+  else
+    Project.with_program proj (relink (Project.program proj) links)
+
+let () =
```
ivg: don't use the old interface in the new plugins. Use Bap_main.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+      let stubs, subs = partition mem (Project.program proj) in
+      find_unambiguous_pairs subs stubs
+end
+
+let relink prog links =
+  (object
+    inherit Term.mapper
+
+    method! map_jmp jmp =
+      match Jmp.kind jmp with
```
ivg: don't use the old interface, use the `Jmp.alt` and `Jmp.resolve`.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+      | Goto _ | Int _ | Ret _ -> jmp
+      | Call c ->
+        match Call.target c with
+        | Indirect _ -> jmp
+        | Direct tid ->
+          match Map.find links tid with
+          | None -> jmp
+          | Some tid' ->
+            Jmp.with_kind jmp (Call (Call.with_target c (Direct tid')))
```
ivg: use the BAP 2.0 interface

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+      ~f:(fun ~key:name ~data:tid pairs ->
+          match Map.find syms name with
+          | None -> pairs
+          | Some tid' -> Map.add_multi pairs tid tid') |>
+    Map.filter_map ~f:(function
+        | [tid] -> Some tid
+        | _ -> None)
+
+  let run proj =
+    let mem = section_memory proj ".plt" in
```
ivg: This is abi specific and the name of the section is hardcoded and there is no way to extend this code without modification.

We have the `Sub.stub` attribute that we shall use instead and use the knowledge interface to provide this attribute in architecture/abi specific plugins.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+
+module Plt_resolver = struct
+
+  let memory_contains m sub =
+    match Term.get_attr sub address with
+    | None -> false
+    | Some addr -> Set.mem m addr
+
+  (* pre: all functions have different names
+     pre: there aren't any stubs with the same alias *)
```
ivg: the second precondition is not holding, devise an algorithm that doesn't require this.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+  let partition mem prog =
+    let add_aliases stubs stub =
+      let tid = Term.tid stub in
+      Set.fold ~init:stubs (find_aliases tid)
+        ~f:(fun stubs alias -> Map.set stubs alias tid) in
+    let add_sym syms s = Map.set syms (Sub.name s) (Term.tid s) in
+    let names = Map.empty (module String) in
```
ivg: `let empty`

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+    Term.to_sequence sub_t prog |>
+    Seq.fold ~init:(names,names) ~f:(fun (stubs, syms) sub ->
```
ivg: code that uses a pair with each constituent having the same type are prone to errors, use records for such cases.

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+module Plt_resolver = struct
+
+  let memory_contains m sub =
+    match Term.get_attr sub address with
+    | None -> false
+    | Some addr -> Set.mem m addr
+
+  (* pre: all functions have different names
+     pre: there aren't any stubs with the same alias *)
+  let partition mem prog =
```
ivg: here `mem` should be `stubs` but we shouldn't classify based on mem anyway

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,110 @@
...
+    let add_aliases stubs stub =
+      let tid = Term.tid stub in
+      Set.fold ~init:stubs (find_aliases tid)
+        ~f:(fun stubs alias -> Map.set stubs alias tid) in
```
ivg: you shall not use `Map.set` when you mean `Map.add`

plugins/stub_resolver/stub_resolver_main.ml
```diff
@@ -0,0 +1,148 @@
...
+    inherit Term.mapper
+
+    method! map_jmp jmp =
+      match Jmp.alt jmp with
+      | None -> jmp
+      | Some alt -> match Jmp.resolve alt with
+        | Second _ -> jmp
+        | First tid -> match Map.find links tid with
+          | Some (Some tid') ->
+            Jmp.reify
```
ivg, line 98: don't create a new jmp term as we will lose as a result al the attributes of the old term. Map the old term using the `with_alt` function. That is not exposed in the interface, so please expose it also (together with the `with_dst` function).

ivg:
Also, please add unit tests and function tests (with a .so that has functions with external and static linkage calling each other).

plugins/stub_resolver/stub_resolver_tests.ml
```diff
@@ -0,0 +1,219 @@
...
+        symbol "quebec";
+        symbol "romeo";
+        symbol "siera" ~is_stub:true ~aliases:["papa"];
+        symbol "tango" ~is_stub:true ~aliases:["quebec"];
+        symbol "uniform" ~is_stub:true ~aliases:["romeo"];
+
+      ]
+      ~expected:["siera",   "papa";
+                 "tango",   "quebec";
+                 "uniform", "romeo";];
```
ivg, line 218: We need cases where symbols have many aliases, some intersecting aliases and some not, but where there is more than two cases, e.g.,

```
test "several intersections" [
  stub, "a",  ["a1"; "a2"; "a3"];
  real, "a0", ["a"];
  stub, "b",  ["b1"; "b2"; "b3"];
  real, "b0", ["b"; "c"];
  real, "c",  ["d"];
  stub, "d",  [];
] ~expect:[
  "a", "a0";
]
```

and more tests like this.

plugins/stub_resolver/stub_resolver_tests.ml
```diff
@@ -0,0 +1,219 @@
...
+
+let symbol ?(is_stub=false) ?(aliases=[]) name =
+  name, {is_stub; aliases}
+
+let test name ~symbols ~expected =
+  name >:: run name symbols expected
+
+let suite = "stub-resolver" >::: [
+
+    test "simple case: we have pairs"
```
ivg, line 134: this notation is very heavy and makes it hard to define and understand the cases. Use a simple list of triples to define inputs. Also, long names distract the attention and increase the cognitive burden. 
E.g., the following is easier to read:

```
test "simple case: we have pairs" [
  real, "a", [];
  stub, "b", ["a"];
] ~expect:[
  "a", "b";
]
```

ivg:
Ok, it's good to go. Merge as soon as Travis greenlights!

PR #1039, opened 2020-01-28
adds omake as an optional compilation backend

ivg:
# Introduction
OMake is one of the compilation backends for OASIS. It is extremely fast and on my machine with 10 cores builds BAP 10 times faster (in two minutes from scratch on mine 2Gh machine). It is also extremely fast on rebuilding and can redeploy BAP after a deep change in the matter of milliseconds. It also has a lot of interesting features, like incremental building, when an omake server is polling for changes in the file system and triggers rebuild as soon as you hit save. OMake also watches for the changes in dependencies outside of the source tree, so if you have any external library updated it will notice it and correctly rebuild the dependencies (no more "inconsistent assumptions"). 

# Caveats and workarounds

There are a couple of caveats though, that is why it is not our default build system. First of all, for omake backend to work, we had to submit a few patches to OASIS. While they are pending, we need to use our own fork:

    opam pin add oasis git@github.com:BinaryAnalysisPlatform/oasis.git 

Another issue, is that OMake backend doesn't support OASIS variables and conditional compilation. This immediatelly bars omake as a default backend, since we rely on the configuration variables a lot. But for the development mode, when we build all targets in the repository, it is not a big deal. As a workaround, we turn all conditional compilation into unconditional and hardcode variables for LLVM and C++ (and any other that we might need in future) in corresponding OMakefiles.

For the same reason, we also hardcode opam as the default OCaml package manager, so this backend doesn't support opamless environemnts.

Another important limitation is that it is in general unsafe to use the omake backend with compilers that are earlier than 4.08. While I didn't have any problems, I can't guarantee that they do not exist :) In any case, we will soon drop 4.07 and everyone are encouraged to use 4.08 or later, especially if they are developing BAP and using omake for that.

There are also a couple of non-essential issues with omake (which are fully offsetted by its compilation speed). We will create a wiki page for omake, where we will accumulate knowledge about omake and share tips and tricks. 

# How to use it

Go to the root directory of the BAP repository and perform the following steps:

0. (optional) install BAP dependencies
```
opam pin add . -n
opam install bap --deps-only
opam pin remove bap
```
1. install the fixed version of oasis
```
opam pin add oasis git@github.com:BinaryAnalysisPlatform/oasis.git 
```
2. run the specially prepared configure script
```
./configure-omake
```
3. (optional) update LLVM flags in `lib/bap_llvm/OMakefile`. This step could be done before `2` or after, shouldn't matter.

4. everything else is as usual
```
make && make reinstall
```


ivg:
See https://github.com/BinaryAnalysisPlatform/bap/wiki/Experimental-OMake-backend for the future development of the omake backend

PR #1070, opened 2020-02-28
fixes the config and run tests

### LINKED ISSUE #1065: the config test destroys bap configuration
The config test (testsuite/bap.tests/config.exp) deletes the whole configuration folder with actual user configuration.

gitoleg:
This PR:
- preserves user configuration file if exists in the config test
- applies `primus-lisp-type-check` option in run test

fixes #1065

### DIFF
--- a/opam/opam
+++ b/opam/opam
@@ -31,7 +31,7 @@ depends: [
   "conf-bap-llvm" {>= "1.1"}
   "parsexp"
   "bap-signatures"
-  "result" {= "1.4.0"}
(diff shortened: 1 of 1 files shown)


opam/opam
```diff
@@ -31,6 +31,7 @@ depends: [
   "conf-bap-llvm" {>= "1.1"}
   "parsexp"
   "bap-signatures"
+  "result" {= "1.4"}
```
ivg: This is a temporal measure that we will revert on the next week

PR #1079, opened 2020-03-11
switches the docbuilder to odig

### LINKED ISSUE #981: Deprecate argot documentation
Now BAP supporting building documentation with [odoc](https://github.com/ocaml/odoc), I think it is safe to drop the [argot](https://github.com/ivg/argot) support, which last time was updated in 2012 (mainstream), and only receives fixes from you. Odoc is actively maintained, and simply has prettier output. I suggest to make it official documentation builder. Moreover, they are working on integration of it with [mdx](https://github.com/realworldocaml/mdx/) so the examples in the documentation can be built and tested. See https://github.com/ocaml/odoc/blob/bb539dbe6892f0c9d9c8a487b9af1dbddf4cb1c2/notes/testable_examples.md and https://github.com/ocaml/odoc/pull/303 for more information

gitoleg:
We have switched to odig in April 2019 but since then we were building the documentation manually. This PR updates our bapdoc.ml to use odig/odoc instead of argot, which we no longer going to support. This is a part of the work to enable automatic building and deployment of the BAP documentation on each merge.

fixes #981


tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+let mkdir path =
+  if not (Sys.file_exists path) then
+    Unix.mkdir path 0o770
+
+let run cmd =
+  let res = Sys.command cmd in
+  if res <> 0 then
+    failwith ("Command: '" ^ cmd ^ "' failed")
+
+let dedot name =
```
ivg: If you just want to chop the .Std suffix, then you can use `String.chop_suffix` if you want to keep Bap.Std untouched you can just match with `"Bap.Std"`. This will make your intentions much more clear. Also, this function should be named `dedot`, it should be `drop_std` or `shorten`.  



However, in general I don't want this function at all, it might beat us later when we will add more libraries. Just extend the libraries specification to include the module name (for the link) and the short name, e.g,
```
"monads", "Monads.Std", "Monads', "The Monads library";
```
You can use an empty string `""` as the default designator, that will mean: take the full name as the short name.

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+    <head>
+    <meta http-equiv="refresh" content="0; url='../odoc/%s/index.html'" />
+    </head>
+    <body></body>
+    </html> |} lib in
+  let name = sprintf "man3/%s.3.html" lib in
+  Out_channel.write_all name ~data:redirection
+
+let build_manual {man; help} tool =
+  let repair_links x =
```
ivg: this function is also misnamed, as it actually creates the manpage as the side-effect.

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+  match Sys.command @@ sprintf "%s >/dev/null" (help tool) with
+  | 0 ->
+    let output = sprintf "man1/%s.1.html" (man tool) in
+    run @@ repair_links output;
+    run @@ redirect_index output
+  | _ ->
+    eprintf "Warning: can't render manpage for %s\n" tool;
+    eprintf "Called as %S, got:\n" (help tool);
+    ignore(Sys.command @@ sprintf "%s >&2" (help tool));
+    eprintf "\n%!"
```
ivg: clean this function to make it more readable

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+    Plugins.list () |> List.iter ~f:(fun p ->
+        build_manual plugin (Plugin.name p))in
+  List.iter packages ~f:(fun lib -> man3_redirection lib)
+
+let odig_pkgs () =
+  let f = Filename.temp_file "odig" ".pkgs" in
+  run (sprintf "odig pkg > %s" f);
+  let pkgs =
+    In_channel.read_lines f |>
+    List.filter_map ~f:(fun s ->
```
ivg: refactor it into the `remove_unresolved_pkgs` or `is_resolved_pkg` if you will refactor only the predicate.

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+        match String.split s ~on:'?' with
+        | x :: _ -> Some (String.strip x)
+        | _ -> None) in
+  Sys.remove f;
+  pkgs
+
+let odig intro =
+  let pkgs' = odig_pkgs () in
```
ivg: `let pkgs_known_to_odig = ...`

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+
+let check_installed x =
+  if not (is_installed x) then
+    failwith (sprintf "%s is not installed\n" x)
+
+let check () =
+  check_installed "bap";
+  check_installed "man2html";
+  check_installed "odig";
+  check_installed "emacs"
```
ivg:
```
+ check_installed "sed"
```

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+
+let () =
+  check ();
+  generate_manual ();
+  hand_written_man ();
```
ivg:
```
install_handwritten_manpages ();
install_lisp_documentation ();
```

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+  man  = ident;
+  help = sprintf "%s --help=groff"
+}
+
+let plugin = {
+  man = sprintf "bap-plugin-%s";
+  help = sprintf "bap --%s-help=groff";
+}
+
+let man3_redirection lib =
```
ivg: `generate_man3_redirection`

tools/bapdoc.ml
```diff
@@ -0,0 +1,232 @@
...
+    ignore(Sys.command @@ sprintf "%s >&2" (help tool));
+    eprintf "\n%!"
+
+let generate_manual () =
+  let _frontends =
+    List.iter frontends ~f:(fun (p,_) -> build_manual program p) in
+  let _plugins =
+    Plugins.list () |> List.iter ~f:(fun p ->
+        build_manual plugin (Plugin.name p))in
+  List.iter packages ~f:(fun lib -> man3_redirection lib)
```
ivg: `List.iter package ~f:generate_man3_redirection`

PR #1082, opened 2020-03-17
disables llvm (< 8.0) disassembler for aarch64

### LINKED ISSUE #1081: aarch64 disassembler segfaults
```
bap --version
2.0.0

bap --llvm-version
6.0

# the following .o file is compiled to arm64
bap arm64_file.o -d
no signatures for the specified compiler and/or architecture
[1]    405427 segmentation fault (core dumped)  bap  -d
```

But if we use `--raw-arch=arm --loader=raw` instead, it does work correctly. The file is attached:
[hash-arm.o.zip](https://github.com/BinaryAnalysisPlatform/bap/files/4324129/hash-arm.o.zip)

gitoleg:
The LLVM disassembler is broken for aarch64 and segfaults with the NULL-pointer dereference (segmentation fault) on versions below 8.0. There is no clear mitigation (passing a non-NULL pointer for the symbolizer function results in unpredictable behavior) and given that aarch64 is not a first-tier architecture for us yet, we will just disable it for older versions of LLVM.

fixes #1081

### DIFF
--- a/lib/bap_llvm/llvm_disasm.cpp
+++ b/lib/bap_llvm/llvm_disasm.cpp
@@ -615,7 +615,8 @@ private:
     }
 };
 
-
+// See https://github.com/BinaryAnalysisPlatform/bap/issues/1081
+// for details
 bool is_error_prone_arch(const char *triple) {
     if (LLVM_VERSION_MAJOR < 8)
         return std::string(triple) == "aarch64";
--- a/lib/bap_llvm/llvm_disasm.cpp
+++ b/lib/bap_llvm/llvm_disasm.cpp
@@ -617,11 +617,10 @@ private:
 
 
 bool is_error_prone_arch(const char *triple) {
-#if LLVM_VERSION_MAJOR >= 4 && LLVM_VERSION_MAJOR < 8
-    return std::string(triple) == "aarch64";
-#else
-    return false;
-#endif
+    if (LLVM_VERSION_MAJOR < 8)
+        return std::string(triple) == "aarch64";
+    else
+        return false;
 }
 
 
--- a/lib/bap_llvm/llvm_disasm.cpp
+++ b/lib/bap_llvm/llvm_disasm.cpp
@@ -615,14 +615,30 @@ private:
     }
 };
 
+
+bool is_error_prone_arch(const char *triple) {
+#if LLVM_VERSION_MAJOR >= 4 && LLVM_VERSION_MAJOR < 8
+    return std::string(triple) == "aarch64";
+#else
+    return false;
+#endif
+}
+
+
 struct create_llvm_disassembler : disasm_factory {
     result<disassembler_interface>
     create(const char *triple, const char *cpu, int debug_level) {
-        auto llvm = llvm_disassembler::create(triple, cpu, debug_level);
         result<disassembler_interface> r;
-        r.dis = llvm.dis;
-        if (!r.dis)
-            r.err = llvm.err;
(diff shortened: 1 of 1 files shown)


lib/bap_llvm/llvm_disasm.cpp
```diff
@@ -615,14 +615,30 @@ class llvm_disassembler : public disassembler_interface {
     }
 };
 
+
+bool is_error_prone_arch(const char *triple) {
+#if LLVM_VERSION_MAJOR >= 4 && LLVM_VERSION_MAJOR < 8
```
ivg: don't use macro here. If anything could be done without a macro, it should be done without it.

gitoleg:
### Some details

When we create a llvm symbolizer (required for llvm disassembler), there are a number of callbacks that we can pass to its constructor. Given that we don't need any, we just pass null pointer instead and it works for our every-day targets like x86, arm or mips.

```
target->createMCSymbolizer(
                           triple,
                           nullptr, // getOpInfo
                           nullptr, // SymbolLookUp
                           nullptr, // DisInfo
                           &*ctx, move(rel_info)));
```
The reason it doesn't work for aarch64 in llvm 6.0 is that there is no check 
for the pointer SymbolLookUp, compare 6.0 and 9.0 versions:

```
lib/Target/AArch64/Disassembler/AArch64ExternalSymbolizer.cpp
...
bool IsBranch, uint64_t Offset, uint64_t InstSize) {
  struct LLVMOpInfo1 SymbolicOp; 
...
```
```
lib/Target/AArch64/Disassembler/AArch64ExternalSymbolizer.cpp
...
bool IsBranch, uint64_t Offset, uint64_t InstSize) {
  if (!SymbolLookUp)  // GOOD
     return false;
  struct LLVMOpInfo1 SymbolicOp;
...
```

But it's not the biggest problem so far. We can write a dummy callback,
that would just return NULL pointer, i.e., that will do nothing.
And it will work for aarch64 and llvm 6.0

Let's quote the documentation in `include/llvm-c/Disassembler.h` about the callback:
```
/**
 * The type for the symbol lookup function.  This may be called by the
 * disassembler for things like adding a comment for a PC plus a constant
 * offset load instruction to use a symbol name instead of a load address value.
 * It is passed the block information is saved when the disassembler context is
 * created and the ReferenceValue to look up as a symbol.  If no symbol is found
 * for the ReferenceValue NULL is returned.  The ReferenceType of the
 * instruction is passed indirectly as is the PC of the instruction in
 * ReferencePC.  If the output reference can be determined its type is returned
 * indirectly in ReferenceType along with ReferenceName if any, or that is set
 * to NULL.
 */
```
I.e. we can solve one problem with the unchecked return value, but there are even more problems with that! There are parameters that are passed by reference and they could also be set to NULL!
Like in the example below, ReferenceName is not checked, but used, and the return
value of the callback is just ignored:

```
lib/MC/MCDisassembler/MCExternalSymbolizer.cpp
...
if (SymbolLookUp) { // GOOD AGAIN
 ...
    const char *ReferenceName;
    (void)SymbolLookUp(DisInfo, Value, &ReferenceType, Address, &ReferenceName);
    if(ReferenceType == LLVMDisassembler_ReferenceType_Out_LitPool_SymAddr)
      cStream << "literal pool symbol address: " << ReferenceName;  // BAD: SEGFAULT
    
 ...
```
The problem is that is llvm 9.0 with the x86_64 target. And to fix this, we need to spend even more time than we spent so far to figure out in the end, that there are even more problems that we had so far.

PR #1090, opened 2020-04-07
Cache optimization

gitoleg:
Bap cache works fine in most cases but slows down when the cache size is very big (more than 100K entries).
For example, running bap with a binary with almost 2K subroutines, with optimization enabled and when the cache size is about 5Gb and contains about 100K entries takes a lot:
```
 |----------------|-------------|--------------|    Input:
 |                | first call  | second call  |    cache size:         ~5Gb
 |----------------|-------------|--------------|    cache entries:      ~100K
 |  --no-cache    |     134     |      134     |    subroutines:        ~2K
 |----------------|-------------|--------------|    optimization level:  1
 |  current bap   |     595     |      297     |
 |----------------|-------------|--------------|
```
In this specific case, i.e. with the big cache size, bap works faster without cache.

This PR introduces a completely new cache, that:
1) delays GC: now we run GC after cache size has reached `max_size + overhead * max_size` level and remove several entries at once to shrink cache size down to `max_size - overhead * max_size`. Also, we changed our eviction policy. Previously, we were removing less frequently used entries and had to update the cache index on each read access to keep track of each entry usage. Now we remove random entries ( with prioritizing large ones), so we don't have to store anything on entry reading. 
2) doesn't rely on the index file: we can use system calls to get all the information we need.
3) implements all operations in a lock-free manner

This PR particularly solves this problem (as described in #966).

Also, this PR updates the command-line interface and adds options to deal with GC.
 - `bap cache --disable-gc` - disables GC, so cache size won't be limited.
- `bap cache --run-gc` - run GC.
- `bap cache --size=N` - sets the capacity of the cached data to `N Mb`
- `bap cache --overhead=N` - sets the overhead to `N %`

And the table with the final results:
```
 |----------------|-------------|--------------|    Input:
 |                | first call  | second call  |    cache size:         ~5Gb
 |----------------|-------------|--------------|    cache entries:      ~100K
 |  --no-cache    |     134     |      134     |    subroutines:        ~2K
 |----------------|-------------|--------------|    optimization level:  1
 |  current bap   |     595     |      297     |
 |----------------|-------------|--------------|
 |  this PR       |     297     |      79      |
 |----------------|-------------|--------------|

```
As one can see, running bap for the first time with significant cache size is still slower than without cache. But now it's faster when we have some cached data to load!


plugins/cache/cache_gc.ml
```diff
@@ -0,0 +1,48 @@
...
+
+include Self()
+
+let remove_entry e =
+  try Sys.remove e.path
+  with exn ->
+    warning "unable to remove entry: %s" (Exn.to_string exn)
+
+let threshold c =
+  Int64.(c.max_size + of_float (to_float c.max_size *. c.overhead))
```
ivg, line 13: use integer arithmetic and represent overhead in percent.

plugins/cache/cache_main.ml
```diff
@@ -6,246 +6,120 @@ include Self()
...
+
+  let write index =
+    Index.write index;
+    update index
+
+  let reload () =
+    match !t with
+    | None -> read ()
+    | Some t ->
+      if t.mtime = mtime () then t.index
```
ivg, line 59: don't compare floats for equality, use `<` or `>` or `<=`, etc. And don't use polymorphic compare with floats.

plugins/cache/cache_main.ml
```diff
@@ -6,246 +6,120 @@ include Self()
...
+      let max_size = Int64.(size * 1024L * 1024L) in
+      {idx with config = {idx.config with max_size;}})
+
+let set_overhead overhead =
+  if Float.(overhead >= 0.0 && overhead < 1.0)
+  then
+    Global.update ~f:(fun idx ->
+        {idx with config = {idx.config with overhead}})
+  else
+    raise (Invalid_argument "Cache overhead should be in the range [0.0; 1.0) ")
```
ivg, line 101: No, it is totally fine for the overhead to be 1000% or whatever large number. There is no need to cap it with 100%

plugins/cache/cache_main.ml
```diff
@@ -6,246 +6,120 @@ include Self()
...
+
+let run_gc () =
+  Global.update ~f:(fun idx -> GC.run idx);
```
ivg, line 104: run_gc shall do GC even if the threshold is not reached.

plugins/cache/cache_main.ml
```diff
@@ -6,246 +6,120 @@ include Self()
...
   exit 0
 
+let disable_gc x =
+  Global.update ~f:(fun idx ->
+      {idx with config = {idx.config with gc_enabled = not x}})
+
+
+let limit_of_config c =
```
ivg, line 112: I've already seen this function :)

plugins/cache/cache_main.ml
```diff
@@ -298,16 +176,29 @@ let () =
...
-
     ] in
   let clean = Config.(flag "clean" ~doc:"Cleanup all caches") in
   let set_size = Config.(param (some int64) "size" ~docv:"N"
                            ~doc:"Set maximum total size of cached data to
                                  $(docv) MB. The option value will persist
                                  between different runs of the program") in
+  let set_overhead =
+    Config.(param (some float) "overhead"
+              ~doc:"Set overhead of maximum total size of cached data,
```
ivg, line 187: Controls the aggressiveness of the garbage collector.  The higher the number the more space will be wasted but the cache system will run faster. It is expressed as a percentage of the max-size parameter.

plugins/cache/cache_types.ml
```diff
@@ -0,0 +1,44 @@
+open Core_kernel
+open Regular.Std
+
+type entry = {
```
ivg, line 4: Given that our cache entry deteriorated to just a path (with probably unnecessary size, as we can get it from stat), I now have a question: why do we need index at all? 

Like the natural step forward would be to use the digest as the name of the file and rely on stat and readdir to perform cleanup. We also, don't need a global file lock anymore, so that different instances of bap could work in parallel without any congestion.

plugins/optimization/optimization_main.ml
```diff
@@ -133,13 +133,40 @@ let process_sub free can_touch sub =
   let sub', dead = loop Tid.Set.empty (Sub.ssa sub) in
   O.create dead sub'
 
+let digest_of_label = function
```
ivg, line 136: please, split it into a separate PR.

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,135 @@
...
+    try Sys.remove path
+    with exn ->
+      warning "unable to remove entry: %s" (Exn.to_string exn)
+
+  let file_size x = Unix.LargeFile.( (stat x).st_size )
+
+  let remove_entries is_entry min_length size_to_free =
+    let dir = cache_dir () in
+    let rec loop entries freed len =
+      if freed < size_to_free && len > min_length then
```
ivg, line 48: `len > min_length` is a loop invariant, move it away. 

Also, get rid of the fragile `is_entry` instead either populate the array with entries only or separate meta files and data files into separate directories, e.g.,

```
$CACHE_DIR/
 |
 +- config
 +- lock
 +- data
```
This setup will make it easier to clean the cache, by just dropping the `data` folder as well as enable a future upgrade to a tree representation (analogous to the git representation).

plugins/cache/bap_cache_utils.ml
```diff
@@ -0,0 +1,42 @@
+open Core_kernel
+
+module Filename = Caml.Filename
+
+let open_temp temp_dir =
```
ivg, line 5: this function is unsafe and has a race condition via TOCTOU, use `Filename.open_temp_file`

plugins/cache/bap_cache_utils.ml
```diff
@@ -0,0 +1,42 @@
...
+[@@warning "-D"]
+
+let to_file : type t.
+  (module Binable.S with type t = t) -> string -> t -> unit =
+  fun b file data ->
+  let module T = (val b) in
+  let dir = Filename.dirname file in
+  let tmp,fd = open_temp dir in
+  let size = T.bin_size_t data in
+  let () =
```
ivg, line 32: use `protect` for this and others try/with code for proper finalization.

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,94 @@
...
+    try
+      let idx = Utils.from_file (module Compatibility.V2) file in
+      let dir = cache_data () in
+      Map.iteri idx.entries ~f:(fun ~key ~data:{path} ->
+          rename path (dir // Data.Cache.Digest.to_string key))
+    with e ->
+      warning "can't read entries from index version 2: %s"
+        (Exn.to_string e)
+
+  let from_index () = match find_index () with
```
ivg, line 51: why is a function named `from_index` returns a value of type `unit`?

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,94 @@
...
+    | None -> ()
+    | Some file ->
+      Cfg.(write default);
+      match get_version file with
+      | Ok 2 ->
+        upgrade_from_index_v2 file;
+        Sys.remove file
+      | Ok ver ->
+        warning "can't read entries from index version %d" ver;
```
ivg, line 60: An unsupported version is basically the same as an unknown version, we can't upgrade from any, so either both should be a warning or both should be an error. I suggest turning any unsupported/unknown case into a warning, but cleanup the old index, so that this warning is not repeated every time.

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,94 @@
...
+
+end
+
+let size () =
+  let path = cache_data () in
+  Sys.readdir path |>
+  Array.fold ~init:0L ~f:(fun s f ->
+      try
+        let file = path // f in
+        Int64.(s + Unix.LargeFile.( (stat file).st_size ))
```
ivg, line 72: TOCTOU. By that time the file `f` may no longer exist and size with fail with an error.

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,94 @@
...
+let init_cache_dir () =
+  let data = Cfg.cache_data () in
+  let conf = Cfg.config_path () in
+  if not (dir_exists data)
+  then mkdir data;
+  if not (Sys.file_exists conf)
+  then Cfg.(write default)
+
+let init () =
+  init_cache_dir ();
```
ivg, line 93: The cache initialization is non-atomic. Create the directory structure in a temporary folder and then move to the required destination. If the destination already exists, treat it as a success.

plugins/cache/bap_cache.mli
```diff
@@ -0,0 +1,7 @@
+(** All operations in this module are atomic. *)
```
ivg, line 1: no operations in this module are atomic, neither init nor size :(

plugins/cache/bap_cache.ml
ivg: sorry, you already has try/catch here

plugins/cache/bap_cache.mli
```diff
@@ -0,0 +1,7 @@
...
+
+open Bap_cache_types
+
+val init : unit -> unit
```
ivg, line 5: init is not atomic.

plugins/cache/bap_cache_gc.ml
```diff
@@ -0,0 +1,154 @@
...
+
+   Notation:
+   1. s(i) - the size of i-th file, where i = 0..m-1 with m being the
+      total number of files;
+   2. Sum(x(i)) = x(0) + ... x(m-1) - is the sigma operator;
+   3. T = Sum(s(i)) - the total size of the cache;
+   4. p(i) = s(i)/T - the discrete probability distrubution of the file
+      sizes in cache, likelihood that a randomly chosen file from the
+      cache will have size s(i).
+   5. F(i) = p(i - 1) + p(i) cumulative discrete distribution function
```
ivg, line 14: `F(i) = p(i) + p(i-1) + ... + p(0)` not just the last two :)

plugins/cache/bap_cache_gc.ml
```diff
@@ -0,0 +1,154 @@
...
+      (CDF). F(i) we can generate a random number u in range 0..1,
+      using a uniform random number generator, and then find such k that
+      F(k-1) < u <= F(k).
+   6. |s| = Sum(p(i) * s(i)) = (1/T) * Sum(s(i)^2) - the expected value
+      of the size in cache
```
ivg, line 19: the expected value of the size of a cache entry.

plugins/cache/bap_cache_gc.ml
```diff
@@ -0,0 +1,154 @@
...
+  with _ -> None
+
+let read_cache path =
+  Sys.readdir path |> Array.filter_map ~f:(entry path)
+
+let total_size =
+  Array.fold ~init:0 ~f:(fun s e -> s + e.size)
+
+let cdf entries =
+  let prob e = max min_entry_size e.size in
```
ivg, line 106: please, let's not call it a probability, just do `pre + max min_entry_size size`

plugins/cache/bap_cache_gc.ml
```diff
@@ -0,0 +1,154 @@
...
+    else indexes in
+  let cdf = cdf entries in
+  loop cdf (Set.empty (module Int)) 0 |> Set.to_array
+
+let remove e =
+  try Sys.remove e.path
+  with exn ->
+    warning "unable to remove entry: %s" (Exn.to_string exn)
+
+let shuffle fs =
```
ivg, line 131: there is already a function for that, see `Array.permute`

plugins/cache/bap_cache_gc.mli
```diff
@@ -0,0 +1,4 @@
+
+val shrink : ?threshold:int64 -> upto:int64 -> unit -> unit
```
ivg, line 2: we always used megabytes to specify sizes in our user interface. Why we need to use bytes here?

plugins/cache/bap_cache_main.ml
```diff
@@ -0,0 +1,120 @@
+open Core_kernel
```
ivg, line 1: let's store and print everything in megabytes. 
The cdf function should still use kilobytes, but the main dimension for us should be megabytes.

plugins/cache/bap_cache_main.ml
```diff
@@ -0,0 +1,120 @@
...
+let update_config sz ov gc =
+  let set f x y = match x with
+    | None -> y
+    | Some x -> f x y in
+  let cfg = Cfg.read () in
+  let cfg' = set size sz cfg |> set overhead ov |> set disable_gc gc in
+  if cfg <> cfg'
+  then
+    let () = Cfg.write cfg' in
+    let () = printf "Config updated, exiting ...\n" in
```
ivg, line 81: use the semicolon or just begin/end instead of multiple `let ()`
also, it is the unix style not to write anything if everything is fine. You can use `info` and `debug` if you want to say something.

plugins/cache/bap_cache_main.ml
```diff
@@ -0,0 +1,120 @@
...
+    exit 0
+
+let () =
```
ivg, line 84: 1. Switch to the `Bap_main` interface
2. Use commands for those options that exit, but provide the backward compatibility. 

I want the following interface at the end:
```
bap cache clean # cleans the cache, but bap --cache-clean should also work
bap cache set-size 4096
bap cache set-overhead ...
bap cache run-gc
bap cache {enable,disable}-gc
bap cache info 
bap cache # prints all available commands
```

Also write the plugin documentation that will thoroughly describe how the cache works.

ivg: to make it clear, it is fine if we will not support the old options like `--cache-set-size` they are not very commonly used, but `bap --cache-clean` should work (just indicate in the documentation that it is deprecated)

plugins/cache/bap_cache.ml
```diff
@@ -0,0 +1,178 @@
...
+let rec mkdir path =
+  let par = Filename.dirname path in
+  if not (Sys.file_exists par) then mkdir par;
+  if not (Sys.file_exists path) then
+    ensure_dir_exists path
+
+let dir_exists dir = Sys.file_exists dir && Sys.is_directory dir
+
+let with_temp_dir path ~f =
+  let tmp = Filename.temp_file "tmp" "" in
```
ivg, line 82: This is very fragile, `temp_file @@ mkdir` is not an atomic operation, not even close, as there is no even a check that the folder/file exists. Therefore, it will be quite often when two or more different processes will try to create a folder with the same name. To make it more robust you shall do the following:
```
1. choose a new random name
2. create a folder with that name, if not successful goto 1.
```
The operating system will guarantee that the created folder, if successful was created by the current process.

plugins/cache/bap_cache.ml
```diff
@@ -78,14 +79,27 @@ let rec mkdir path =
 
 let dir_exists dir = Sys.file_exists dir && Sys.is_directory dir
 
+let mkdtemp ?(mode=0o0700) ?tmp_dir ?(prefix="") ?(suffix="") () =
+  let rng = Random.State.make_self_init () in
```
ivg, line 83: move the `rng` initialization to the toplevel

PR #1093, opened 2020-04-16
exports abi processors

aoidos-bin:
This adds 2 new functions to C.Abi, to be able to register a new abi processor
and then access it later.

The processors are stored in a mutable StringMap with the abi names as keys, using the function `C.Abi.register name abi`.
They can then be reached from any plugin using `C.Abi.get_processor name`.

### DIFF
--- a/lib/bap_c/bap_c_abi.ml
+++ b/lib/bap_c/bap_c_abi.ml
@@ -182,10 +182,9 @@ let stage2 stage1 = object
     else prog
 end
 
-module StringMap = Map.Make(String)
-let registry = ref StringMap.empty
-let register name abi = registry := StringMap.set !registry ~key:name ~data:abi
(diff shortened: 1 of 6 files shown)


lib/bap_c/bap_c_abi.ml
```diff
@@ -182,6 +182,11 @@ let stage2 stage1 = object
     else prog
 end
 
+module StringMap = Map.Make(String)
```
ivg, line 185: The `Map` module is already available in the `String` module, vid., `String.Map`. But I think that using `Hashtbl` here is much more natural, e.g.,

```
let registry = Hashtbl.create (module String)
...
```

Could you please switch to Hashtbl here?

ivg:
thanks, waiting for Travis for the green light!

Issue #1089, opened 2020-04-03
Enable support for windows pdb files


bin2415:
Hi, I want to train bap-byteweight with some windows binaries. And I found that there is an option `bap-byteweight train` option to train `bap-byteweight` with new binaries. But I'm not sure how to pass ground truth files to `bap-byteweight`.

Thanks in advance!

bin2415:
Yeah, But I have separate `pdb` files that contain debugging information.

bin2415:
No. It does not output any symbol.

ivg:
Gotcha, can you, please, upload an archive with the binary and pdb here, so that we can debug it here? I don't have msvs nearby :)

bin2415:
The binary and pdb is attached.
[Archive.zip](https://github.com/BinaryAnalysisPlatform/bap/files/4429674/Archive.zip).

Is it because I compiled the binary in windows and ran bap in linux so the bap can't find the proper path of `pdb` file?

ivg:
I believe it's because we've never tried the new bap-byteweight tool with pdb files before, so no surprises it didn't work out of the box. We will try to enable support for them, before the 2.1 release.

bin2415:
Great! Looking forward to the new feature, thanks!

PR #1099, opened 2020-05-13
adds support for windows PDB files

gitoleg:
This PR adds the support of windows PDB files and makes it possible to extract symbols addresses and names. Also this PR fixes https://github.com/BinaryAnalysisPlatform/bap/issues/1089 
## about PDB files
PDB (stands for Program Data Base) is a multi-stream file format, where each
stream (aka file) can represent arbitrary data. The stream we need is DBI - Debug Information Stream, which is the largest stream in a file. It references modules - aka object files - and for each
of them, we can have access to a symbols stream. Finally, we pick only such records from symbols stream, that corresponds to functions starts. On the low level, the symbols we need can have one the next kinds:
`S_GPROC32, S_LPROC32, S_GPROC32_ID, S_LPROC32_ID, S_LPROC32_DPC and
S_LPROC32_DPC_ID`. These records contain information about symbols names, addresses, and sizes, i.e. all what we need.
## Implementation details
The implementation is totally inspired by `llvm-pdbtool` that dumps PDB files content. Each dumper is derived from base visitor class, provided by `llvm` library.
Basically, we need to do the same: to derive from that class and override
only one method (so far), that gives us access to functions starts records.
## Usage
From a user perspective, the PDB file should:
 - have the same name as a corresponded executable (but have .pdb extension)
 - be in the current working directory
This could be changed via `--llvm-pdb-path` option.
## Useful links:
[1] https://llvm.org/docs/PDB/index.html
[2] https://github.com/microsoft/microsoft-pdb
[3] https://pierrelib.pagesperso-orange.fr/exec_formats/MS_Symbol_Type_v1.0.pdf


lib/bap_llvm/bap_llvm_ogre_loader.ml
```diff
@@ -82,28 +79,38 @@ module Make(M : Monad.S) = struct
...
       "Llvm_loader_fail" (Llvm_loader_fail 0)
 
+  let pdb_file filename =
+    let file =
+      if Sys.is_directory P.pdb_path then
+        let pdb_file = sprintf "%s.pdb"
+            Filename.(remove_extension @@ basename filename) in
+        Filename.concat P.pdb_path pdb_file
+      else P.pdb_path in
+    if Sys.file_exists file then file
```
ivg, line 103: `Sys.is_directory p` raises an exception if `p` doesn't exist. Therefore this code is unreachable. You should first check if it exists, then if it is, check if it is a directory or not.

lib/bap_llvm/llvm_loader_stubs.c
```diff
@@ -18,14 +18,17 @@ static void loader_fail(const struct bap_llvm_loader *loader, int n) {
...
-    CAMLparam1(arg);
+value bap_llvm_load_stub(value arg, value pdb_path) {
+    CAMLparam2(arg, pdb_path);
     CAMLlocal1(result);
     const struct caml_ba_array* array = Caml_ba_array_val(arg);
     if ((!array->dim[0]) || (array->num_dims != 1))
         failn(1);
+    const char * pdb = NULL;
+    if (caml_string_length(pdb_path))
+        pdb = String_val(pdb_path);
```
ivg, line 29: the `pdb` pointer will be invalidated on the next call to OCaml runtime. So if `bap_llvm_loader_create` uses any OCaml callbacks (or will use in future) it will segfault. It is safer to strdup it (and then free at the end of the function). When freeing make sure that the `pdb` pointer is not stored somewhere in LLVM (on the first glance it shouldn't, but it is better to be sure, we are here to detect UAF not to create them).

Issue #848, opened 2018-06-26
Bil Store Eval behaves badly


DIJamner:
In the following code, `Exp.eval read` should return `0x1:16u`. However, it returns `0x101:16u`. Note that this behavior is only observable on pull request #847 since without that fix, it will give a type error at runtime.
```
open Bap.Std;;
open Bil;;
let a1 = int (Word.zero 64);;
let a2 = int (Word.one 64);;
let m = (unknown "mem" (Type.mem `r64 `r16));;
let m1 = store ~mem:m ~addr:a1 (int (Word.one 16)) LittleEndian `r16;;
let m2 = store ~mem:m1 ~addr:a2 (int (Word.one 16)) LittleEndian `r16;;
let read = load ~mem:m2 ~addr:a1 LittleEndian `r32;;
Type.infer read;;
Exp.eval read;;
```
This is likely due to incorrect behavior of `store` on memories with a cell size other than 8 bits.

ivg:
I believe you meant `r16 in the last read, as `r32 would result in the undefined value, since we have only three defined bytes

ivg:
While I believe that this issue should be fixed in BAP 2.0 in a sense that the semantics of loads and stores are now correct in the Core Theory and Core Theory terms should be reified into BIL loads and stores correctly (and vice versa), but it looks like that `Exp.eval` is still broken, but differently. Now, `Exp.eval read` doesn't terminate.

PR #1102, opened 2020-05-20
fixes Bil.eval memory operations for different sizes of values

gitoleg:
`Bil.Eval` still relies on the assumption that memory can address only values of 1-byte length. And once the stored/loaded value size differs from memory value size, we try to simplify an expression. But the simplification in its turn can leave an expression unchanged since it takes into account both memory value size and size of a stored/loaded value. And that's how we get an infinite loop: since an expression is unchanged, we try to eval it again, simplify, and so on.

This PR fixes this problem and checks both the size of the stored/loaded value and memory value size.

fixes https://github.com/BinaryAnalysisPlatform/bap/issues/848

### DIFF
--- a/lib/bap_types/bap_eval.ml
+++ b/lib/bap_types/bap_eval.ml
@@ -26,6 +26,11 @@ let is_shift = function
   | Binop.LSHIFT | Binop.RSHIFT | Binop.ARSHIFT -> true
   | _ -> false
 
+let infer_value_size m =
+  match Bil.Type.infer m with
+  | Ok (Type.Mem (_,x)) -> x
+  | _ -> invalid_arg "type error"
+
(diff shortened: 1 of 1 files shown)


lib/bap_types/bap_eval.ml
```diff
@@ -88,7 +93,7 @@ module Make2(State : Monad.S2) = struct
       self#eval exp >>= self#storage_of_value
 
     method eval_load ~mem ~addr endian sz =
-      if Size.(sz <> `r8)
+      if Size.(sz <> infer_value_size mem)
```
ivg, line 96: `infer_value_size` shall not raise, instead `self#type_error TE.bad_mem` should be called if `mem` is not of type `Mem`.

Issue #1111, opened 2020-06-04
opam depext requires llvm-3.8 and llvm-9


ivg:
```
$ opam depext bap
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=ubuntu, os-family=debian
# The following system packages are needed:
clang
debianutils
dejagnu
libcurl4-gnutls-dev
libgmp-dev
libzip-dev
llvm
llvm-3.8-dev # wow?
llvm-9-dev
m4
perl
pkg-config
python2.7
time
zlib1g-dev
```

ivg:
Already fixed

PR #1114, opened 2020-06-09
Try Ubuntu Bionic on Travis CI

XVilka:
Trusty is the EOL anyway: https://docs.travis-ci.com/user/reference/trusty/

Also, I think it's time to drop 1.x opam support, along with unnecessary local copy of Travis CI scripts - you can just use https://github.com/ocaml/ocaml-ci-scripts directly


ivg:
I like that sort of patches, +9, -221) Thanks a lot! This will extend a bit Travis life, well until we move fully to GH actions (we are again very close to the timeout because of the addition of z3).

ivg:
also, I hope it will fix #1111 which also breaks #1115

Issue #1109, opened 2020-06-02
handling of XDG_CACHE_DIR is broken in the cache plugin


ivg:
1) we're trying to hijack the whole directory instead of creating a folder relative to it, which is at least not clean.
2) if a directory is a file we fail with an error:
```
$ touch cache
$ XDG_CACHE_HOME=cache bap .
Internal error: Unix.Unix_error(Unix.ENOTDIR, "rename", ".//1770a8bf-71bc-433a-a608-ff33bb3293a3/")
Backtrace:
Raised at file "src/exn.ml", line 69, characters 6-15
Called from file "bap_cache.ml", line 195, characters 2-19
Called from file "bap_cache_main.ml", line 91, characters 2-15
Called from file "bap_cache_main.ml", line 138, characters 2-27
Called from file "bap_main.ml", line 789, characters 37-46
Called from file "bap_main.ml", line 945, characters 25-28
```
3) if a directory is not empty (which the common case) we also fail, because the initialization procedure is unable to initialize the expected directory structure:
```
$ mkdir cache
$ touch cache/thing
$ XDG_CACHE_HOME=cache bap .
Internal error: (Sys_error "cache/data: No such file or directory")
Backtrace:
Raised by primitive operation at file "bap_cache_gc.ml", line 101, characters 2-18
Called from file "bap_cache_gc.ml", line 143, characters 16-43
Called from file "bap_cache_main.ml", line 138, characters 2-27
Called from file "bap_main.ml", line 789, characters 37-46
Called from file "bap_main.ml", line 945, characters 25-28
```

4) It is also assumed that `dirname p` returns a parent directory, which is not true in general, e.g.,
```
XDG_CACHE_HOME=. bap .
Internal error: Unix.Unix_error(Unix.EBUSY, "rename", ".//6aa795be-60aa-49de-9fe0-87332ea18558/")
Backtrace:
Raised at file "src/exn.ml", line 69, characters 6-15
Called from file "bap_cache.ml", line 195, characters 2-19
Called from file "bap_cache_main.ml", line 91, characters 2-15
Called from file "bap_cache_main.ml", line 138, characters 2-27
Called from file "bap_main.ml", line 789, characters 37-46
Called from file "bap_main.ml", line 945, characters 25-28
```

5) The diagnostic messages could be better, for example, when the cache directory is not writable, we should provide a descriptive error message, and maybe even indicate this as a working and disable cache.

ivg:
[...]

PR #1115, opened 2020-06-09
Try to use GitHub Actions

XVilka:
Travis is too clunky, also it is possible to use new action template from https://discuss.ocaml.org/t/github-actions-for-ocaml-opam-now-available/4745

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,17 +31,17 @@ jobs:
         with:
           ocaml-version: ${{ matrix.ocaml-version }}
 
-      - name: Pinning BAP
+      - name: Pin BAP
         run: opam pin add bap . --no-action
 
-      - name: Configuring Homebrew LLVM
+      - name: Configure Homebrew LLVM
         if: matrix.os == 'macos-latest'
         run: echo '::set-env name=LLVM_CONFIG::/usr/local/opt/llvm/bin/llvm-config'
 
-      - name: Installing system dependencies
+      - name: Install system dependencies
         run: opam depext bap --yes
 
-      - name: Installing opam dependencies
+      - name: Install opam dependencies
         run: opam install . --deps-only
 
       - name: Install BAP
@@ -49,12 +49,12 @@ jobs:
         env:
           XDG_CACHE_HOME: .cache
 
-      - name: running the testsuite
+      - name: Run the testsuite
         run: |
           eval $(opam env)
           make check
 
-      - name: running bap-veri
+      - name: Run bap-veri
         run: |
           eval $(opam env)
           cd testsuite
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,8 +36,7 @@ jobs:
 
       - name: Configuring Homebrew LLVM
         if: matrix.os == 'macos-latest'
-        env:
-            LLVM_CONFIG: '/usr/local/opt/llvm/bin/llvm-config'
+        run: echo '::set-env name=LLVM_CONFIG::/usr/local/opt/llvm/bin/llvm-config'
 
       - name: Installing system dependencies
         run: opam depext bap --yes --with-doc --with-test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,8 +14,8 @@ jobs:
           #- macos-latest
           #- windows-latest
         ocaml-version:
-          - 4.09.1
-          - 4.08.1
+          # - 4.09.1
+          # - 4.08.1
           - 4.07.1
 
     runs-on: ${{ matrix.os }}
@@ -39,20 +39,24 @@ jobs:
         run: echo '::set-env name=LLVM_CONFIG::/usr/local/opt/llvm/bin/llvm-config'
 
       - name: Installing system dependencies
-        run: opam depext bap --yes --with-doc --with-test
+        run: opam depext bap --yes
 
       - name: Installing opam dependencies
-        run: opam install . --deps-only --with-doc --with-test
+        run: opam install . --deps-only
 
       - name: Install BAP
         run: opam install bap -v --yes --with-test
+        env:
+          XDG_CACHE_HOME: .cache
 
       - name: running the testsuite
+        run: make check
+
+      - name: running bap-veri
         run: |
           eval $(opam env)
           cd testsuite
           git submodule init
           git submodule update
           opam pin add bap-veri veri/bap-veri/ -y
-          make check
           make veri
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,6 @@
 name: Main workflow
 
 on:
-  - push
   - pull_request
 
 jobs:
@@ -12,7 +11,6 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
-          #- windows-latest
         ocaml-version:
           - 4.09.1
           - 4.08.1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -48,6 +48,8 @@ jobs:
         run: opam install bap -v --yes --with-test
         env:
           XDG_CACHE_HOME: .cache
+          TMPDIR: /tmp
+          BAP_LOG_DIR: /tmp/bap-log
 
       - name: Run the testsuite
         run: |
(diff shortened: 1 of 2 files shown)


ivg:
It looks very promising, the main roadblocker is #1111, which I think is resolved by #1114 but it looks like that we need to rebase/merge master into this PR. 

The other missing thing is running, tests and checks, with `make test` and `make check` (as adding --with-tests and --with-docs won't really run them, and they (tests) are built despite the presence of this flag.

XVilka:
@dbrumley there is already existing GitHub Action for publishing to Docker, we use it in radare2 project: https://github.com/marketplace/actions/publish-docker

XVilka:
I wanted to setup also Windows worker, but strangely z3 is disabled for win32 for some reason by @fdopen in https://github.com/fdopen/opam-repository-mingw/commit/7ca1bb19c7943db886499aad20b68712fa4bf7d4

XVilka:
`homebrew/versions` tap is deprecated and now is empty, so I just used plain `llvm@9` from Homebrew for MacOS build, but then `conf-bap-llvm` fails with the error `llvm-config` not found:
```
# [...]
# command which  failed: exit status 1
# command which llvm-config-3.8 failed: exit status 1
# command which llvm-config-38 failed: exit status 1
# command which llvm-config-3 failed: exit status 1
# command which llvm-config failed: exit status 1
# command ls /usr/local/Cellar | grep 3.4 failed: exit status 1
# command which llvm-config-mp-3.4 failed: exit status 1
# command which llvm-config-3.4 failed: exit status 1
# command which llvm-config-34 failed: exit status 1
# command which llvm-config-3 failed: exit status 1
# command which llvm-config failed: exit status 1
# LLVM not found



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
┌─ The following actions failed
│ λ build conf-bap-llvm 1.6
```

XVilka:
Strange error on MacOS that I have no idea why:
```
[ERROR] The compilation of bap-veri failed at "/Users/runner/.opam/opam-init/hooks/sandbox.sh build make".

#=== ERROR while compiling bap-veri.0.2.5 =====================================#
# context              2.0.7 | macos/x86_64 | ocaml-base-compiler.4.09.1 | pinned(file:///Users/runner/runners/2.263.0/work/bap/bap/testsuite/veri/bap-veri)
# path                 ~/.opam/4.09.1/.opam-switch/build/bap-veri.0.2.5
# command              ~/.opam/opam-init/hooks/sandbox.sh build make
# exit-code            2
# env-file             ~/.opam/log/bap-veri-91700-ae7898.env
# output-file          ~/.opam/log/bap-veri-91700-ae7898.out
### output ###
# make: *** No targets specified and no makefile found.  Stop.
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: opam install . --deps-only --with-doc --with-test
+
+      - name: Install BAP
+        run: opam install bap -v --yes --with-test
+
+      - name: running the testsuite
+        run: |
+          eval $(opam env)
+          cd testsuite
+          git submodule init
```
XVilka, line 54: I think those unnecessary since we already cloned with submodules recursively.

gitoleg: well, that's one of the problems - I don't see that it's done recursively :(

.github/workflows/main.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Pinning BAP
+        run: opam pin add bap . --no-action
+
+      - name: Configuring Homebrew LLVM
+        if: matrix.os == 'macos-latest'
+        run: echo '::set-env name=LLVM_CONFIG::/usr/local/opt/llvm/bin/llvm-config'
+
+      - name: Installing system dependencies
+        run: opam depext bap --yes --with-doc --with-test
```
ivg, line 42: we don't need `--with-doc` and `--with-test`

.github/workflows/main.yml
ivg: even if `submodule=true` won't work (which I doubt) it is still not needed, as our `Makefile` is already doing this. Thus doing, `make check` is enough even if submodules are not initialized. 

So the question, why just not?
```
name: Running unit tests
run: make test

name: Running functional tests
run: make check
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Installing opam dependencies
+        run: opam install . --deps-only --with-doc --with-test
```
ivg, line 45: in order to run tests during the installation, keep `--with-tests`

and add to opam/opam's install field the following
```
[make "test"] {with-test}
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install BAP
+        run: opam install bap -v --yes --with-test
+
+      - name: running the testsuite
+        run: |
+          eval $(opam env)
+          cd testsuite
+          git submodule init
+          git submodule update
+          opam pin add bap-veri veri/bap-veri/ -y
```
ivg, line 56: please split veri into a separate step

opam/opam
```diff
@@ -166,6 +166,7 @@ install: [
   ["cp" "run_piqi_tests.native" "%{bin}%/bap_piqi_tests"]
   ["cp" "run_traces_tests.native" "%{bin}%/bap_traces_tests"]
   ["cp" "run_stub_resolver_tests.native" "%{bin}%/bap_stub_resolver_tests"]
+  [make "test"] {with-test}
```
ivg, line 169: great, we already have the tests up and running!

XVilka:
After trying MacOS build again it seems there is a new error, and it might be in the code:
```
┌─ The following actions failed
│ ∗ install bap master
└─ 
╶─ No changes have been performed
# context              2.0.7 | macos/x86_64 | ocaml-base-compiler.4.09.1 | pinned(git+file:///Users/runner/runners/2.263.0/work/bap/bap#HEAD#856b159b)
# path                 ~/.opam/4.09.1/.opam-switch/build/bap.master
# command              ~/.opam/opam-init/hooks/sandbox.sh install make test
# exit-code            2
# env-file             ~/.opam/log/bap-91713-4f4523.env
# output-file          ~/.opam/log/bap-91713-4f4523.out
### output ###
# [...]
# bil.info> providing a lifter for all BIL lifters
# powerpc.info> Providing PowerPC semantics in BIL
# cache.warning> Failed to read config file: ("Unix.Unix_error(Unix.ENOENT, \"open\", \".cache/config.3\")")
# Failed to initialize BAP: Internal error: (Sys_error ".cache/data: No such file or directory")
# Backtrace:
# Raised by primitive operation at file "plugins/cache/bap_cache_gc.ml", line 101, characters 2-18
# Called from file "plugins/cache/bap_cache_gc.ml", line 143, characters 16-43
# Called from file "plugins/cache/bap_cache_main.ml", line 138, characters 2-27
# Called from file "lib/bap_main/bap_main.ml", line 789, characters 37-46
# Called from file "lib/bap_main/bap_main.ml", line 945, characters 25-28
# 
# make: *** [test] Error 1

# Run eval $(opam env) to update the current shell environment
'opam install bap -v --yes --with-test' failed.
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Install system dependencies
+        run: opam depext bap --yes
+
+      - name: Install opam dependencies
+        run: opam install . --deps-only
+
+      - name: Install BAP
+        run: opam install bap -v --yes --with-test
+        env:
+          XDG_CACHE_HOME: .bap-cache
```
ivg, line 50: looks like that we're blocked by #1109 and the fact that this variable is actually used on MacOS by other applications. 
As a temporary workaround, let's remove these two lines and change `make test` in opam/opam to `XDG_CACHE_HOME=.bap-cache make test` to limit the effect of this variable.

ivg:
Ok, I've pushed the fix to #1109, reset Oleg's morning attempts and rebased the original patch onto the new master.

ivg:
Finally, we passed the test and sandboxing issues, but still failing on macos since we're picking not the installed LLVM, but the LLVM which is shipped with the os (and which is stripped and doesn't have arm,powerpc, and other targets).

ivg:
Finally, thanks to everyone for the hard work! This is the first step, we will move more of our infrastructure to the actions and optimize the testsuite that takes  more than hour on Mac (for some unknown to us reason). But anyway, we now have a working CI system, and soon we will get CICD :)

PR #1116, opened 2020-06-09
updates to OCaml 4.{10,11}, drops 4.07, and switches to core_kernel v0.14

drvink:
There's a few changes in here due to API incompatibility with the Jane Street libraries (but nothing that changes bap's own APIs), so you'll either want to make them conditional on version (not sure there's an easy way to do this in e.g. the `oasis` file that needed changing) or just wait to merge this for the next release of bap.

---

Thanks to ppx_js_style (sigh), this also depends on:

* ocaml/ocamlfind#9
* alavrik/piqi-ocaml#15

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -19,7 +19,11 @@ jobs:
          - name: Install OCaml
            uses: avsm/setup-ocaml@v1
            with:
-             ocaml-version: 4.07.1
+             ocaml-version: 4.08.1
+
+         - name: Add the testing Repository
+           run: opam repo add bap git://github.com/BinaryAnalysisPlatform/opam-repository#testing
+
 
          - name: Pin OASIS
            run: opam pin add oasis https://github.com/BinaryAnalysisPlatform/oasis.git
--- a/.github/workflows/build-dev-repo.yml
+++ b/.github/workflows/build-dev-repo.yml
@@ -11,9 +11,10 @@ jobs:
             - ubuntu-latest
             - macos-latest
           ocaml-version:
+            - 4.11.1
+            - 4.10.1
             - 4.09.1
             - 4.08.1
-            - 4.07.1
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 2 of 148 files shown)


ivg:
This ppx_js_style is acting totally wrong, under no circumstance the style violation shall break code that doesn't belong to the company that implements this style and especially on the code on which they depend. I have submitted an issue to JS and hope that they will rectify this
https://github.com/janestreet/ppx_jane/issues/5

XVilka:
There are some conflicts that need to be resolved, also `piqi` has merged the PR but didn't made a new release yet. `ocamlfind` PR was ignored it seems.

ivg:
A small status update. To bypass the ppx_js_style issues I have created our own ppx_bap rewriter with blackjack and no styles. I will soon push it to the upstream opam repository but so far it is in our own opam-repository. Once we pass the tests, I will update our opam repository with the new bounds and we're ready to go. 

I also plan to investigate if it is possible to build for two versions of core_kernel, namely v0.13 and v0.12 or we will completely skip the v0.12 version, @drvink maybe you have already tried this? Are those two compatible?

drvink:
@ivg There were definitely some changes between 0.12.x and 0.13.x, because this PR was originally for compatibility with 0.13.x. I should still have the previous patch in my local git if you want it.

ivg:
> @ivg There were definitely some changes between 0.12.x and 0.13.x, because this PR was originally for compatibility with 0.13.x. I should still have the previous patch in my local git if you want it.

I am talking more about the changes between v0.13 and v0.14 :) I would be cool if we will handle both versions (v0.12 is definitely out of scope). 

The cool thing about this PR that makes me happy is that now we can officially went to OCaml 4.08 and that means that we can safely use binding operators. I plan to add support for them to monads/knowledge/core-theory. Yay!

drvink:
@ivg I pushed a branch `janestreet-actually-0.13` from what was lying around in my git reflog; I think it more or less reflects what changed between 0.13.x and 0.14.x.

ivg:
Yep, there are a few road-blockers, namely the changed interface of Univ_map.Key (which became a binary functor in v0.14 as well as the Caml_unix, and core_kernel.caml_unix that are absent in v0.13). Given this, and that v0.14 officially opens 4.08 for us, I think we can forfeit the 13th version (anyway, it is a good tradition to miss the 13th version). We are bringing two new versions of OCaml instead of 4.07, namely, 4.11 and 4.10.

PR #1128, opened 2020-06-15
fix rev16 instruction

### LINKED ISSUE #1127: Incorrect instruction interpretation within ARM lifter
In the legacy ARM-BIL lifter, there's a interpretation of instruction `REV16` as follows:
```ocaml
  | `REV16, [|`Reg dest; src; cond; _|] ->
    let s = exp_of_op src in
    let i16 = int32 16 in
    let rev = Bil.(s lsl i16 lor s lsr i16) in
    exec [assn (Env.of_reg dest) rev] cond
```
which is a misinterpretation of such instruction, in *ARM Architecture Reference Manual 4.1.57* we have
> REV16 (Byte-Reverse Packed Halfword) reverses the byte order in each 16-bit halfword of a 32-bit register.

which, as the reference said, should be
> Operation
if ConditionPassed(cond) then
 Rd[15: 8] = Rm[ 7: 0]
 Rd[ 7: 0] = Rm[15: 8]
 Rd[31:24] = Rm[23:16]
 Rd[23:16] = Rm[31:24]

obviously the current interpretation in the lifter acts like the following
> Operation
if ConditionPassed(cond) then
 Rd[31: 16] = Rm[15: 0]
 Rd[ 15: 0] = Rm[31: 16]

resulting an incorrect behavior under the `REV16` instruction

Phosphorus15:
Hopefully this PR will fix #1127, not sure if this would affects related tests.


lib/arm/arm_lifter.ml
```diff
@@ -334,8 +334,12 @@ let lift_bits mem ops (insn : bits_insn ) =
     exec [assn (Env.of_reg dest) rev] cond
   | `REV16, [|`Reg dest; src; cond; _|] ->
     let s = exp_of_op src in
-    let i16 = int32 16 in
-    let rev = Bil.(s lsl i16 lor s lsr i16) in
+    let i8 = int32 8 in
+    let rev = Bil.((s land 0xff lsl i8) lor
```
ivg, line 338: It will not compile. Also, keep in mind that `s` here is not changing, so you're shifting the same value left or right by 8 bits and never reach higher bits.

Phosphorus15:
> please use `extract` and express the semantics exactly as in the Reference Manual.
> 
> And thanks for fixing it.

Thanks for the fast and detailed response, would you think it a good idea that I change the `REV` instruction to use `extract` too? since its current implementation is using bitmask & logand too:
```ocaml
| `REV, [|`Reg dest; src; cond; _|] ->
    let s = exp_of_op src in
    let i24 = int32 24 in
    let i8 = int32 8 in
    let umask = int32 0xff0000 in
    let lmask = int32 0xff00 in
    let rev =
      let open Bil in
      s              lsl i24 lor
      s              lsr i24 lor
      (s land umask) lsr i8  lor
      (s land lmask) lsl i8
    in
```

ivg:
> Thanks for the fast and detailed response, would you think it a good idea that I change the `REV` instruction to use `extract` too? since its current implementation is using bitmask & logand too:

Certainly, it is a good idea!

Phosphorus15:
Please check if the fix is valid can be merged.

ivg:
@Phosphorus15, we use the `ready for review` label to make it easier to identify PRs which are, well, ready for review :)

lib/arm/arm_lifter.ml
```diff
@@ -322,20 +322,22 @@ let lift_bits mem ops (insn : bits_insn ) =
...
   | `REV16, [|`Reg dest; src; cond; _|] ->
     let s = exp_of_op src in
-    let i16 = int32 16 in
-    let rev = Bil.(s lsl i16 lor s lsr i16) in
+    let i8 = int32 8 in
+    let rev = Bil.(((extract 7 0 s) lsl i8) lor 
+                   ((extract 15 8 s) lsr i8) lor
+                   ((extract 23 16 s) lsl i8) lor
+                   ((extract 31 24 s) lsr i8)
+                  ) in
```
ivg, line 340: Semanitcs notes:

The code is not well-typed and `rev` is always `0`.  The expression `extract 7 0 s` has type `i8` (i.e., an 8-bit word). Thus you're shifting 8-bit words left or right by eight bits (getting zeros always).  When you `lor` N 8-bit words, you will still get an 8-bit word. Finally, you assing the `dest` variable (which is a 32-bit register) an 8-bit word. And booms.

Instead of using shifts and ors, you can use `extract` and `concat` which lets you express the semantics _exactly_ as in the reference manual. 

Style notes: 

1) Do not leave `in` `) in`, `)`, or any other purely grammatical/service words on a single line. Each line of code, shall be a line of code. The exception is `begin/end` and other structural expressions, such as while, for, etc

2) Do not put superfluous parentheses. They clutter the code and confuse as it is not clear what were the author's intentions.  

In this case, parenthesis around extract are acceptable, but keep in mind that function applications bind tighter than infix operators. Since in this case the infix operator is fully made of ASCII characters, it is really easy to confuse it with function application especially when there is no syntax highlighting, e.g., `extract 7 0 s lsl i8` is hard to parse without highlighting, thoug it has the same `(extract 7 0 s) lsl i8` meaning.

lib/arm/arm_lifter.ml
```diff
@@ -322,20 +322,22 @@ let lift_bits mem ops (insn : bits_insn ) =
...
     let i8 = int32 8 in
-    let umask = int32 0xff0000 in
-    let lmask = int32 0xff00 in
     let rev =
       let open Bil in
       s              lsl i24 lor
       s              lsr i24 lor
-      (s land umask) lsr i8  lor
-      (s land lmask) lsl i8
+      (extract 23 16 s) lsr i8  lor
```
ivg, line 329: see the comment about REV16, it is also applicable here. And please, also just use a concatenation of 4 bytes, obtained with `extract`.

lib/arm/arm_lifter.ml
```diff
@@ -320,22 +320,19 @@ let lift_bits mem ops (insn : bits_insn ) =
...
-    let umask = int32 0xff0000 in
-    let lmask = int32 0xff00 in
-    let rev =
-      let open Bil in
-      s              lsl i24 lor
-      s              lsr i24 lor
-      (s land umask) lsr i8  lor
-      (s land lmask) lsl i8
-    in
+    let rev = Bil.(concat
```
ivg, line 323: First of all, the code is not indented correctly. Second, and most importantly, you can use the [(^)] operator to make your code much more readable, e.g.,
```ocaml
    let rev = Bil.(extract 7 0 s ^
                   extract 15 8 s ^
                   extract 23 16 s ^
                   extract 31 24 s) in
```

[1]: https://binaryanalysisplatform.github.io/bap/api/master/bap/Bap/Std/Bil/index.html#val-(^)

lib/arm/arm_lifter.ml
```diff
@@ -320,22 +320,19 @@ let lift_bits mem ops (insn : bits_insn ) =
...
+                    (concat (extract 7 0 s)
+                            (extract 15 8 s))
+                    (concat (extract 23 16 s)
+                            (extract 31 24 s))) in
     exec [assn (Env.of_reg dest) rev] cond
   | `REV16, [|`Reg dest; src; cond; _|] ->
     let s = exp_of_op src in
-    let i16 = int32 16 in
-    let rev = Bil.(s lsl i16 lor s lsr i16) in
+    let rev = Bil.(concat
```
ivg, line 331: the same is true for this snippet, use `^` for concatenation.

ivg:
thanks for fixing it and making the code nicer, good job!

PR #1129, opened 2020-06-15
handles correctly XDG_CACHE_HOME

ivg:
we shall not take the whole chaching directory, but create a subfolder there.


ivg:
it partially addresses #1109

PR #1141, opened 2020-06-19
Arm AArch64 instruction set plugin.

zhaoyonghe:
This pr presents a draft of Core Theory/KB based ARM AArch64 instructions' lifter, which is mostly a incomplete skeleton presenting how the final lifter will be. Now it can only lift one instruction `LDR Xn addr`. No Float number and SIMD related supports for now.

Because the instruction support two set of registers: `Xn` and `Wn`, but they actually share the same physical register, so `Xn` and `Wn` will be mapped onto same register `xn`(see a64_env_registers.ml:Line86). 

**Question:**
Which manual I should reference?
[https://www.element14.com/community/servlet/JiveServlet/previewBody/41836-102-1-229511/ARM.Reference_Manual.pdf](url) This one has the detailed instruction set (a lot of instructions, which parts are needed?), but seems too old?
[https://static.docs.arm.com/den0024/a/DEN0024A_v8_architecture_PG.pdf](url) This one seems very offical, but has much more less about the instructions.


XVilka:
@zhaoyonghe here is the latest official ARMv8 instruction set reference:
- https://developer.arm.com/docs/ddi0487/latest/arm-architecture-reference-manual-armv8-for-armv8-a-architecture-profile
- https://static.docs.arm.com/ddi0487/fb/DDI0487F_b_armv8_arm.pdf

XVilka:
It will require some tests as well, see e.g.:
- https://github.com/BinaryAnalysisPlatform/bap-testsuite/tree/master/mc.tests
- https://github.com/BinaryAnalysisPlatform/bap-testsuite/blob/master/bap.tests/run.exp
- https://github.com/BinaryAnalysisPlatform/bap-testsuite/blob/master/bap.tests/api-main.exp

ivg:
and unit tests as well,

* https://github.com/BinaryAnalysisPlatform/bap/tree/master/lib_test/powerpc

In general, for the development setup you can follow the powerpc lifter. So far it looks ok, as far as I understand you're not building from the BAP source repository and installing bap from opam, is that right?

plugins/arm_a64/a64.ml
```diff
@@ -0,0 +1,23 @@
+open Core_kernel
```
ivg, line 1: Please, indent the code using ocp-indent. Which development environment are you using? It shall be set up so that it automatically indents your code. Tell us what you're using and we will help you in setting things up.

zhaoyonghe: I use vscode.

zhaoyonghe:
> and unit tests as well,
> 
> * https://github.com/BinaryAnalysisPlatform/bap/tree/master/lib_test/powerpc
> 
> In general, for the development setup you can follow the powerpc lifter. So far it looks ok, as far as I understand you're not building from the BAP source repository and installing bap from opam, is that right?

Yes, I just install bap from opam.

plugins/arm_a64/a64.ml
ivg: you can now just do `make indent` and it will indent your code (provided that you have ocp-indent installed)

zhaoyonghe: Sorry for that. I am about to do it this evening. So I just need to, with ocp-indent installed, run `make indent` without modifying the Makefile?

ivg: Yes, just `git pull` your branch to get the latest master (I've already updated the PR)

zhaoyonghe: ok

plugins/arm_a64/a64_defs.ml
```diff
@@ -0,0 +1,213 @@
...
+  | `B
+  | `BL
+  (* unconditional branch (register) *)
+  | `BLR
+  | `BR
+  | `RET
+]
+
+type mem_access_insn_t = [
+  (* load-store single register TODO *)
```
zhaoyonghe, line 101: **Question:**
There are a lot of variants for one single instruction, eg. the following three segments are all variant of `LDR`, and I need to distinguish them. 

Now, I define their type at here(very ugly), and lift the specific variant using more general functions here [https://github.com/BinaryAnalysisPlatform/bap/pull/1141/files#diff-6b5b4bf2eb12277cf6bf719ac07f9bd3R20](url).  

Am I on the right way? If yes, any suggestions?

XVilka: It will apply only to LDR and STR instructions, while the rest are less complex. I think it shouldn't be a problem. I don't know what @ivg think of that.

plugins/arm_a64/a64_defs.ml
```diff
@@ -0,0 +1,213 @@
+open Core_kernel
```
ivg, line 1: We don't really need this module at all (or at least its current contents). There is no need to maintain two sets of definitions for registers as variables and as polymorphic constructors. This is a design error inherited from the ARMv7 lifter.

plugins/arm_a64/a64_defs.ml
ivg: I provided a big review and suggest starting from scratch, I think we have a false start. See the comment below.

zhaoyonghe: Hi @ivg , just read your comments. Thank you for the suggestions!

 Is [extend_reg](https://github.com/BinaryAnalysisPlatform/bap/blob/1d848484fc20f0d07202804a02e9eedb028d6cba/plugins/arm_a64/a64_utils.ml#L92) the "small and easy to use function" you commented (although it may be a bad implementation)? They are all in part J(Armv8 Pseudocode) of the official manual. Once we have these small functions, we still need to implement pseudo-code in the specification(actually it is what I am trying to do, I wrote these code following those pseudo-code). Thus, if I understand what you mean correctly, your suggestion is actually let me stop merging all load variants into one function. Instead, write the ocaml/core theory version of the pseudo-code for every instruction variant(one ocaml function per instruction variant). 

Please correct me if I am wrong.

plugins/arm_a64/a64_defs.ml
ivg: btw, by this module I didn't mean `Core_kernel` as it may look like, it was a comment to the whole file, i.e., `a64_defs.ml`

plugins/arm_a64/a64_utils.ml
```diff
@@ -0,0 +1,113 @@
...
+    | `Reg _ -> raise Unexpected_Situation
+
+  let little_load_bv ~bv_sort ~address = 
+    loadw bv_sort b0 (var Env.memory) address
+
+  let extend_raw ~assem_reg_sort ~unsign ~x = 
+    let extend = if unsign then unsigned else signed in
+    extend assem_reg_sort x 
+
+  let extend_reg ~r ~extend_type ~is_64_reg ~shift_len =
```
ivg, line 92: For the reference, here's the [reference implementation][1]: 

```
/ ExtendReg()
// ===========
// Perform a register extension and shift

bits(N) ExtendReg(integer reg, ExtendType exttype, integer shift)
    assert shift >= 0 && shift <= 4;
    bits(N) val = X[reg];
    boolean unsigned;
    integer len;

    case exttype of
        when ExtendType_SXTB unsigned = FALSE; len = 8;
        when ExtendType_SXTH unsigned = FALSE; len = 16;
        when ExtendType_SXTW unsigned = FALSE; len = 32;
        when ExtendType_SXTX unsigned = FALSE; len = 64;
        when ExtendType_UXTB unsigned = TRUE;  len = 8;
        when ExtendType_UXTH unsigned = TRUE;  len = 16;
        when ExtendType_UXTW unsigned = TRUE;  len = 32;
        when ExtendType_UXTX unsigned = TRUE;  len = 64;

    // Note the extended width of the intermediate value and
    // that sign extension occurs from bit <len+shift-1>, not
    // from bit <len-1>. This is equivalent to the instruction
    //   [SU]BFIZ Rtmp, Rreg, #shift, #len
    // It may also be seen as a sign/zero extend followed by a shift:
    //   LSL(Extend(val<len-1:0>, N, unsigned), shift);

    len = Min(len, N - shift);
    return Extend(val<len-1:0> : Zeros(shift), N, unsigned);
```

I like very much that your implementation follows closely the reference so we can match it one-to-one and verify its correctness. 

A few ways to improve it:

1) do not abuse labeled parameter (i.e., `~`) they will just clutter the code in the place of use (and we want it to be as close to the specification pseudocode as possible). Especially in this function, they are not needed, as all arguments are of different types so we can't mess with them at the call site. Also, whenever you find that you want to add labels to your arguments, it is a good indicator that the function is not simple anymore and has too complicated interface. 

2) the `is_64_reg` parameter is superfluous here and may introduce potential anomalies. The reason is that the register, which is passed to the function, already has the sort (and correspondingly size) attached to it. So you should use the information provided by the register. 

3) there is a mismatch between the `extend_type` from the reference and our representation. Why is `LSL` there? We need to rectify this asap.

4) at the epilogue of the function we start to diverge from the reference. You shall use the [extend][3] function here, and I fail to see a reason why it could be used here. 
```

// Extend()
// ========

bits(N) Extend(bits(M) x, integer N, boolean unsigned)
    return if unsigned then ZeroExtend(x, N) else SignExtend(x, N);

// Extend()
// ========

bits(N) Extend(bits(M) x, boolean unsigned)
    return Extend(x, N, unsigned);
```
Don't forget that you can always get the sort for any value or register, so you can go, `_ bitv => _ sort => int` and from `_ var => _ sort => int` and the other way around, `int -> _ sort => _ bitv`, etc. We commonly define the following operator (which we should probably add to the Core Theory, given how useful it is),


Basically, here's I would define this function, 
```ocaml
  let extend_reg reg etype shift =
    let s = Theory.Var.sort reg in
    let n = Theory.Bitv.size s in
    let unsigned,len = match etype with
      | SXTB -> false,8
      | SXTH -> false,16
      | SXTW -> false,32
      | SXTX -> false,64
      | UXTB -> true,8
      | UXTH -> true,16
      | UXTW -> true,32
      | UXTX -> true,64 in
    let len = Int.min len (n - shift) in
    let zeros = zero (Theory.Bitv.define shift) in
    let s' = Theory.Bitv.define (len + shift) in
    let v = append s' (static_extract (len-1) 0 (var reg)) zeros in
    extend v s unsigned
```

I had to define a few helpers, first of all, we have the [extract][4] function which is a little bit more generic than we need for our purposes. The Theory's extract takes runtime values for the hi and lo bits of the range of bytes that are extracted but in our case they are static constants defined by the code itself. So we will define a helper called `static_extract` that will take to ML integers (denoting statically known boundaries) and use the generic extract to extract the range,
```
  let static_extract hi lo x =
    let s = Theory.Bitv.define (hi-lo+1) in
    extract s (byte hi) (byte lo) x
```
The `extract` operation requires us to provide the sort of the output value, which we can statically determine as `hi-lo+1`. As you can see, we can define sorts in runtime, as long as you can compute the size of the bitvector. We also need to wrap ML integers into the target language bitvectors. Since we know that they are going to be small (they denote the register sizes) and since the type of `extract`,
```
    val extract : 'a Bitv.t Value.sort -> 'b bitv -> 'b bitv -> _ bitv -> 'a bitv
```
doesn't require `hi` and `lo` to be of the same sort as the extractee, we will define some helpers (which we will certainly use a lot in other places) to create small bitvectors from integers. We pick 8 bits for small integers and add the following definitions to our theory,
```
  module Byte = Bitvec.M8
  let bytes = Theory.Bitv.define 8

  let byte x = int bytes (Byte.int x)

  let static_extract hi lo x =
    let s = Theory.Bitv.define (hi-lo+1) in
    extract s (byte hi) (byte lo) x
```
which corresponds to 

Here, module `Byte` are ML bitvectors with modulus 8 arithmetic. (Do not confuse `Bitv`, which is the Theory (aka the target language) bitvectors with `Bitvec` wich is ML (aka OCaml, aka the host language) bitvectors aka words, the latter are concrete representation the former is an abstraction).  We also defined `bytes` as the sort for bytes (you may find a better place for it). And the small helper `byte` that takes an ML integer and lifts it to the Theory byte value that represents a byte constant with the given bits.

 One more comment about this final expression of our `extract_reg` function
```
    let s' = Theory.Bitv.define (len + shift) in
    let v = append s' (static_extract (len-1) 0 (var reg)) zeros in
    extend v s unsigned
```
which corresponds to
```
 return Extend(val<len-1:0> : Zeros(shift), N, unsigned);
```

our notation is a little bit heavier: 
- `x<n-1:0>` we represent as `static_extract (n-1) 0 x`, 
- `x:y` we represent as `append s x y`, where `s` is the result sort, which in our case is `len+shift`.

You can see that many generic operators in Core Theory are parameterized by sorts which you have to provide explicitly to ensure that they are statically (from the perspective of the target language).


Side question: is your setup OK or you need any help? Like does merlin work for you? Showing type annotation, indicating no errors (when there are none), jumping to the definition (including your own definitions), etc? Indenter working automatically? It is very important that tools are not failing you so that you can focus on the main topic without spending your focus on things that can be done by automation. So if anything is wrong, just tell what and we will guide you how to set it up.


[1]: https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/aarch64-instrsextendreg-pseudocode
[2]: https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/aarch64-instrsextendreg-pseudocode#ExtendType
[3]: https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/shared-functionscommon-pseudocode#impl-shared.Extend.3
[4]: http://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Basic/index.html#val-extract

plugins/arm_a64/a64_defs.ml
ivg: Yep, it is close to be "small and easy to use") See my comment for this function. 

> Thus, if I understand what you mean correctly, your suggestion is actually let me stop merging all load variants into one function. Instead, write the ocaml/core theory version of the pseudo-code for every instruction variant(one ocaml function per instruction variant).

Exactly and perfectly worded! As close to the spec as possible. 

Also, another advice, try not to correct the spec and follow it, no matter how it looks. You might find yourself thinking, why is this expressed so ugly, we can write it nicely using this core theory operation, or this OCaml construct. And while it is in general a good idea when you write a program in our case we should refrain ourselves from doing this, as our task to express the semantics as close as possible in our meta language.

plugins/arm_a64/a64_utils.ml
zhaoyonghe: Hi @ivg, thanks for your detailed directions! 

For the setup, I use VSCode on Ubuntu and have installed a plugin just called OCaml.
>  Like does merlin work for you? Showing type annotation, indicating no errors (when there are none), jumping to the definition (including your own definitions), etc? 

These things works well. One awkward thing is when I write a function in file A and use it in file B, this function will not be recognized(with read wave underline). And the "red wave underline" will only disappear after `bapbuild`.

> Indenter working automatically?

It can automatically indent, but does not fit the ocp-indent format you need. Now I just have a shell file to run `cd ../../` then `make indent` then `cd ./plugins/arm_a64/`.

zhaoyonghe: Hi @ivg, thanks for your detailed directions! 

For the setup, I use VSCode on Ubuntu and have installed a plugin just called OCaml.
>  Like does merlin work for you? Showing type annotation, indicating no errors (when there are none), jumping to the definition (including your own definitions), etc? 

These things works well. One awkward thing is when I write a function in file A and use it in file B, this function will not be recognized(with read wave underline). And the "red wave underline" will only disappear after `bapbuild`.

> Indenter working automatically?

It can automatically indent, but does not fit the ocp-indent format you need. Now I just have a shell file to run `cd ../../` then `make indent` then `cd ./plugins/arm_a64/`.

zhaoyonghe: @ivg Another question. We have one function per instruction, but what are the parameters of the function? 

Still use instruction **`C6.2.133 LDR (register)`** as an example.  It seems that using a bit vector with 32 bits as an input is the best fit. Because we have pseudo-code like this:

**Decode for all variants of this encoding**
```
integer scale = UInt(size);
if option<1> == '0' then UNDEFINED; // sub-word index
ExtendType extend_type = DecodeRegExtend(option);
integer shift = if S == '1' then scale else 0;
```

`size`, `option` and `S` are all bits in the 32-bit instruction encoding. We can simply extract them out and use them. But the thing is, we are lifting from assembly code, which is essentially characters. So I am confused now.

By the way, how to deal with the UNDEFINED?

zhaoyonghe: > there is a mismatch between the extend_type from the reference and our representation. Why is LSL there? We need to rectify this asap.

In the description of `C6.2.133 LDR (register)` (sorry I do not know how to add a link for this), there is a field "Assembler symbols" says:

> <extend> Is the index extend/shift specifier, defaulting to LSL, and which must be omitted for the LSL option
when <amount> is omitted. encoded in the "option" field. It can have the following values:
UXTW when option = 010
LSL when option = 011
SXTW when option = 110
SXTX when option = 111

It is strange, but it does can have LSL, but at here its behavior is same as UXTX when register size is 64 bits and UXTW when 32 bits. So I write code like this [LSL](https://github.com/BinaryAnalysisPlatform/bap/blob/1d848484fc20f0d07202804a02e9eedb028d6cba/plugins/arm_a64/a64_utils.ml#L96).

zhaoyonghe: Also, when I try to implement [CheckSPAlignment](https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/aarch64-functionsmemory-pseudocode#CheckSPAlignment), I need to define a lot of things:
```
  let bv64s = Theory.Bitv.define 64
  let bvs_of_size (size: int) = Theory.Bitv.define size

  type bv1
  type bv2
  type bv4
  type bv8

  let bv_of (s: int) (n: int) = 
    let module M = Bitvec.Make(struct let modulus = Bitvec.modulus s end) in
    int (bvs_of_size s) (M.int n)

  let _EL3 = bv_of 2 3
  let _EL2 = bv_of 2 2
  let _EL1 = bv_of 2 1
  let _EL0 = bv_of 2 0

  let sctlr_el1 = Theory.Var.define bv64s "SCTLR_EL1"
  let sctlr_el2 = Theory.Var.define bv64s "SCTLR_EL2"
  let sctlr_el3 = Theory.Var.define bv64s "SCTLR_EL3"

  type proc_state = {
    _N: bv1 Theory.bitv;
    _Z: bv1 Theory.bitv;
    _C: bv1 Theory.bitv;
    _V: bv1 Theory.bitv;
    _D: bv1 Theory.bitv;
    _A: bv1 Theory.bitv;
    _I: bv1 Theory.bitv;
    _F: bv1 Theory.bitv;
    _SP: bv1 Theory.bitv;
    _EL: bv2 Theory.bitv
    (* todo more *)
  }

  let _PSTATE: proc_state = {
    _N = zero (bvs_of_size 1); (* maybe some other default value *)
    _Z = zero (bvs_of_size 1);
    _C = zero (bvs_of_size 1);
    _V = zero (bvs_of_size 1);
    _D = zero (bvs_of_size 1);
    _A = zero (bvs_of_size 1);
    _I = zero (bvs_of_size 1);
    _F = zero (bvs_of_size 1);
    _SP = zero (bvs_of_size 1);
    _EL = zero (bvs_of_size 2)
  }

  let sctlr_el1 = Theory.Var.define bv64s "SCTLR_EL1"
  let sctlr_el2 = Theory.Var.define bv64s "SCTLR_EL2"
  let sctlr_el3 = Theory.Var.define bv64s "SCTLR_EL3"

  let sctlr regime = 
    match regime with
    | _EL1 -> sctlr_el1
    | _EL2 -> sctlr_el2
    | _EL3 -> sctlr_el3
    | _ -> raise (Failure "unreachable")

  let have_el = (* TODO *)

  let s1_translation_regime el = 
    if not is_zero (logxor el _EL0) then el
    else if (* TODO *)

  let check_SP_alignment = 
    if is_zero (logxor _PSTATE._EL _EL0) then (* TODO *)
```
1. I try to model the [ProcState](https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/shared-functionssystem-pseudocode#ProcState) just like the pseudo-code, and I found it is awkward to compare two bit vectors. Now I use `is_zero (logxor a b)` as `==`. Is there any elegant way? I think it should have but I cannot find it :( 
2. I found the way to define bit vector types like ` _N: bv1 Theory.bitv;` is awkward. What do think of my defines? These are important because these things and the code style will exist everywhere in the lifter. 
3. Am I on the right way? Because I found this function seems "endless-nested" and a lot of small functions to be implemented.

ivg: > `size`, `option` and `S` are all bits in the 32-bit instruction encoding. We can simply extract them out and use them. But the thing is, we are lifting from assembly code, which is essentially characters. So I am confused now.

This part is handled by the LLVM-MC library so we are already handled with decoded parts of the instructions, see the output of `llvm-mc` or `bap-mc`.

> By the way, how to deal with the UNDEFINED? Raise a failure?

Certainly not! Any valid sequence of bits is an expected input to the lifter/disassembler. And undefined expression has [unk][1] semantics. To represent an undefined side-effect, you can use [perform][2]

[1]: https://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Core/index.html#val-unk
[2]: https://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Core/index.html#val-perform

zhaoyonghe: > This part is handled by the LLVM-MC library so we are already handled with decoded parts of the instructions, see the output of llvm-mc or bap-mc.

I know this. What I mean is that we are lifting from assembly code and the assembly code is characters like `LDR <Wt>, [<Xn|SP>, (<Wm>|<Xm>){, <extend> {<amount>}}]`. So, we still need a step to transfer these characters to those variant:
```
integer scale = UInt(size);
if option<1> == '0' then UNDEFINED; // sub-word index
ExtendType extend_type = DecodeRegExtend(option);
integer shift = if S == '1' then scale else 0;
```
And then pass them to a function like this to implement [LDR (register)](https://developer.arm.com/docs/ddi0596/f/base-instructions-alphabetic-order/ldr-register-load-register-register):
```
let ldr_register scale option extend_type shift regsize datasize = 
  bits(64) offset = ExtendReg(m, extend_type, shift); if HaveMTEExt() then
  SetTagCheckedInstruction(TRUE); bits(64) address;
  bits(datasize) data;
  if n == 31 then 
    CheckSPAlignment(); 
    address = SP[];
  else
    address = X[n];
  address = address + offset;
  data = Mem[address, datasize DIV 8, AccType_NORMAL];
  X[t] = ZeroExtend(data, regsize);
```
I just want to say this step is not close to the specification and not that obvious to be implemented(although, yes, we can interpret those variants from the characters' instruction). I want to check if this step is needed.

ivg: >   I try to model the [ProcState](https://developer.arm.com/docs/ddi0596/f/shared-pseudocode-functions/shared-functionssystem-pseudocode#ProcState) just like the pseudo-code, and I found it is awkward to compare two bit vectors. Now I use `is_zero (logxor a b)` as `==`. Is there any elegant way? I think it should have but I cannot find it :(

We have the full set of ordering operations, including equality and comparison. 

> I found the way to define bit vector types like _N: bv1 Theory.bitv; is awkward. What do think of my defines? These are important because these things and the code style will exist everywhere in the lifter.

Leading underscores are having special meaning for OCaml compiler, it means that the value is not used, so it won't trigger the warning if it is really unused. So please do not start any of your identifiers with an undescore, unless you really not going to use them. 

Also, don't pack CPU state into an OCaml structure, you're confusing here the meta language (ML) and the target language (that we model). The CPU state is a record in the target language so we shall represent it is a record in the target language. We don't have any special provisions for structures or bit-fields in Core Theory but we commonly model CPU state using a single word and access/modify its bits using extract and bit-wise logic operations. 

> let bv_of (s: int) (n: int) = 
    let module M = Bitvec.Make(struct let modulus = Bitvec.modulus s end) in
    int (bvs_of_size s) (M.int n)

In case when you don't know the modulus beforehand (which is a rare case) you could use the generic interface that works for any modulus, e.g.,
```
let small ~width x = Bitvec.(int x mod modulus width)
```

Though it is rare that you have unknown and fully parameterized witdh, in most cases, e.g., you can use the existing modulus, e.g., to create a one-bit bitvector of an integer you can use `Bitvec.(int x mod m1)`. Also, there are `Bitvec.one` and `Bitvec.zero` for such well-known constants. 

> Also, when I try to implement CheckSPAlignment, I need to define a lot of things:

Please don't implement it, we don't need that level of detail. You can either totally ignore this function or model it as,
```
let aligment_check = Theory.Effect.Sort.ctrl "Aarch64.SPAlignmentCheck"
```
and then just
```
perform alignment_check
```

This will preserve enough information for us so that we can implement this check in runtime, but won't clutter the semantics with service details. 



[1]: https://binaryanalysisplatform.github.io/bap/api/master/bap-core-theory/Bap_core_theory/Theory/module-type-Core/index.html#val-eq
[2]: https://binaryanalysisplatform.github.io/bap/api/master/bitvec/Bitvec/index.html#val-int

ivg: > > This part is handled by the LLVM-MC library so we are already handled with decoded parts of the instructions, see the output of llvm-mc or bap-mc.
> 
> I know this. What I mean is that we are lifting from assembly code and the assembly code is characters like `LDR <Wt>, [<Xn|SP>, (<Wm>|<Xm>){, <extend> {<amount>}}]`. So, we still need a step to transfer these characters to those variant:
> 
> ```
> integer scale = UInt(size);
> if option<1> == '0' then UNDEFINED; // sub-word index
> ExtendType extend_type = DecodeRegExtend(option);
> integer shift = if S == '1' then scale else 0;
> ```
> 
> And then pass them to a function like this to implement [LDR (register)](https://developer.arm.com/docs/ddi0596/f/base-instructions-alphabetic-order/ldr-register-load-register-register):
> 
> ```
> let ldr_register scale option extend_type shift regsize datasize = 
>   bits(64) offset = ExtendReg(m, extend_type, shift); if HaveMTEExt() then
>   SetTagCheckedInstruction(TRUE); bits(64) address;
>   bits(datasize) data;
>   if n == 31 then 
>     CheckSPAlignment(); 
>     address = SP[];
>   else
>     address = X[n];
>   address = address + offset;
>   data = Mem[address, datasize DIV 8, AccType_NORMAL];
>   X[t] = ZeroExtend(data, regsize);
> ```
> 
> I just want to say this step is not close to the specification and not that obvious to be implemented(although, yes, we can interpret those variants from the characters' instruction). I want to check if this step is needed.

Just to prevent any confusion, let's say it clear that we are not lifting from the assembly (which is indeed a string of characters) but from the LLVM Machine Code (MC), represented as [Disasm_expert.Basic.Insn][1], which is an algebraic data type and is an n-tuple of opcode and a number of operands that are a [sum type][2] of registers, bitvectors, and floating-point constants. So no characters are involved here. Subfields of an instruction encoding, when are relevant are usually provided as immediate operands and we commonly translate them to sum types that more or less correspond to the types in the original specification. We can translate directly in the semantic definition function if such a translation is needed only once (in that case probably even defining a type is not necessary) or we can write a helper function when the variant is used in more than one instruction.  

I  think that we should stick to `tid -> op array -> insn knowledge` as the type for all our semantic definition functions. The lifter will maintain a hashtable from strings, that represent opcodes, to semantic functions of type `tid -> op array -> insn knowledge`. There will be a driver function, that will extract from the knowledge base the MC instruction and select a proper semantic function based on opcode. It will also handle exceptions (e.g., from array indexing operations) and map them to undefined semantics (or even mark them as data in the knowledge base). 

As a side note, it is possible that LLVM decoder misses some of the fields or decodes them incorrectly, in that case, we can always retrieve the memory representation and access any bits of it. 


[1]: https://binaryanalysisplatform.github.io/bap/api/master/bap/Bap/Std/Disasm_expert/Basic/Insn/index.html
[2]: https://binaryanalysisplatform.github.io/bap/api/master/bap/Bap/Std/Op/index.html#type-t

Phosphorus15:
The works on this job should now be transferred to this repo https://github.com/Phosphorus15/arm-isa, which is an implementation & interpretation to ARM ISA's standard machine-readable instructions' semantics, and finally gets the semantics translated to BAP's Core Theory.

Currently this repo is still WIP, with a full set of functions on parse & generation done, some delicate issues on collaborating with both ARM ISA and BAP+LLVM sides still have to be addressed.

ivg:
closing in favor of #1291

Thanks a lot for this contribution I was relying on it, when I was writing the Primus Lisp-based lifter.

PR #1153, opened 2020-06-25
adds more workflows and revamps the existing one

ivg:
1. The main workflow which was run on each PR is now known as
build-dev-repo, and it is now only building the dev-repo using
oasis/ocamlbuild on the full matrix (no tests)

2. A new workflow `build-and-test` that builds using omake and runs
unit, functional, and bap-veri tests but using ubuntu with OCaml 4.07.1

3. A new workflow `check-the-style` that checks the style
independently of the build process (to keep the later clear of extra
packages brough by ocp-indent), obviously using only one system/ocaml version

4. A new workflow `build-from-opam` that builds on each push and
checks that our rolling-release testsing repository works correctly

5. A new workflow `nigtly-testing` that runs nightly on the full
matrix, installs bap from the testing repository and runs functional
tests. It is important to run it recurrently, since updates to the
upstream packages in opam may eventually break us, both syntactically
and semantically (i.e., we can fail to compile or to run).

6. A new workflow `weekly-regress` that builds the latest upstream
release of BAP to detect regressions from upstream packages.

All workflows are uploading log artifacts for ease of debugging the
failures.


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,76 @@
...
+           run: opam exec -- make
+
+         - name: Install BAP
+           run: opam exec -- make reinstall
+
+         - name: Run Unit Tests
+           run: opam exec -- make test
+
+         - name: Run Functional Tests
+           run: pam exec -- make check
```
XVilka, line 53: A typo - thus failing build. Should be `opam exec -- make check`.

PR #1164, opened 2020-07-01
splits oasis multipackages into sub-packages

gitoleg:
This PR adds next sub-packages:
- primus-exploring-scheduler
- primus-greedy-scheduler
- primus-limit
- primus-loader
- primus-mark-visited
- primus-print
- primus-promiscuous
- primus-round-robin-scheduler
- primus-wandering-scheduler
- primus-track-visited
- primus-taint
- primus-propagate-taint
- legacy-llvm


oasis/elf
```diff
@@ -1,11 +1,11 @@
 ########## Bap_elf Library ########################
-Flag elf
+Flag elf_loader
```
ivg, line 2: the file shall be also named elf_loader

ivg, line 2: the flag shall be named `elf`

ivg: I think you just forgot to remove this file.

PR #1167, opened 2020-07-02
tweaks nightly and weekly builds - do not install bap twice 

ivg:
We pin bap's repo to portably grab extra system dependencies needed for running the testsuite, but accidentally also reinstall bap.


ivg:
merging right away, since no tests will cover this on PR.

PR #1171, opened 2020-07-06
creates a separate package for the strings plugin

gitoleg:
this PR removes strings plugin from `beagle` package and introduce a new
package `bap-beagle-strings` for it


oasis/beagle-strings
```diff
@@ -0,0 +1,12 @@
+Flag beagle_strings
+  Description: Build string analysis library
```
ivg, line 2: It is not a library but a plugin. The description would be `Build the strings plugin`

PR #1178, opened 2020-07-15
enables ARM Thumb/Thumb2 and interworking 

Phosphorus15:
This PR provides a new Thumb lifter that is implemented using the new Core Theory abstraction and upgrades the disassemler engine to support interworking, i.e., switching between mutiple encodings in the same piece of memory. 

This PR also includes a few optimizations around the code that we need to implement as this PR (the interworking part) heavily impacted performace. We were 100% slower than the 2.1 version. The good news is that now we're a even a bit faster than 2.1 .

The PR also provides minimal support for Thumb v2 instructions (via recoding them to ARM instructions) and extends our disassembler backend interface with an option to pass attributes. 

There are much more details in the commit messages.


ivg:
So we're closing #1122, right?

plugins/arm_thumb/thumb.ml
```diff
@@ -0,0 +1,195 @@
...
+let run_lifter _label addr insn _mem 
+    (lifter : Bitvec.t -> Defs.insn -> Defs.op array -> unit Theory.eff) =
+  match Insns.of_basic insn with
+  | None -> raise (Defs.Lift_Error "unknown instruction")
+  | Some arm_insn -> Disasm_expert.Basic.sexp_of_full_insn insn |> Sexp.to_string |> Stdio.print_endline;
+    match Insns.arm_ops (Disasm_expert.Basic.Insn.ops insn) with
+    | Error err -> raise (Defs.Lift_Error (Error.to_string_hum err))
+    | Ok ops -> lifter addr arm_insn ops
+
+let () =
```
ivg, line 185: We missed an important check here, we need to request the `Arch.slot` and returns semantics only if the architecture is matching with thumb, otherwise this lifter will provide semantics for x86, mips, powerpc, arm, and any other architecture. The same is true for the arm lifter.

Phosphorus15:
> So we're closing #1122, right?

Yes, looks like there's no other choices.

plugins/arm_thumb/thumb.ml
Phosphorus15: Just added the check for both ARM and ARM Thumb

ivg:
Ok, the good news is that I was able to disassemble an interworking binary using your lifter and `enable-interworking` branch (see `interwork+thumb` branch. There is still a lot of work to be done, for example, we are currently missing PLT entries since we can't obtain information about their architecture (I hope to make it work soon), but you can already start working with that branch. Feel free to merge it in this branch (or rebase your branch over enable-interworking).

plugins/arm_thumb/thumb.ml
```diff
@@ -0,0 +1,209 @@
...
+              if Word.equal (Word.of_int 10 0x10b) insn 
+              then Some `tCMNz 
+              else None)
+    else None
+  | Some insn -> Some insn
+
+let run_lifter _label addr insn mem 
+    (lifter : Bitvec.t -> Defs.insn -> Defs.op array -> unit Theory.eff) =
+  match fix_cmnz (Insns.of_basic insn) mem with
+  | None -> raise (Defs.Lift_Error "unknown instruction")
```
ivg, line 191: Please, no exceptions, a lifter shall never fail - any input sequence of bytes is valid input. 
```suggestion
  | None -> KB.return Insn.empty
```

Phosphorus15:
@ivg the Thumb lifter should be fine with `pc` and branch instructions now.

plugins/arm_thumb/thumb.ml
ivg: this still needs to be addressed

Phosphorus15: We do have a `try_with` at the place which run_lifter is called, I think this should be enough to address the underlying problem
```ocaml
    (match Or_error.try_with (
         fun () -> run_lifter label addr insn mem Lifter.lift_with
       ) with
```

ivg:
with #1225 we're now finally ready to integrate this PR. I will start working on it right now.

XVilka:
Looks like CI failure is unrelated:
```
*** omake: done (2.93 sec, 0/0 scans, 237/237 rules, 1/1634 digests)
*** omake: reading OMakefiles
*** omake error:
   File setup.log.om: line 52, characters 59-65
   unbound variable: private.key
Makefile:29: recipe for target 'reinstall-libs' failed
make: *** [reinstall-libs] Error 1
Error: Process completed with exit code 2.
```

ivg:
> Looks like CI failure is unrelated:
> 
> ```
> *** omake: done (2.93 sec, 0/0 scans, 237/237 rules, 1/1634 digests)
> *** omake: reading OMakefiles
> *** omake error:
>    File setup.log.om: line 52, characters 59-65
>    unbound variable: private.key
> Makefile:29: recipe for target 'reinstall-libs' failed
> make: *** [reinstall-libs] Error 1
> Error: Process completed with exit code 2.
> ```

Yep, it is a common omake-related issue, just needs a restart.

ivg:
Looks like we're passing the tests! I will add more tests tomorrow (the armv5t-linux-gnueabi-echo file that we're now available to run), polish the code a little bit and I think we can merge it. This PR also enables partial support for thumb2 and the latest armv7 code, but with some limitations (probably something with abi, the lifted code looks correct and I have compared it with ghidra's p-code line-by-line, but we still getting a segmentation fault at `_start`). 

@Phosphorus15, you may have noticed that I removed a lot of code and removed many instructions. It is not because they are wrong, it is because due to the time limitation I had to focus on implementing the minimal subset of the instruction set that is sufficient for runnng a non-trivial application. The removed instructions are still in git so I encourage you to restore them and re-implement using the framework that I have set up in this lifter. Anyway, you did a great job on this lifter! Thanks a lot!

Phosphorus15:
> Looks like we're passing the tests! I will add more tests tomorrow (the armv5t-linux-gnueabi-echo file that we're now available to run), polish the code a little bit and I think we can merge it. This PR also enables partial support for thumb2 and the latest armv7 code, but with some limitations (probably something with abi, the lifted code looks correct and I have compared it with ghidra's p-code line-by-line, but we still getting a segmentation fault at `_start`).
> 
> @Phosphorus15, you may have noticed that I removed a lot of code and removed many instructions. It is not because they are wrong, it is because due to the time limitation I had to focus on implementing the minimal subset of the instruction set that is sufficient for runnng a non-trivial application. The removed instructions are still in git so I encourage you to restore them and re-implement using the framework that I have set up in this lifter. Anyway, you did a great job on this lifter! Thanks a lot!

Thanks for all the works for making the lifter being able to get integrated! I'll start to work on restoring the full instructions set in respect of the current framework, then.

ivg:
A small status update. The PR works, the PR works correctly (we are able to execute thumb interworking binaries without the help of any other external tools and even lift thumb2 binaries). But it is slow. We have a 33% performance loss wrt the master branch and a 100% performance loss wrt the 2.1 version. I am currently trying to reduce this performance regression (some regression is assumed as 2.2 is doing much more that 2.1 but nothing close to x2 is acceptable).

ivg (approved):

Issue #1161, opened 2020-06-30
remove symbolic information from the testsuite


ivg:
We are excluding a large and important portion of bap functionality from the tests by adding static symbolic information. I believe it is reminiscent of the bad old times, when we weren't able to setup binutils on Travis (not sure what stopped us then, though), but today we can have both radare and objdump on our runners so let's try to run as many tests as possible without relying on `--read-symbols-from`.

Issue #1135, opened 2020-06-17
implement support for  R_X86_64_GLOB_DAT relocations


ivg:
Binaries on Manjaro do not employ R_X86_64_JUMP_SLOT relocations but are calling external functions via  R_X86_64_GLOB_DAT which we in general support through the relocatable plugin, but either it is not triggered in (I think it doesn't enter the play for non-relocatable files, which is probably wrong) or is not detecting them correctly.

Additionally, `obdump` symbols provider (and the future `r2` provider) could handle them too. While `objdump` itself is confused with these relocations, we can still get them using the `-DRz` option, which will output them as
```
 9f78: R_X86_64_GLOB_DAT malloc@GLIBC_2.2.5
```

[echo.zip](https://github.com/BinaryAnalysisPlatform/bap/files/4790969/echo.zip)

_Originally posted by @Enkelmann in https://github.com/BinaryAnalysisPlatform/bap/issues/1131#issuecomment-645210970_

ivg:
@gitoleg, I want you (after you will finish your current tasks) and after #1119 is merged to look into your relocatable symbolizer and see if (1) it handles `R_X86_64_GLOB_DAT` relocations at all (I remember it was originally designed for direct jump relocations, but it could be the case that it will handle these also) and (2) if we can enable it for all binaries. Don't bother with the objdump symbolizer, I would probably look into it during #1119 or implement it later.

PR #1187, opened 2020-07-23
renovates the LLVM backend

### LINKED ISSUE #1166: remove support for older versions of LLVM
They induce a lot of support burden and the code is much more complicated than needed because of all this dependency management. It's time to clean up. I am totally ok with removing 3.4 and 3.8 but ready for a bargain. 

We should also delete the legacy-llvm backend. Was was the original reason to have it, by the way?

### LINKED ISSUE #1183: The base address is computed incorrectly for relocatable files
Since object files and dynamic libraries do not have addresses we rely on offsets instead. However, just turning an offset into an address doesn't work. We want to have the default addressing scheme that matches what other state-of-the-art tools are using, e.g., for ELF file we will be off by the size of the ELF header if we will choose just offsets.

As a workaround, we introduced a [hack][1] that subtracts the address of the first code section from the specified base. Obviously, this hack is not guaranteed to work even with the ELF files not to say about others formats, but it just happens that for elf the first code section is indeed located just after the ELF header, at least in the dynamic libraries and object files that we have seen.

[1]: https://github.com/BinaryAnalysisPlatform/bap/pull/988

### LINKED ISSUE #1189: remove the concept of relocatable files from the loader
Historically, we were designating a special class of binaries that were treated specially, the relocatable files. With time as we develop our backends we enabled relocation (via llvm-offset and bias) for all files, so there is no real need to separate them and we can remove a lot of code duplication.

What is more important, is that presumably due to some confusion, we output relocation and other special classes of entries only for relocatable files, despite the fact that they can also occur in regular files. So we loose important information here.

In the same venue I would like our backed not to ignore any information, as I noticed that we prune some relocations and symbols if they have zero addresses or some other values, that are not yet filled in by the system loader. This information shouldn't be hidden and, if necessary, filtered later.

Basically, I would like to see as few if/then/else as possible in the part that produces the binary specification. All the loading logic should go to the dissemblers and other downstream components. The loader should just provide information as is.

We started removing special treatment of relocatable files in #1187 but we need more work. Basically, this issue could be closed when we remove is-relocatable property from the specification.

ivg:
renovates the LLVM backend

Droping support of old LLVM and legacy backends
-----------------------------------------------

We drop a lot of old code (minus 3k lines of code) thus removing the support
burden and making it easier to maintain, fix, and upgrade the code.

Fixes #1166

Simplifies the implementation
-----------------------------

The remaining code base is significanly simplified. We dropped the
separation between relocatable and non-relocatable files, removed any
transformations of addresses from the LLVM backend (we now emit
absolute virtual addresses). The whole logic of transforming from the
llvm view to the bap image view now fits into a hundred lines of
code (instead of hundreds lines spread across 16 files as it was
before).

Fixes #1183
Fixes #1189

Produces more information
-------------------------------

The relocation information is now emitted for all files (not only for
relocatable). Also, removes tons of checks that were preventing our
backends from emitting valuable symbolic information.

Paves the road to #1135 and #1161


XVilka:
Relocatable files are probably PI[C|E] ELFs?
- https://codywu2010.wordpress.com/2014/11/29/about-elf-pie-pic-and-else/
- https://eklitzke.org/position-independent-executables

gitoleg:
Speaking about relocatable files, I didn't mean just files with relocations, but files that contain a code that will be relocated, i.e. code that will be linked at some different address.
https://refspecs.linuxbase.org/elf/gabi4+/ch4.intro.html

Such files can contain symbols with no address assigned. For example, in the `Elf_sym` struct, the `st_value` field contains symbol address for the executable files but can contain any other value, depending on the file type or even ABI, and we shouldn't try to use this field as an address.

https://docs.oracle.com/cd/E19683-01/816-1386/6m7qcoblj/index.html#chapter6-35166

Also, in `LLVM 3.8` (and maybe `3.4` as well), an attempt to access relocations of an executable file would fail the whole bap application, so we had to treat them separately.

Speaking about `MachO`, I still believe that something should be done:
for our `testsute/x86_64-macho-obj`, `bap` and `objdump` show different addresses for the first symbol: it's 1d0 for the former and 0 for the latter.

ivg:
The main misconception that your brought into the implementation is that relocations could only happen in the relocatable files, while the only difference is that relocatable files do not have any fixed virtual addresses so we have to give them some base and this base must be more or less in sync with the state-of-the-art tools.

As this PR shows, we can easily handle relocatable files by just making the base calculation more robust and amend it for the relocatable files. The issue that I have with the current implementation, is that we are not providing a lot of valuable information, for non-relocatable files, e.g., the rellocations itself, indirect symbols, external symbols, etc. Again, all these features commonly occur in regular binaries. But this is a topic for another PR. This should be considered done. We have fixed macho and cleanup ELF and removed lots of unnecessary code.

I will just update the testsuite and revert the disabling of the macho changes and, if GA allows, we could merge it.

PR #1188, opened 2020-07-23
enables interworking

ivg:
What is interworking
--------------------

Interworking is a feature of some architectures that enables mixing
several instruction sets in the same compilation unit. For example, arm
and thumb interworking that this branch is trying to add. See also microMIPS,

What is done
-------------

1. We add the switch primitive to the basic interface that changes the
disassembler in the current disassembling state. It is a bold move
and can have consequences, should be carefully reviewed

2. Attributes each destination in the disassembler driver state with
the architecture and calls switch every time we are going to
disassemble the next chunk of memory.

3. The default rule that extends the unit architecture to all
instructions in that unit is disabled for ARM/Thumb and is overridden
in the arm plugin with the following behavior, if an arm unit has a file
and that file has a symbol table then we provide information based on
the last bit of that symbol table (todo: we should also check for
abi), otherwise, we propagate the unit arch to instructions.

What is to be done
------------------

Next, the arm lifter shall provide a promise to compute
destinations (which itself will require destinations because we don't
really want to compute them) and provide the destination architecture,
based on the source encoding. We can safely examine any representation
of the instruction since it is already will be lifted by that moment.

See also the tightly PR #1178.

### DIFF
--- a/lib/bap/bap_project.ml
+++ b/lib/bap/bap_project.ml
@@ -92,17 +92,6 @@ module Spec = struct
     | Ok doc -> doc
 end
 
-let with_arch arch mems =
-  let open KB.Syntax in
-  let width = Size.in_bits (Arch.addr_size arch) in
-  KB.promising Arch.slot ~promise:(fun label ->
-      KB.collect Theory.Label.addr label >>| function
-      | None -> `unknown
-      | Some p ->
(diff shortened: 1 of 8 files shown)


ivg:
closing in favor of #1126 which was based on this branch.

PR #1195, opened 2020-08-03
calls radare2 directly

### LINKED ISSUE #1184: the radare2 plugin leaves unterminated processes
It is most likely an issue on the ocaml-radare2 site, and we just probably should stop using it and call radare2 manually. But let's create the issue here, to track it, this is how my ps looks like after a day of work (well after I have noticed that I am running out of the memory),

```
$ ps ax | grep radare | wc -l
205
$ ps ax | grep radare | head
  301 pts/1    S      0:00 /bin/sh /home/ivg/warehouse/radare2/env.sh /home/ivg/bin/prefix/radare2 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
  308 pts/1    S      0:00 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
  383 pts/1    S      0:00 /bin/sh /home/ivg/warehouse/radare2/env.sh /home/ivg/bin/prefix/radare2 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
  395 pts/1    S      0:00 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
  682 pts/1    S      0:00 /bin/sh /home/ivg/warehouse/radare2/env.sh /home/ivg/bin/prefix/radare2 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
  689 pts/1    S      0:00 radare2 -2 -q0 bin/arm-linux-gnueabi-echo
 2908 pts/1    S      0:00 /bin/sh /home/ivg/warehouse/radare2/env.sh /home/ivg/bin/prefix/radare2 radare2 -2 -q0 testsuite/ko/arm-can-bcm.ko
 2915 pts/1    S      0:00 radare2 -2 -q0 testsuite/ko/arm-can-bcm.ko
 2930 pts/1    S      0:00 /bin/sh /home/ivg/warehouse/radare2/env.sh /home/ivg/bin/prefix/radare2 radare2 -2 -q0 testsuite/ko/arm-can-bcm.ko
 2937 pts/1    S      0:00 radare2 -2 -q0 testsuite/ko/arm-can-bcm.ko
```

CCint @XVilka

ivg:
The ocaml-radare2 library is suffering from a few issues and it is
easier and much more efficient to call radare2 directly.

Here the list of issues that we have with ocaml-radare2
1) leaking file descriptors (they are never closed)
2) suboptimal inter-process communication (to get a 100 bytes
ocaml-radare makes a hundred system calls)
3) lingering processes (partially fixed)

In fact, we don't really need to keep the process open as we run
radare2 once and immediately close, so it is much easier for us just
to pass the command using the `-c` command line option.

@Xvilka, I saw your fix and it is not yet working, the output channels
should be closed and exhausted for waitpid to return (or NOHANG)
should be passed. Also, without NOHANG, the code that sends signals is
actually unreachable. I've started [fixing][1] this and the other issues,
but later decided that for us it is much easier and much-much more
efficient to call radare2 directly.

@gitoleg, please update the corresponding opam package and drop the
dependency on ocaml-radare2. The new implementation also lacks the
version check as I hope that the new implementation will not hang even
with the older versions of radare2. If that is true, then it will help
us with #1161. Also, please add the conf-radare2 package to the
upstream repo and add it as a depenedency to the bap-radare2. Please,
merge this PR after that.

Fixes #1184.

[1]: https://github.com/fxfactorial/ocaml-radare2/pull/13

### DIFF
--- a/oasis/radare2
+++ b/oasis/radare2
@@ -9,7 +9,7 @@ Library radare2_plugin
   CompiledObject:   best
   BuildDepends:     bap, re.pcre, regular, bap-core-theory,
                     bap-future, core_kernel, bap-knowledge,
-                    zarith, bitvec, radare2, yojson
+                    zarith, bitvec, yojson
   InternalModules:  Radare2_main
   XMETADescription: use radare2 to provide a symbolizer
   XMETAExtraLines:  tags="symbolizer, radare2"
--- a/opam/opam
+++ b/opam/opam
@@ -33,7 +33,7 @@ depends: [
   "bap-signatures"
   "result"
   "z3"
-  "radare2"
+  "conf-radare2"
   "yojson"
   "result" {= "1.4"}
 ]
--- a/plugins/radare2/.merlin
+++ b/plugins/radare2/.merlin
@@ -1,4 +1,4 @@
 PKG bap
-PKG radare2
+PKG yojson
 B ../../_build/plugins/radare2
 REC
--- a/plugins/radare2/radare2_main.ml
+++ b/plugins/radare2/radare2_main.ml
@@ -28,40 +28,37 @@ let provide_roots file funcs =
   promise_property Theory.Label.is_valid;
   promise_property Theory.Label.is_subroutine
 
-module Field = struct
-  (* unfortunately, we can't use Yojson.*.Utils because they are
-     undefined for the Yojson type *)
-
-  let member name = function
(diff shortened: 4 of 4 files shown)


ivg:
So it looks like that with our implementation the checks are passing even for the ancient version of radare2. So we can merge it as soon as @gitoleg will get conf-radare2 into the upstream and updated our bap-radare2 package with the correct dependencies.

PR #1198, opened 2020-08-05
fixes the glibc runtime check

### LINKED ISSUE #1130: the glibc-runtime heuristic is not precise enough
We have it triggered on .o and .so files, as well as on custom-made executables. We need to collect more information to establish a set of good indicators. 

Note, when glibc-runtime is triggered it will rename the target of the first jump of the entry point to `__libc_start_main` (and insert a synthetic subroutine is the destination is unresolved).

ivg:
Since #1187 (on which this PR is based) we have full information about name references so we can devise a precise check for whether we need to insert the `__libc_start_main` function, i.e., we can just look if it is referenced in the binary.

This PR also adds `format` and `is-executable` properties both to the OGRE representation and to the Core Theory. They are not used for the glibc runtime check (though meant to be used originally), but deem useful.

Fixes #1130.


ivg:
Once #1187 goes into master, I will rebase this PR.

PR #1205, opened 2020-08-11
redirects all the channels when radare2 is spawned

ivg:
We're experience random failures on the arm-symexec-crc32 test that
appears like a race condition between bap and the Tcl's expect
framework. It looks like that while the symbolic executor is able to
hack the program and emits the 'access granted' token, the expect
command is either not seeing it or is not able to stop bap. We don't
have strong evidences that it has anything to do with radare2, but our
internal testing and bisecting shows that the issues likelihood raised
with the #1195 PR merged and that if we will redirect all channels, we
have much more consistent behavior of the test.

Given this, we still have no explanation, why is this happening and
why only the arm test is showing this problem. Therefore, this PR may
fix nothing. It is still, however, a good idea to redirect the input
channel, as radare2, being an interactive program, may play tricks
with the stdin channel.


XVilka:
Radare2 indeed does affect stdin. Not sure if it's a bug though. cc @trufae @ret2libc

ivg:
Tests show that it doesn't help :(

PR #1206, opened 2020-08-11
fixes archlinux docker image

gitoleg:
After we added `conf-radare2` dependency, the `bap` installation requires `radare2` to present in the system.
Unfortunately, `opam-depext.1.1.3` doesn't properly support `archlinux`,
so we have to install `radare2` manually.


XVilka:
Unrelated to this particular case, but I also remember that `opam-depext` didn't install `radare2` at FreeBSD either: https://travis-ci.org/github/ocaml/opam-repository/jobs/716058654#L508

PR #1222, opened 2020-09-21
Fixes lifting disassembler bug for Pcmp

percontation:
There's a decoding error in some Pcmp instructions, and the same type error in Pcmp & Ppackedbinop semantics (left-shift of a vector-register-sized operand by an 8-bit shift value).

### DIFF
--- a/plugins/x86/x86_disasm.ml
+++ b/plugins/x86/x86_disasm.ml
@@ -999,9 +999,9 @@ let parse_instr mode mem addr =
           (Punpck(prefix.mopsize, elemt, order, r, rm, rv), na)
         | 0x64 | 0x65 | 0x66 | 0x74 | 0x75 | 0x76  as o ->
           let r, rm, rv, na = parse_modrm_vec None na in
-          let elet = match o land 0x6 with | 0x4 -> reg8_t | 0x5 -> reg16_t | 0x6 -> reg32_t | _ ->
+          let elet = match o land 0x0F with | 0x4 -> reg8_t | 0x5 -> reg16_t | 0x6 -> reg32_t | _ ->
             disfailwith "impossible" in
-          let bop, bstr = match o land 0x70 with
+          let bop, bstr = match o land 0xF0 with
             | 0x70 -> Bil.EQ, "pcmpeq"
             | 0x60 -> Bil.SLT, "pcmpgt"
             | _ -> disfailwith "impossible" in


percontation:
Hmm, also, `Ppackedbinop` needs the same change to lsl.

(Am I correct in thinking bil wants both sides of shifts to have the same type? If that's changed, it could just be that I'm working with older code that wants to make that assumption. Either way, the pcmp decoder error fixed here is definitely a problem.)

ivg:
> Am I correct in thinking bil wants both sides of shifts to have the same type? 

Nope, they can have different types, no problem here.

lib/.merlin
```diff
@@ -0,0 +1,180 @@
+EXCLUDE_QUERY_DIR
```
ivg, line 1: can you, please, remove this file?

ivg (changes requested):
Could you please remove the .merlin files and provide some references that identify what "some bugs" you're fixing? E.g., what is the expected behavior and how the current behavior differs from it, and how it is reflected in the ISA manual.

plugins/x86/x86_lifter.ml
```diff
@@ -481,7 +481,8 @@ module ToIR = struct
                 iv := int @@ Word.of_int ~width:byte i;
                 tmp_dst :=
                   var tmp_dst lor
-                  ((cast unsigned t_width (var elt)) lsl (var iv * int bits))
+                  ((cast unsigned t_width (var elt))
```
ivg, line 484: the changes to this file are not necessary, could you please discard them?

plugins/x86/x86_disasm.ml
```diff
@@ -999,9 +999,9 @@ let parse_instr mode mem addr =
           (Punpck(prefix.mopsize, elemt, order, r, rm, rv), na)
         | 0x64 | 0x65 | 0x66 | 0x74 | 0x75 | 0x76  as o ->
           let r, rm, rv, na = parse_modrm_vec None na in
-          let elet = match o land 0x6 with | 0x4 -> reg8_t | 0x5 -> reg16_t | 0x6 -> reg32_t | _ ->
+          let elet = match o land 0x0F with | 0x4 -> reg8_t | 0x5 -> reg16_t | 0x6 -> reg32_t | _ ->
```
ivg, line 1002: any references or explanations?

percontation:
> > Am I correct in thinking bil wants both sides of shifts to have the same type?
> 
> Nope, they can have different types, no problem here.

Ah, ok. I'll note that I think this type difference is causing primus to send our code stuff like `extract(hi=31, lo=0, <8-bit constant>)`. Should it always be the case that over-extraction is interpreted as "zero pad to recover non-existent bits"?

lib/.merlin
percontation: whoops, my bad.

plugins/x86/x86_lifter.ml
percontation: 👍

plugins/x86/x86_disasm.ml
percontation: it was a typo, this is what the author meant.

percontation: (you can check the behavior against http://ref.x86asm.net/coder64.html ; pretty clearly the case that leaves elet=reg16_t isn't supposed to be dead code :P )

ivg (approved):

ivg:
Thanks!

ivg:
> > > Am I correct in thinking bil wants both sides of shifts to have the same type?
> > 
> > 
> > Nope, they can have different types, no problem here.
> 
> Ah, ok. I'll note that I think this type difference is causing primus to send our code stuff like `extract(hi=31, lo=0, <8-bit constant>)`. 

Yes, since 2.1.0 we relaxed the type system of the Primus interpreter and it now permits binary operations, including the arithmetic operations, on operands with different sizes. To enable this, the interpreter [coerces operands to the same bitwidth][1]. We apply this coercion to all operations, even those that permit operands of different sizes, which is probably too conservative and results in unnecessary operations such as `extract(hi=31, lo=0, <8-bit constant>)` that you mentioned. If those bogus coercions introduce any issues on your side or bother you, I can remove them real quick, no problem here. Do you want to make this coercions more precise?


> Should it always be the case that over-extraction is interpreted as "zero pad to recover non-existent bits"? 

Yes, if the resulting width is larger than the original it is the same as zero-padding, e.g.,

```
# #install_printer Word.pp_hex_full;;
# x;;
- : word = 0xA5:8u
# Word.extract_exn ~hi:31 ~lo:0 x;;
- : word = 0xA5:32u
```

[1]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/lib/bap_primus/bap_primus_interpreter.ml#L406-L412

percontation:
👍 Thanks for the info.

In our particular case, removing the "bogus" coercions would actually just cause errors elsewhere in our code, because our expressions use traditional width-strictness in bitvec operands (and assume that's what Primus is feeding the handlers).

So what'd be strictly easiest for me would be if Primus sent the "correct" coercion (cast or extract), rather than omitting them. An approach like that would allow Primus to avoid breaking the API, even if operator definitions are relaxed/otherwise expanded in the future.

ivg:
> +1 Thanks for the info.
> 
> In our particular case, removing the "bogus" coercions would actually just cause errors elsewhere in our code, because our expressions use traditional width-strictness in bitvec operands (and the code assumes that Primus is as strict as it was previously).

I am not saying about removing all coercions, just removing coercions for shifts, which are not necessary from the perspective of the BIL type system, but indeed are necessary from the perspective of SMTLIB, where shifts are having sort `(op2 (_ BitVec m) (_ BitVec m) (_ BitVec m))`, so indeed there is some impedance mismatch here. 

> So what I'd suggest, is that Primus be updated to send the "correct" coercion (cast or extract), rather than omitting them. An approach like that would allow Primus to avoid breaking the API, even if operator definitions are relaxed/otherwise expanded in the future.

In other words, instead of the swiss-knife `extract` you want to have either `cast unsigned m x` when we zero extend and `extract m 0 x` when we do the narrow cast?

I can do this, but here is the caveat. Even if Primus will begin to emit more precise coercions it doesn't guarantee that someone else will not be using `extract` for zero extending. In fact, all our lifters do this and do this a lot. So anyways you have to handle extracts correctly, because this is the semantics of the `extract` operation in BIL.  For the reference, here is how extract shall be lifted to Z3, 

```  
  module Bitv = Z3.BitVector

  let extract hi lo x =
    let xs = Bitv.get_size (Expr.get_sort x)
    and ns = hi-lo+1 in
    if ns > xs
    then if lo = 0
      then Bitv.mk_zero_ext ctxt (ns-xs) x
      else
        Bitv.mk_extract ctxt hi lo @@
        Bitv.mk_zero_ext ctxt (ns-xs) x
    else
      Bitv.mk_extract ctxt hi lo x
```

See also our [symbolic executor formula][1] module that maps Primus observations to Z3 formulae, which we should probably make public.

And sorry for the late reply, I missed somehow the notification :(

[1]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/plugins/primus_symbolic_executor/primus_symbolic_executor_main.ml#L116

PR #1250, opened 2020-11-30
updates bapdoc to the latest changes in core_kernel

ivg:
re-enables back the hidden Unix module

That PR still doesn't fix the documentation building process, as we are failing with a stack overflow error. Trying to build with the master branch of odoc also didn't give any positive results. See https://github.com/ocaml/odoc/issues/539 for more information on the issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build: setup.ml
 
 .PHONY: doc
 doc:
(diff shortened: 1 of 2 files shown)


ivg:
merged in https://github.com/BinaryAnalysisPlatform/bap/pull/1249

PR #1255, opened 2020-12-03
Legacy floating point lifter

percontation:
Portions of this code that were formerly ForAllSecure proprietary
are hereby released under the terms of the MIT license.

### DIFF
--- a/lib/bap_types/bap_helpers.ml
+++ b/lib/bap_types/bap_helpers.ml
@@ -91,9 +91,7 @@ class substitution x y = object(self)
   method! map_let z ~exp ~body =
     if Bap_var.(z = x)
     then Let (z,self#map_exp exp,body)
-    else super#map_let z
-        ~exp:(self#map_exp exp)
-        ~body:(self#map_exp body)
+    else super#map_let z ~exp ~body
   method! map_var z =
     match super#map_var z with
     | Exp.Var z when Bap_var.(z = x) -> y
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ testsuite:
 	git clone https://github.com/BinaryAnalysisPlatform/bap-testsuite.git testsuite
 
 check: testsuite
-	make REVISION=7f51548cc3e -C testsuite
+	make REVISION=3eb55ac2b5085445 -C testsuite
 
 .PHONY: indent check-style status-clean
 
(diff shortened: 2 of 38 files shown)


percontation:
This is basically just the patch that @ivg formulated, but I have with proper authority edited out the notices that this is proprietary code.

I added in a couple fixes, but there's possibly still more to be made before all this is really working. (There's definitely still issues in the Mayhem code on our side that's using this FP lifter, that are blocking me from confirming that stuff largely works as intended.)

percontation:
(Side note: Thanks for all the help Ivan! I've been working off a fork with this patch already applied, so, there's no pressure from our side to merge quickly. This PR is being issued right now because I was just informed that there was value in having all this out in public sooner rather than later.)

lib/bap_types/bap_helpers.ml
```diff
@@ -91,9 +91,7 @@ class substitution x y = object(self)
   method! map_let z ~exp ~body =
     if Bap_var.(z = x)
     then Let (z,self#map_exp exp,body)
-    else super#map_let z
-        ~exp:(self#map_exp exp)
-        ~body:(self#map_exp body)
+    else super#map_let z ~exp ~body
```
percontation, line 94: This looks right to me, but, the other change that was previously present in this file was a rogue reversion so someone should confirm.

ivg:
I will add tests on Monday and will merge it then. Thanks!

lib/bap_types/bap_helpers.ml
ivg: I am confirming this change.

The other one wasn't really rogue (in a sense that it is not git that we should blame, but it was me playing with the code and accidentally reverting to polymorphic compare, which wasn't banned at that times).

percontation:
FWIW: There's some critical bug fixes to the x87 semantics that I'm working on, but, since the test seems to be SSE then that might work fine.

I feel like it'll be most efficient if I hold off on submitting those until I've got a full "working run" of the thing I'm trying to test right now (that way I can validate that my fixes are actually an improvement 😛), but @ivg what would you prefer between:

1. I'll leave this PR as is, and send a follow up PR (sometime in the next couple of weeks) with float bugfixes.
2. I can put the the current crop of bugfixes that are ready in this PR right now.
3. I can keep adding to this PR until I'm done with my current sprint of fixes, and ping when I'm no longer actively pursing short-term issues (again, that'll be up to 2 weeks)

ivg (approved):

PR #1263, opened 2021-01-04
fixes PE/COFF sections

ivg:
We were missing non-executable sections in coff files. Thanks to
@TodAmon and @ReversingWithMe for noticing this.


ivg:
The macOS build is broken again, the PR is waiting for the resolution of https://github.com/actions/virtual-environments/issues/2391

PR #1271, opened 2021-02-17
Fixed missing dependency on focal

Shivanshu10:
Hey, 
I tried fixing #1269 please review my commits
Thanks

### DIFF
--- a/.github/workflows/release-focal.yml
+++ b/.github/workflows/release-focal.yml
@@ -1,0 +1,65 @@
+name: release-focal
+
+on:
+  schedule:
+    - cron: '0 0 * * SUN' # every Saturday
+  workflow_dispatch:
+
+jobs:
+  build:
+     runs-on: ubuntu-20.04
+
+     env:
+       OPAMJOBS: 2
+       OPAMRETRES: 8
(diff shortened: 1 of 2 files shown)


ivg:
Dear @Shivanshu10, please accept my apologies for not reviewing your PR, I missed it in the notifications. Totally my fault. Please reopen it if you would like to have it finished (as well as your previous #1270). I still believe that we can make it happen without breaking backward compatibility. If it won't work, then we might consider having an alternative mechanism for focal releases. Essentially, our main problem here is where to upload the built binaries, not the build process itself. We are picking the release artifacts upload are as the most convenient location (well at least from our users perspective), but it is hard to access it from the actions. There are other locations, maybe less obvious (which we could circumvent by properly documenting them), such as docker hub or build artifacts of the github actions itself. The latter is probably the cheapest solution. 

Anyway, feel free to reopen this if you're interested in finishing this work. Also, don't hesitate to use our [Gitter][1] channel and ping me if I am not replying. My git notification panel is overwhelmed with noise. Again, apologies for not replying to this before. 

[1]: https://gitter.im/BinaryAnalysisPlatform/bap

PR #1283, opened 2021-03-18
Update bap.mli for auto-documentation

CryHavoc21:
Added a newline so that the auto-documentation will correctly format the `var entry_point` entry correctly.

### DIFF
--- a/lib/bap/bap.mli
+++ b/lib/bap/bap.mli
@@ -5512,6 +5512,7 @@ module Std : sig
 
     (** [entry_point addr] is an address from which a kernel should start *)
     val entry_point : t -> addr
+    
     (** [filename image] a name of file from which an image was
         loaded (if any) *)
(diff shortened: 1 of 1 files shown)


lib/bap/bap.mli
```diff
@@ -5512,6 +5512,7 @@ module Std : sig
 
     (** [entry_point addr] is an address from which a kernel should start *)
     val entry_point : t -> addr
+
```
ivg, line 5515:
```suggestion

```

ivg: You have inserted accidentally four whitespaces on the empty line, which makes the style checker [unhappy][1] :)

[1]: https://github.com/BinaryAnalysisPlatform/bap/pull/1283/checks#step:4:52

ivg (changes requested):

PR #1312, opened 2021-05-25
Fine-tune GitHub Actions workflows

smorimoto:


.github/workflows/release.yml
```diff
@@ -2,39 +2,40 @@ name: release
...
+        uses: actions/checkout@v2
 
-         - name: Install OCaml
-           uses: ocaml/setup-ocaml@v2
-           with:
-             ocaml-compiler: 4.11.1+flambda
+      - name: Install OCaml
+        uses: ocaml/setup-ocaml@v2
+        with:
+          ocaml-compiler: ocaml-variants.4.11.2+flambda
```
smorimoto, line 24: There are quite a few changes, but the only thing to keep in mind is here. Since v2, we need the full package name, so we need to do this.

smorimoto:
Most of the changes are just indentation... Use the "hide whitespace changes" option for easier viewing.

smorimoto:
Okay, here is a comparison. It became faster for about 10 minutes.
https://github.com/smorimoto/bap/actions/runs/875670796
https://github.com/BinaryAnalysisPlatform/bap/actions/runs/875144031

ivg (approved):

PR #1313, opened 2021-05-25
enables dune-cache

ivg:
This might make our builds faster.


smorimoto:
I already opened a PR for that! #1312

ivg:
> I already opened a PR for that! #1312

Yep thanks for it a lot, I wish I noticed it before, but at least I learned something today. Closing this one in favor of yours.

PR #1321, opened 2021-06-02
enables push flag in the publish-docker-image action

ivg:
The docker images are no longer pushed to dockerhubn after #1312.
First of all, because the secret names were accidentally changed,
and secondly, because the `push` parameter defaults to `false` in
the `v2` version of the action.


ivg:
again bypassing CI as no code was changed

PR #1332, opened 2021-07-09
adds the flatten pass

0poss:
The flatten pass will turn the BIR code into the three-address code turning each non-flat operand of an expression into the flat form (and adding more definitions). An expression is flat if it is either a variable, a constant, or an unknown expression. All other operand expressions are recursively flattened (reduced). 

To enable the pass use the `--passes` option (or short version `-p`), e.g., `--pass=flatten` or `-pflatten`.

### DIFF
--- a/plugins/flatten/flatten_main.ml
+++ b/plugins/flatten/flatten_main.ml
@@ -130,7 +130,7 @@ Config.manpage [
   `S "DESCRIPTION";
   `P "Flatten all AST in the program.";
   `S "EXAMPLE";
-`Pre {|
+  `Pre {|
   ;; input 
   #10 := 11 * (#9 + 13) - 17
   ;; output
--- a/plugins/flatten/flatten_main.ml
+++ b/plugins/flatten/flatten_main.ml
@@ -1,0 +1,150 @@
+open Bap.Std
+open Core_kernel
+include Self()
+
+exception Err of string
+
+let get_direct_typ (e : exp) : Type.t = match e with
+  | Bil.Var v -> (match Var.typ v with
+    | Type.Imm n -> Type.Imm n
+    | Type.Mem (n, m) -> Type.Mem (n, m)
+    | Type.Unk -> Type.Unk)
+  | Bil.Int w -> Type.Imm (Word.bitwidth w)
+  | _ -> raise (Err "Provided expression isn't direct")
+
+let flatten_exp (exp : exp) (blk : blk term) (before : tid) : exp * blk term =
+  let is_virtual = true in
+  let fresh = true in
+  let rec aux (exp : exp) (blk : blk term) = match exp with
+    | Bil.Load (x, y, endian, s) ->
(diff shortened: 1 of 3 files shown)


plugins/flatten/flatten_main.ml
```diff
@@ -0,0 +1,150 @@
...
+  (setq #10
+    (-
+      (* 11 (+ #9 13))
+    17)
+  )
+
+  To :
+  (setq #11 (+ #9 13))
+  (setq #12 (* 11 #11))
+  (setq #10 (- #12 17))"
```
ivg, line 146: I think it is better to use infix notation of BIL (which is more readable), and pack this into `Pre`formatted block. 

```suggestion
`Pre {|
  ;; input 
  #10 := 11 * (#9 + 13) - 17
  ;; output
  #11 := #9 + 13
  #12 := 11 * #11 
  #10 := #12 - 17
  |}
```

ivg (changes requested):
Let's do a couple of minor tweaks, like better error messaging and a nicer man page. And wait until the CI opinion. Be careful, my patches were made directly in the browser and are untested :)

ivg:
Ah, the CI is already complaining that your code is not well-formatted. Just do `make indent` in the bap folder. (It will require the `ocp-indent` tool, which you can install, if it is not, with `opam install ocp-indent`). Make sure that your switch is activated with `eval $(opam env)` before running `make ident`.

ivg:
Waiting for the ocp-indent commit. If you have trouble with it, just tell me, I can do it easily myself.

plugins/flatten/flatten_main.ml
```diff
@@ -0,0 +1,146 @@
+open Bap.Std
+open Core_kernel
+include Self()
+
+
+let get_direct_typ (e : exp) : Type.t = match e with
+  | Bil.Var v -> (match Var.typ v with
+    | Type.Imm n -> Type.Imm n
+    | Type.Mem (n, m) -> Type.Mem (n, m)
+    | Type.Unk -> Type.Unk)
```
ivg, line 10: We can simplify the var case (probably you forgot to clean it after some debugging), and we forgot an important case of a flat expression - `Unknown`. It is also flat. 
```suggestion
  | Bil.Var v -> Var.typ v
  | Bil.Unknown (_,t) -> t
```

ivg (changes requested):
We forgot one more case!

ivg:
It looks like that the last commit also removed the improperly indented code :)

ivg:
there is still some code that requires [reindenting][1] https://github.com/BinaryAnalysisPlatform/bap/pull/1332/checks?check_run_id=3030609602#step:4:56

I am unable to push directly to this PR as you probably didn't mark the "allow edits checkbox"

ivg:
I pushed the indentation commit myself to keep things moving. So don't worry about it anymore.

ivg (approved):

PR #1338, opened 2021-07-30
fixes core:eff theory

ivg:
This theory is used for debugging and introspection purposes. The bug
was in the denotation of `branch` and `repeat` operations, which were
denoted as empty denotations if any of its effects are empty. It is
now corrected.


ivg:
The CI is broken upstream, since this doesn't affect anything meaningful, merging it.

PR #1346, opened 2021-09-16
nightly tests are failing due to a bug upstream

ivg:
which is presumabely fixed but is only available in the master and is
not released.

We will start with the master branch and if it will break stick to the
last working sha.


ivg:
bypassing CI as this file is not used in it

PR #1358, opened 2021-10-23
Update LLVM backend to work with version 12

bmourad01:
I'm using Void Linux on my personal computer, but they are not maintaining older versions of the `llvm` package (current version is at `12.0.1`).

This may also work with later versions than 12, but I did not test them.


ivg (approved):

ivg:
thanks!

PR #1365, opened 2021-11-18
handles conditional instructions inside IT blocks in thumb

ivg:
When an unconditional thumb instruction belong to an IT block, it has to be turned into conditional but it must not set the flags if it is 16 bit and is not CMP, CMN, or TST (A8.8.55). This PR updates all handled t1 instructions to properly handle this constraint.

### DIFF
--- a/plugins/thumb/thumb_mov.ml
+++ b/plugins/thumb/thumb_mov.ml
@@ -17,8 +17,8 @@ module Make(CT : Theory.Core) = struct
 
   let borrow_from_sub ~rn ~rm = bit (rn < rm)
 
-  let movi8 rd x = data [
-      rd := const x;
+  let movi8 rd x = it_set rd (const x) @@ fun v -> [
+      v := const x;
       nf := if Int.(x lsr 7 = 1) then bit1 else bit0;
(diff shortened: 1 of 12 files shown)


plugins/thumb/thumb_mov.ml
```diff
@@ -29,123 +29,114 @@ module Make(CT : Theory.Core) = struct
...
       zf := is_zero (var r);
       cf := carry_from_add (var r) (var rn);
       vf := overflow_from_add (var r) (var rn) (var rm);
     ]
 
-
-  let adcs rd rn rm = with_result rd @@ fun r -> [
-      r := var rn + var rm + CT.unsigned s32 (var zf);
+  let adcs rd rn rm =
+    it_set rd (var rn + var rm + CT.unsigned s32 (var zf)) @@ fun r -> [
```
bmourad01, line 54: I think the flag to be added here should be `cf` instead of `zf`. Is that correct? The original code seems to have had this mistake.

ivg: Good catch, thanks!

PR #1366, opened 2021-11-18
fixes a bug in the `cmp` semantics

ivg:
also moves arm-specific patterns actions to the arm directory and
constrains their context (so that they do not trigger for non-arm
targets and do not fail when arm is disabled and `arm-set-encoding`
is not available).

Thanks @bmourad01 for timely catch of this bug!

### DIFF
--- a/plugins/thumb/thumb_mov.ml
+++ b/plugins/thumb/thumb_mov.ml
@@ -135,8 +135,8 @@ module Make(CT : Theory.Core) = struct
     ]
 
   let cmpi8 rd x = Theory.Var.fresh s32 >>= fun r ->
-    data @@ cmp (var rd) (const x) r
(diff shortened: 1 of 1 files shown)


plugins/thumb/thumb_mov.ml
```diff
@@ -135,8 +135,8 @@ module Make(CT : Theory.Core) = struct
     ]
 
   let cmpi8 rd x = Theory.Var.fresh s32 >>= fun r ->
-    data @@ cmp (var rd) (const x) r
+    data @@ (r := var rd) :: cmp (var rd) (const x) r
```
bmourad01, line 138: Should `r` be the result of the subtraction of `x` from `rd`?

PR #1369, opened 2021-11-19
Adds some missing `t2LDR.*i12` instructions to the Thumb lifter

bmourad01:
These were missing when I encountered them.


ivg (approved):
Awesome, thanks!

PR #1380, opened 2021-12-03
Fixes handling of fallthroughs in IR theory

bmourad01:
The previous change did not account for calls which are conditional, and thus would fail to include the fallthrough edge.


ivg (approved):
Looks great, thanks!

ivg:
the macOS failure is irrelevant and sporadic

PR #1386, opened 2021-12-09
Adds missing ARM target ABI information

bmourad01:
For 32-bit ARM targets (such as v7 and below) we were missing information about calling conventions (and their associated registers). I've looked over the AACPS document and decided to model the target based on that information.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ jobs:
     env:
       OPAMJOBS: 2
       OPAMRETRES: 8
-      VERSION: 2.4.0
+      VERSION: 2.5.0
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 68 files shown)


ivg:
I would suggest rebasing over the upstream/master branch to get a cleaner commit history (I am referring to those dozens of merge commits), also starting a fresh new branch for each new feature from upstream/master helps with that. 

And if you're using emacs you can [integrate ocp-indent][1] with it, so that you will never ever be seeing this check-your-style errors :) 

[1]: https://github.com/BinaryAnalysisPlatform/bap/wiki/Emacs

ivg (approved):
Awesome, and special thanks for adding docs!

ivg:
we have issues with the macos workers, I will try to fix it, but so far we can ignore them.

PR #1388, opened 2021-12-10
Add aliasing information for x86

fortunac:
This PR adds register aliasing information for the x86 target. `AX` is the lower half of `EAX`, which is the lower half of `RAX`, `XMM0` is the lower half of `YMM0`, etc.


ivg (approved):
Great, thanks!

PR #1389, opened 2021-12-17
Adds `insn-code` to the `Theory.Semantics` class

bmourad01:
The `Memory.slot` property is not serialized, which is a bummer. This change aims to provide a more lightweight version which can be serialized.

It may be of interest to analyst to observe the memory contents of a particular instruction (or simply to know its physical length).

### DIFF
--- a/lib/bap_core_theory/bap_core_theory_program.ml
+++ b/lib/bap_core_theory/bap_core_theory_program.ml
@@ -20,13 +20,11 @@ let word = Knowledge.Domain.optional "word"
     ~equal:Bitvec.equal
     ~inspect:Bitvec_sexp.sexp_of_t
 
-let bytes = Knowledge.Domain.optional "bytes"
+let hexcode = Knowledge.Domain.optional "hexcode"
     ~equal:String.equal
-    ~inspect:(fun s ->
-        Sexp.Atom (
-          String.to_list s |> List.map ~f:(fun c ->
-              Format.sprintf "%02X" @@ Char.to_int c) |>
-          String.concat ~sep:" "))
+    ~inspect:(Fn.compose sexp_of_string @@
+              String.concat_map ~sep:" " ~f:(fun c ->
+                  Format.sprintf "%02x" @@ Char.to_int c))
 
 let name = Knowledge.Domain.optional "name"
     ~equal:String.equal
@@ -229,14 +227,6 @@ module Label = struct
       ~public:true
       ~desc:"the program virtual address"
 
-  let bytes = Knowledge.Class.property ~package cls "label-bytes" bytes
-      ~persistent:(Knowledge.Persistent.of_binable (module struct
-                     type t = string option
-                     [@@deriving bin_io]
-                   end))
-      ~public:true
-      ~desc:"the program memory contents"
-
   let name = string_property cls "label-name"
       ~desc:"the program linkage name"
 
@@ -312,6 +302,13 @@ module Semantics = struct
       ~persistent:(Knowledge.Persistent.of_binable (module Value.Top))
       ~public:true
       ~desc:"the program semantics"
+  let code = Knowledge.Class.property ~package cls "insn-code" hexcode
+      ~persistent:(Knowledge.Persistent.of_binable (module struct
+                     type t = string option
+                     [@@deriving bin_io]
+                   end))
+      ~public:true
+      ~desc:"the program memory contents"
 
   include Self
 end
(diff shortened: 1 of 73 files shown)


lib/bap_core_theory/bap_core_theory_program.ml
```diff
@@ -221,6 +229,14 @@ module Label = struct
       ~public:true
       ~desc:"the program virtual address"
 
+  let bytes = Knowledge.Class.property ~package cls "label-bytes" bytes
```
ivg, line 232: I would suggest `insn-code` as the name of the property, `bytes` is too generic and is more about the representation of datum rather than its contents. But most importantly it is not a property of a label but of an instruction itself. The label is just a generalization of an address, so as a sanity check you can substitute "label" with an "address", e.g., "address-bytes" or "address-code". Is bytes a property of an address? Moreover, is code itself (or a sequence of bytes) is a property of an address? No. It is the property of a machine instruction. So other than changing the name, we also need to move this property to the `Theory.Semantics.cls` class. So, please move it to the `Semantics` module (in the same file).

lib/bap_core_theory/bap_core_theory_program.ml
```diff
@@ -20,6 +20,14 @@ let word = Knowledge.Domain.optional "word"
...
     ~inspect:Bitvec_sexp.sexp_of_t
 
+let bytes = Knowledge.Domain.optional "bytes"
+    ~equal:String.equal
+    ~inspect:(fun s ->
+        Sexp.Atom (
+          String.to_list s |> List.map ~f:(fun c ->
+              Format.sprintf "%02X" @@ Char.to_int c) |>
+          String.concat ~sep:" "))
+
```
ivg, line 30: Instead of `to_list + map + concat` we can use `String.concat_map`, e.g.,

```suggestion
let hexcode = Knowledge.Domain.optional "hexcode"
    ~equal:String.equal
    ~inspect:(Fn.compose sexp_of_string @@
              String.concat_map ~sep:" " ~f:(fun c ->
                  sprintf "%02x" @@ Char.to_int c))
```

lib/bap_image/bap_memory.ml
```diff
@@ -464,6 +464,12 @@ let () =
   KB.collect slot label >>|? fun mem ->
   Some (Addr.to_bitvec (min_addr mem))
 
+let () =
+  let open KB.Syntax in
+  KB.promise Theory.Label.bytes @@ fun label ->
+  KB.collect slot label >>|? fun {data; off; size} ->
+  Some (Bigstring.to_string ~pos:off ~len:size data)
+
```
ivg, line 472: Since we now moved it to the property of the semantics itself, we have to provide the semantics instead, 

```suggestion
let () =
  KB.promise Theory.Semantics.slot @@ fun label ->
  let+ {data; off; size} = label-->?Memory.slot in
  KB.Value.put Theory.Semantics.code Insn.empty @@
  Option.some @@
  Bigstring.to_string ~pos:off ~len:size data
```

lib/bap_disasm/bap_disasm_driver.ml
```diff
@@ -359,6 +359,7 @@ let label_for_mem mem =
 let new_insn mem insn =
   label_for_mem mem >>= fun code ->
   KB.provide Memory.slot code (Some mem) >>= fun () ->
+  KB.collect Theory.Label.bytes code >>= fun _ ->
```
ivg, line 362: Please remove it :)

lib/bap_core_theory/bap_core_theory_program.ml
ivg: And on the OCaml level, let's name it `code`, e.g., `Theory.Semantics.code`

ivg (changes requested):
let's move it to the theory semantics class, see the review comments for the detailed conversation.

lib/bap_disasm/bap_disasm_driver.ml
bmourad01: 👍

lib/bap_image/bap_memory.ml
bmourad01: 👍

lib/bap_core_theory/bap_core_theory_program.ml
bmourad01: 👍

lib/bap_core_theory/bap_core_theory_program.ml
bmourad01: 👍

ivg (approved):

ivg:
Looks good and I have restarted the tests, they failed on the omake racing bug. But I have the question, did you check that it works as expected? I.e., that the promise is called and the `insn-code` is stored and properly read from the serialized knowledge base.

bmourad01:
Yes, my testing indicates that it is working properly.

ivg:
awesome, thanks!

PR #1390, opened 2021-12-22
Move Flatten to Lib

philzook58:
We have some code for which we would like to use the flatten pass programmatically. This pull request moves the bulk of the functionality  from the plugin directory to bap_sema_flatten, which I attempted to make look similar to bap_sema_ssa. In addition, we need to thread the knowledge base monad over the flattener for fresh variable and tid creation. I have done this on a private branch, but I thought I'd ask how it is appropriate to organize this interface for `Bap.Std`, or whether it should be a separate pull request.


lib/bap/bap.mli
```diff
@@ -9066,6 +9072,11 @@ module Std : sig
         [def] in [blk].  *)
     val occurs : t -> after:tid -> tid -> bool
 
+    (** [flatten blk] returns [blk] in flattened form in which
+        let expressions and compound expressions are flattened
+        into sequences of simple definitions. *)
```
ivg, line 9077: I think we should elaborate a little bit on what the flattened form means, e.g.,

```suggestion
    (** [flatten blk] translates [blk] into the flattened form.
         In the flattened form, all operations are applied to variables, constants, or unknowns, i.e., 
         the operands could not be compound expressions. E.g.,
         {v
            #10 := 11 * (#9 + 13) - 17
         v}
         is translated to,
         {v
            #11 := #9 + 13
            #12 := 11 * #11 
            #10 := #12 - 17
         v}
         @since 2.5.0 *)
```

lib/bap/bap.mli
```diff
@@ -8502,6 +8502,12 @@ module Std : sig
     *)
     val compute_liveness : t -> (tid, Var.Set.t) Solution.t
 
+
+    (** [flatten sub] returns [sub] in flattened form in which
+        Let expressions and compound expressions are flattened
+        into sequences of simple definitions. *)
```
ivg, line 8508:
```suggestion
    (** [flatten sub] returns [sub] in flattened form in which all operands are trivial.
        @see Blk.flatten for more information about flattening. 
        
        @since 2.5.0 *)
```

ivg (changes requested):
Thanks! But you forgot to update the oasis files, you need to update the `oasis/bap-std` file and in the `sema` library section add `Bap_sema_flatten` to the list of internal modules. 

Also, see my documentation suggestions.

ivg (approved):

PR #1391, opened 2021-12-30
Fixes ARM/Thumb `movt` semantics

bmourad01:
The lifter was incorrectly re-using the upper 16 bits of the destination register, when the manual states that it is completely overwritten by the source operand:

"F5.1.114 MOVT
Move Top writes an immediate value to the top halfword of the destination register. It does not affect the contents
of the bottom halfword."

```
if ConditionPassed() then
EncodingSpecificOperations();
R[d]<31:16> = imm16;
// R[d]<15:0> unchanged
```

So, the previous result of lifting, say, `movt r0, #28` would be:

`R0 := R0 | 0x1C0000`

This PR changes the behavior to be:

`R0 := pad:32[low:16[R0]] | 0x1C0000`

### DIFF
--- a/lib/arm/arm_lifter.ml
+++ b/lib/arm/arm_lifter.ml
@@ -227,7 +227,8 @@ let lift_move ~encoding mem ops (insn : move_insn) : stmt list =
 
   | `MOVTi16, [|`Reg dest; _; src; cond; _wflag|] ->
     let dest = Env.of_reg dest in
-    [Bil.move dest Bil.(var dest lor exp_of_op src lsl int32 16)] |>
+    let src16 = Bil.(cast unsigned 32 (cast low 16 (var dest))) in
+    [Bil.move dest Bil.(src16 lor exp_of_op src lsl int32 16)] |>
     fun ins -> exec ins cond
   | insn,ops ->
     fail [%here] "ops %s doesn't match move insn %s"
--- a/lib/arm/arm_lifter.ml
+++ b/lib/arm/arm_lifter.ml
@@ -227,8 +227,8 @@ let lift_move ~encoding mem ops (insn : move_insn) : stmt list =
 
   | `MOVTi16, [|`Reg dest; _; src; cond; _wflag|] ->
     let dest = Env.of_reg dest in
-    let src16 = Bil.(cast unsigned 32 (cast low 16 (var dest))) in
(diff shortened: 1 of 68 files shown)


lib/arm/arm_lifter.ml
```diff
@@ -227,7 +227,8 @@ let lift_move ~encoding mem ops (insn : move_insn) : stmt list =
 
   | `MOVTi16, [|`Reg dest; _; src; cond; _wflag|] ->
     let dest = Env.of_reg dest in
-    [Bil.move dest Bil.(var dest lor exp_of_op src lsl int32 16)] |>
+    let src16 = Bil.(cast unsigned 32 (cast low 16 (var dest))) in
+    [Bil.move dest Bil.(src16 lor exp_of_op src lsl int32 16)] |>
```
ivg, line 231: It is better not to abuse casts for bit arithmetic. Casts should be used only to change the size of the word. It is much easier to analyze bit-wise operations (especially for SMT solvers) than casts, so my suggestion is to encode it as,

```
movt r0, #0x1c
{
  R0 := R0 & 0xFFFF | 0x1C0000
}
```
i.e.,

```suggestion
    let dest = Env.of_reg dest and src = exp_of_op src in
    Bil.[dest := var dest land int32 0xFFFF lor src lsl int32 16] |>
```

ivg (changes requested):
Thanks for finding it, I am surprised that this bug survived for so many years! 

The only thing is that it is better to use the good old bit operations and encoded it as 
```
rd := rd & 0xffff | (rn << 16)
```
see the suggestion in the review!

lib/arm/arm_lifter.ml
bmourad01: 👍

lib/arm/arm_lifter.ml
```diff
@@ -227,7 +227,8 @@ let lift_move ~encoding mem ops (insn : move_insn) : stmt list =
 
   | `MOVTi16, [|`Reg dest; _; src; cond; _wflag|] ->
     let dest = Env.of_reg dest in
```
ivg, line 229: It's hard to code in the browser :D 
```suggestion
```

ivg (changes requested):

ivg (approved):

PR #1394, opened 2022-01-07
Adds `Bitvec.modular` function

bmourad01:
It's a bit clumsy to construct a Bitvec instance when the bit width is
not statically known, e.g.:

```
let add n x y =
  let module M = Bitvec.Make(struct
    let modulus = Bitvec.modulus n
  end) in
  M.(x + y)
```

Instead, we can create a first-class module more concisely:

```
let add n x y =
  let open (val Bitvec.modular n) in
  x + y
```

Unrelatedly, this also adds the missing `Bitvec.M16` specialization and
its corresponding modulus.


lib/bitvec/bitvec.mli
```diff
@@ -602,4 +616,8 @@ module M32 : S with type 'a m = 'a
     will not overflow the OCaml int representation.
 
 *)
-module M64 : S with type 'a m = 'a
+module M64 : D
+
+(** [modular n] returns a module [M] which implements
+    all operations in [S] modulo the bitwitdh [n].  *)
```
ivg, line 622:
```suggestion
(** [modular n] returns a module [M], which implements
    all operations in [S] modulo the bitwitdh [n]. 
    @since 2.5.0 *)
```

lib/bitvec/bitvec.mli
```diff
@@ -21,6 +21,9 @@ val m1  : modulus
 (** [m8 = modulus 8] = $255$ is the modulus of bitvectors with size [8]  *)
 val m8  : modulus
 
+(** [m16 = modulus 16] = $65535$ is the modulus of bitvectors with size [16]  *)
```
ivg, line 24:
```suggestion
(** [m16 = modulus 16] = $65535$ is the modulus of bitvectors with size [16]
    @since 2.5.0  *)
```

lib/bitvec/bitvec.mli
```diff
@@ -582,7 +586,17 @@ module M1  : S with type 'a m = 'a
...
-module M8  : S with type 'a m = 'a
+module M8 : D
+
+
+(** [M16] specializes [Make(struct let modulus = m16 end)]
+
+    This specialization relies on a fact, that 16 bitvectors
+    always fit into OCaml integer representation, so it avoids
+    calls to the underlying arbitrary precision arithmetic
+    library.
```
ivg, line 597:
```suggestion
    library.
    @since 2.5.0
```

ivg (changes requested):
good to go, but please add the since-notes to the comment section of the newly introduced module item, the suggestions are there.

Thanks!

bmourad01:
Should be all set.

PR #1395, opened 2022-01-09
Fixes for LLVM 13/14 compatibility

EmmaJaneBonestell:
Fixes LLVM 13 and LLVM 14 compatibility as of the current master branch.

* LLVM 13 -- llvm::MCContext now requires the TargetTriple ; https://reviews.llvm.org/D101462
* LLVM 13 -- FPImm has been split into two functions, DFPImm retains original semantics ; https://reviews.llvm.org/D77384
* LLVM 14 -- Support/TargetRegistry.h has been moved to MC/TargetRegistry.h ; https://reviews.llvm.org/D111454

LLVM split FPImm functions into DFPImm and SFPImm. getDFPImm still returns FPImmVal, but they are now both of type uint64_t, not double. Likewise, SFPImm functions, and getSFPImmVal's returned SFPImmVal, are a uint32_t, not a float. It looks like other changes may be needed in different parts of BAP's code, or if not, perhaps something like adding the following would be enough:
```
if (mcop.isSFPImm()) {
            op.type = bap_disasm_op_fmm;
            op.fmm_val = bit_cast<float>(mcop.getSFPImm());
            return op;
        }
 ```

I believe the bit_cast to double is correct, from looking at LLVM source, but I am uncertain. 

This review link for the FPImm split (above) should show whatever you may need to resolve it further, if anything.

### DIFF
--- a/lib/bap_llvm/llvm_disasm.cpp
+++ b/lib/bap_llvm/llvm_disasm.cpp
@@ -5,13 +5,18 @@
 #include <llvm/MC/MCRegisterInfo.h>
 #include <llvm/Support/DataTypes.h>
 #include <llvm/Support/FormattedStream.h>
-#include <llvm/Support/TargetRegistry.h>
 #include <llvm/Support/CommandLine.h>
 #if LLVM_VERSION_MAJOR >= 12
 #include <llvm/Support/Process.h>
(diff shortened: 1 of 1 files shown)


ivg (approved):
LGTM! Thanks a lot!

PR #1404, opened 2022-01-14
improves the peformance of the byte patterns matcher (1/3)

ivg:
By precomputing all positions and storing them in arrays.

This is the first PR in a series of optimizations PRs. Once all are merged
bap will be faster by 40%


ivg:
correction: the third PR was broken so the actual improvement will be 30% from the two PRs. I will still revisit the fixed-point computation as it looks like it is the hot point.

PR #1406, opened 2022-01-14
improves the knowledge base fixed-point computation (3/3)

ivg:
By storing the computation state in a flat map, instead of a map of maps.


ivg:
okay, it looks like that it gave performance improvement only due to an incorrect implementation (the omission of the computation deletion), the correct implementation is infinitely slower, as it requires filtering the whole map.

PR #1414, opened 2022-01-30
Fix pc value in pc-relative thumb ldr

thestr4ng3r:
For reference: https://developer.arm.com/documentation/ddi0487/latest
on page 4252:
> When executing a T32 instruction, PC reads as the address of the current instruction plus 4.

then later on page 4743 in the pseudocode for "LDR (literal)":

> `base = Align(PC,4);`

Here are some concrete examples while debugging on real hardware. Notice especially which value from the `thumb_data` array has been loaded into `r0` after executing the `ldr` instructions. The `;-- pc:` flag shows the "typical" notion of a program counter, not pc+4 or so.
Bildschirmfoto 2022-01-30 um 11 31 43
Bildschirmfoto 2022-01-30 um 11 31 51
Bildschirmfoto 2022-01-30 um 11 32 01
Bildschirmfoto 2022-01-30 um 11 32 08
Bildschirmfoto 2022-01-30 um 11 32 14
Bildschirmfoto 2022-01-30 um 11 32 21

### DIFF
--- a/plugins/thumb/thumb_main.ml
+++ b/plugins/thumb/thumb_main.ml
@@ -111,9 +111,10 @@ module Thumb(CT : Theory.Core) = struct
       !!Insn.empty
 
 
-  let lift_mem pc opcode insn =
+  let lift_mem addr opcode insn =
     let module Mem = Thumb_mem.Make(CT) in
     let open Mem in
+    let pc = W32.(addr + int 4) in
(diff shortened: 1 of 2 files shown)


ivg (approved):
looks good, thanks! 

I am currently working on adding thumb tests.

PR #1416, opened 2022-01-31
Reimplements x86 bitscan and popcnt

bmourad01:
The previous implementation did a naive test of every single bit. This resulted in a giant chain `Ite` expressions, which is probably hard on SMT solvers and at the very least does not seem very readable to me.

This reimplements the lifting of `bsr`, `bsf`, `lzcnt`, `tzcnt`, and `popcnt` instructions based on the branchless algorithms described in Hacker's Delight, chapters 5-3 and 5-4:

http://index-of.es/Security/Addison%20Wesley%20-%20Hackers%20Delight%202002.pdf

as well as the efficient implementation of Hamming weight:

https://en.wikipedia.org/wiki/Hamming_weight#Efficient_implementation


lib/x86_cpu/x86_types.ml
```diff
@@ -111,7 +111,7 @@ type opcode =
   | Shiftd of binop * typ * operand * operand * operand
   | Rotate of binop * typ * operand * operand * bool (** left or right, type, src/dest op, shift op, use carry flag *)
   | Bt of typ * operand * operand
-  | Bs of typ * operand * operand * direction
+  | Bs of typ * operand * operand * direction * bool
```
ivg, line 114: This module is exposed to the public interface so changing it will require us to bump BAP's version to 3.0.0.

Could we do this without breaking the interface? I see a few options, like handling them on the modern lifter, but I would first like to hear your opinion.

bmourad01: I think the prefix information is passed to the `to_ir` function, so I could infer it that way.

ivg (dismissed):
looks like magic) LGTM!

PR #1423, opened 2022-02-07
reimplements clz using the branchless/loopless algorithm

ivg:
Inspired by Benjamin Mourad (@bmourad01) implementation in #1416.

It is not only efficient but also helps with bap-veri-fication, as the latter do not support while loops (which is used in the previous implementation).

@bmourad01, could you please review this? No rush at all.


bmourad01:
Looks good to me!

PR #1424, opened 2022-02-08
Stur instructions

Thomas-Malcolm:
STURXi
STURWi
STURHHi
STURBBi


Thomas-Malcolm:
This was a, uhhh... GH mishap.

ivg:
No worries :) But keep in mind that this kind of PR would be very welcome and we wait for it eagerly!

PR #1433, opened 2022-02-14
adds mode events to traces

thestr4ng3r:
A mode event expresses a switch of some processor mode affecting the
execution of all following instructions. The possible values and their
meanings depend on the target. At the moment, the only one that uses
modes is arm, where the mode indicates whether or not thumb mode is
used.

In order to make use of it, it can be provided in `Mode.slot` of
`Theory.program`, such that the computation `Theory.Label.encoding`
can access it, if available.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -64,6 +64,7 @@ B lib/bap_sema
 B lib/bap_strings
 B lib/bap_types
 B lib/bap_core_theory
+B lib/bap_traces
 B lib/graphlib
 B lib/monads
 B lib/ogre
--- a/lib/arm/arm_target.ml
+++ b/lib/arm/arm_target.ml
@@ -7,6 +7,7 @@ open KB.Syntax
 open Poly
 
 module CT = Theory
+module Mode = Bap_trace_event_types.Mode
 
 type r128 and r80 and r64 and r32 and r16 and r8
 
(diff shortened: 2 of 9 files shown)


thestr4ng3r:
An alternative would be to just add an encoding field here, as part of `code_exec` events:
https://github.com/BinaryAnalysisPlatform/bap/blob/5b1acc00804a8090c09f6785138e50a42cf00827/lib/bap_traces/bap_trace_event_types.ml#L11-L16
That would be much simpler and would mean "the encoding of this instruction" rather than "switch to a different encoding mode". Depends on what we actually want specifically...

ivg:
> An alternative would be to just add an encoding field here, as part of `code_exec` events:
> 
> https://github.com/BinaryAnalysisPlatform/bap/blob/5b1acc00804a8090c09f6785138e50a42cf00827/lib/bap_traces/bap_trace_event_types.ml#L11-L16
> 
> 
> That would be much simpler and would mean "the encoding of this instruction" rather than "switch to a different encoding mode". Depends on what we actually want specifically...

The chunk representation is fixed, we do not want to break protocols and interfaces by changing it.

ivg (approved):
Awesome! Proceeding to the next PR ;)

PR #1457, opened 2022-04-01
Adds optional `join` for `Knowledge.Domain.mapping`

bmourad01:
It may be useful to define a mapping where the codomain can grow monotonically. This PR adds an optional `join` function to facilitate this (by default it will just use the `equal` predicate).


lib/knowledge/bap_knowledge.mli
```diff
@@ -1469,6 +1469,7 @@ module Knowledge : sig
       ('a,'e) Map.comparator ->
       ?inspect:('d -> Base.Sexp.t) ->
       equal:('d -> 'd -> bool) ->
+      ?join:('d -> 'd -> 'd option) ->
```
ivg, line 1472: It looks like the common type for the `join` function in KB is `'a -> 'a -> ('a,conflict) result`, so could you please update to keep consistent? And can you please move `join` up one line so that all optional parameters are grouped together? 

Also, can you please add a `@after` note that indicates that `join` is only available in 2.5.0 and after?

ivg (changes requested):

ivg (approved):

PR #1458, opened 2022-04-05
Fix aarch64 bitmask immediate encoding and implement some aarch64 instructions

DukMastaaa:
### File restructuring
To manage growth of file sizes when implementing more of the ARMv8 ISA, this PR splits the instructions implemented in `aarch64.lisp` into smaller files based on category:
- arithmetic
- atomic (`CAS`, `CSEL` and the like)
- branch
- data movement (load, store, move)
- logical
- processor state
- special (barriers, etc)

A full list of the instructions belonging to each category has been developed [here](https://github.com/UQ-PAC/bap/wiki/All-aarch64-Instructions-by-Category).

### aarch64 Bitmask Immediate decoding
Before this PR, running `bap mc --arch=aarch64 --show-bil -- 20 00 7f 92` (assembly instruction `and x0, x1, 2`) incorrectly lifted the instruction:
```
{
  R0 := R1 & 0x7C0
}
```
The constant `0x7C0` appears in place of `2` due to ARMv8's "bitmask immediates" which encode immediate values for the logical instructions `AND`, `ORR` and `EOR`. This PR implements helper functions in Primus Lisp which decode these immediates using the algorithm from the ARMv8 ISA pseudocode, and updates the semantics of those logical instructions. Helper functions specific for aarch64 such as these have been moved to `aarch64-helper.lisp`.

### Barriers
This PR represents the `DMB`, `DSB` and `ISB` barriers as external function calls using the new `(special)` primitive implemented in #1410. For example, running `bap mc --arch=aarch64 --show-bil --show-bir -- bf 3b 03 d5` (instruction `dmb ish`) gives
```
{
  call(special:barrier:dmb:ish)
}
0000000c:
0000000b: goto @special:barrier:dmb:ish
```
### Load-acquire and store-release
This PR represents load-acquire and store-release semantics using the `(special)` primitive like barriers described above. It's used in the implementation of `CAS*` instructions. For example, running `bap mc --arch=aarch64 --show-bil -- 41 fc e0 88` (instruction `casal w0, w1, [x2]`) gives
```
{
  #0 := mem[R2, el]:u32
  call(special:load-acquire)
  #1 := #0 = extract:31:0[R0]
  if (#1) {
    call(special:store-release)
    mem := mem with [R2, el]:u32 <- low:32[extract:31:0[R1]]
  }
  R0 := #0
}
```
All `CAS*` instructions except the `CASP*` family are implemented in this PR.

### Other instructions
Here are all the instructions that have been implemented in this PR.
- **Arithmetic**: `SDIV`, `UDIV`, `MADD`, `MSUB`
- **Atomic**: `CSEL`, `CSINC`, `CSINV`, `CSNEG`, `CAS*` except `CASP*`
- **Branch**: `TBZ`, `TBNZ`
- **Data movement**: `STUR`, `STURH`, `STURB`
- **Logical**: `AND*ri`, `ORR*ri`, `EOR*ri`
- **Processor state**: `CCMN`, `CCMP`, `RMIF`, `SETF8`, `SETF16`
- **Special**: `DMB`, `DSB`, `ISB`

### Possible issues
Update: the below issues have been fixed with #1461; these instructions now work correctly.

~~LLVM can't seem to disassemble the `RMIF`, `SETF8` and `SETF16` instructions (try `bap-mc --show-bil --arch=aarch64 -- "00 04 00 ba"` corresponding to `rmif x0, 0, 0`). These three are part of ARMv8.4; perhaps LLVM does not implement this version? Since `--show-knowledge` output gives no information, the LLVM opcodes for these instructions in `aarch64-pstate.lisp` may be incorrect, and the semantics have not been tested yet.~~

~~Another ARMv8.4 instruction, `CFINV`, does get disassembled but gets turned into _MSR (register)_ even though CFINV isn't an alias. `--show-knowledge` reports that LLVM returned `msr 0x200 XZR` for `CFINV` but this isn't even a valid opcode; both arguments of _MSR (register)_ need to be registers. Not sure what's going on here.~~


ivg:
The tests are failing with `Unresolved function or primitive `aarch64:CASordXr'` and indeed I can only find a definition of `CASordX`. Maybe a typo?

Besides, you can run tests using `make test`

ivg:
See #1461, it should enable instructions such as RMIF, SET8, CASX, and others that are available only in armv8.1a through v8.6a. Also, if you enable the "allow edits from the maintainers" checkbox on the pull request, then I can push fixes directly to your branch. But if you have reasons to keep it unchecked, then no worries, I can create pull requests.

plugins/arm/semantics/aarch64-atomic.lisp
```diff
@@ -0,0 +1,41 @@
...
+   acquire-ordering and release-ordering are booleans indicating whether
+   load-acquire and store-release ordering is to be enforced."
+   (let ((data (load-word rn)))
+    (when (acquire-ordering) (special :load-acquire))
+    (when (= data rs)
+      (when (release-odering) (special :store-release))
+      (store-word rn rt))
+    (set$ rs data)))
+
+(defun CASX   (rs rt rn) (CASordXr rs rt rn false false))
```
ivg, line 19: maybe it should be `CASordX` without the trailing `r`?

plugins/arm/semantics/aarch64-pstate.lisp
```diff
@@ -0,0 +1,69 @@
...
+(defun CCMPWi (rn imm nzcv cnd) (CCMP** rn imm nzcv cnd))
+(defun CCMPXi (rn imm nzcv cnd) (CCMP** rn imm nzcv cnd))
+(defun CCMPWr (rn rm  nzcv cnd) (CCMP** rn rm  nzcv cnd))
+(defun CCMPXr (rn rm  nzcv cnd) (CCMP** rn rm  nzcv cnd))
+
+;; CFINV gets turned into MSR (register) by LLVM even though they're not aliases.
+;; --show-knowledge shows that LLVM returned "MSR 0x200 XZR" for CFINV, but
+;; this isn't even a valid opcode? For MSR (register) both arguments need to be registers..
+
+;; (defun CFINV () (set CF (lnot CF)))
```
ivg, line 38: after #1461 `CFINV` will be lifted correctly and the signature matches, so you can uncomment this.

plugins/arm/semantics/aarch64-pstate.lisp
```diff
@@ -0,0 +1,69 @@
...
+;; (defun MSR () )
+
+;; LLVM can't seem to disassemble the RMIF instruction at all;
+;; try `bap-mc --show-bil --arch=aarch64 -- "00 04 00 ba"`
+;; which corresponds to "RMIF x0, 0, 0".
+;; I also don't know whether "RMIF" is the correct LLVM opcode to
+;; use as the function name.
+;; It also can't do SETF8 and SETF16, and all three are from ARMv8.4.
+;; I suspect ARMv8.4 support has not been implemented yet.
```
ivg, line 47: also fixed with #1461

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
          (logand (msb rn) (lnot (msb rd)))))
 
-
 (defun overflow (rd rn rm)
   "(overflow RD RN RM) is true if the sum RD = RN + RM results
    in two's complement overflow."
   (logor (logand (msb rn) (msb rm) (lnot (msb rd)))
          (logand (lnot (msb rn)) (lnot (msb rm)) (msb rd))))
 
+(defun highest-set-bit (bitv)
```
ivg, line 28: We now have `clzN` in `init.lisp` that is implemented using branchless/loopless algorithm (but only works on immediates of up to 64 bits. You can use it, instead. If you think that we can have larger immediates, then we need to implement a generic clz that depending on the word-size either call the optimized versions or fall-back to the while-loop version.

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
+   Translated from ARMv8 ISA pseudocode."
+  (assert-msg (not (is-zero bitv)) "highest-set-bit bitv is zero")  ; at least 1 bit must be set
+  (let ((i (- (word-width bitv) 1)))
+    (while (and (> i 0) (= (select i bitv) 0))
+      (decr i))
+    i))
+
+(defun replicate (bitv n)
+  "(replicate bitv n) returns a bitvector with bitv repeated n times.
+   Translated from ARMv8 ISA pseudocode."
```
ivg, line 40: The `bits.lisp` module is the part of the standard Primus Lisp library, so I don't think that references to the ARM manual are necessary or appropriate here. Only generic functions should be kept here. The `replicate` function is generic enough so we can just drop the confusing reference to the ARM manual.

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
+  (let ((output 0:0))
```
ivg, line 41: while this works it is on the edge of Primus Lisp semantics. Strictly speaking, `0:0` is an ill-formed bitvector and it is ill-typed (the set of zero-width bitvectors is uninhabited). Eventually, this will  be fixed in the implementation and this code will stop working. So my suggestion is to rewrite it so that the initial `output` is equal to `bitv` with the updated loop condition `(> n 1)`

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
+    (while (> n 0)
+      (decr n)
+      (set output (concat output bitv)))
+    output))
+
+(defun replicate-to-fill (bitv n)
```
ivg, line 47: I think that this function is too specific and should be moved to the arm-bits.lisp. We already have `replicate` and this function is just a specific wrapper around it that makes sense for implementing arm instructions.

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
+  "(replicate-to-fill bitv n) returns the result of repeating bitv
+   to a total of n bits. Requires that n is a multiple of bitv's length.
+   Modified from the bits(N) Replicate(bits(M) x) function from
+   ARMv8 ISA pseudocode."
+  (let ((bitv-length (word-width bitv)))
+    (assert-msg (= 0 (mod n bitv-length)) "replicate-to-fill n not multiple of len(bitv)")
+    (replicate bitv (/ n bitv-length))))
+
+(defun zeros (n)
```
ivg, line 56: This could be done much more efficiently with `(cast-unsigned n 0:1)` the same is true for `ones`, e.g., `(cast-signed n 1:1)`.

plugins/primus_lisp/semantics/bits.lisp
```diff
@@ -19,13 +19,87 @@
...
+  "(zeros n) returns an empty bitvector of length n.
+   Modified from ARMv8 ISA pseudocode."
+  (replicate 0:1 n))
+
+(defun ones (n)
+  "(ones n) returns a bitvector of length n with all bits set.
+   Modified from ARMv8 ISA pseudocode."
+  (replicate 1:1 n))
+
+(defun zero-extend (bitv result-length)
```
ivg, line 66: We already have `cast-unsigned` primitive for that.

DukMastaaa:
Ah sorry, I didn't realise that by pushing new commits to the branch, those commits would also be part of the PR 😔
Will fix this up now.

DukMastaaa:
> Also, if you enable the "allow edits from the maintainers" checkbox on the pull request, then I can push fixes directly to your branch.
Unfortunately it seems like [this is only possible](https://github.com/isaacs/github/issues/1681#ref-pullrequest-1187750761) for branches made from a user account, not an organisation account like UQ-PAC. Pull requests definitely welcome, though!

plugins/arm/semantics/aarch64-atomic.lisp
DukMastaaa: yep 👍

plugins/arm/semantics/aarch64-pstate.lisp
DukMastaaa: yes

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: ok, the immediates should only be at most 64 bits wide.

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: done 👍

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: ok

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: ok

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: oh, that makes sense.

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: Signed extension doesn't do what we want here; `(cast-signed 4 1:1)` gives `0x9` which is `0b1001`, not `0b1111`. I may just invert `(zeros n)`.

DukMastaaa:
Here's some things I noticed while fixing up the code:

1. Can the code for `SETF8` be optimised in any way? Currently, the code outputs the following BIL for `setf8 w0`:
```
{
  NF := extract:7:7[extract:31:0[R0]]
  ZF := extract:7:0[extract:31:0[R0]] = 0
  VF := extract:8:8[extract:31:0[R0]] ^ extract:7:7[extract:31:0[R0]]
}
```
  Maybe we can move the extract:7:0 into a local variable or something?

2. Is the output of the bitmask immediate code correct? Given the instruction `and w0, w1, 3`, BIL output is
```
{
  R0 := extend:64[extract:31:0[R1]] & 0x300000003
}
```
  Just wanting to confirm that the leading `3` there will still zero out the upper 32 bits of R0, since we've zero-extended the bottom 32 bits then applied AND to the mask. This weird `30000...` occurs for all immediate values with W registers but not X.

3. Every time I invoke `bap mc` to test lifter output, it tells me that `'+v8.6a' is not a recognized feature for this target (ignoring feature)`. Is this anything to be worried about?

plugins/primus_lisp/semantics/bits.lisp
ivg: This is a bug, `(cast-signed 4 1:1)` must give `0b1111`. I will provide the fix soon!

ivg: The fix is out #1462

ivg:
> 3\. Every time I invoke `bap mc` to test lifter output, it tells me that `'+v8.6a' is not a recognized feature for this target (ignoring feature)`. Is this anything to be worried about?

It looks like v8.6a is not supported by your version of LLVM. We need to be more careful when adding the features and consider the llvm version, I will work on the fix.



> 2\. Is the output of the bitmask immediate code correct? Given the instruction `and w0, w1, 3`, BIL output is

Yep, looks like that it doesn't work for some reason for 32-bit version of instructions, see how llvm encodes them,
```
echo "and w0, w1, 3" | llvm-mc -triple aarch64 --show-encoding --show-inst
	.text
	and	w0, w1, #0x3                    // encoding: [0x20,0x04,0x00,0x12]
                                        // <MCInst #784 ANDWri
                                        //  <MCOperand Reg:186>
                                        //  <MCOperand Reg:187>
                                        //  <MCOperand Imm:1>>
                                        
```
vs.
```
echo "and x0, x1, 3" | llvm-mc -triple aarch64 --show-encoding --show-inst
	.text
	and	x0, x1, #0x3                    // encoding: [0x20,0x04,0x40,0x92]
                                        // <MCInst #786 ANDXri
                                        //  <MCOperand Reg:217>
                                        //  <MCOperand Reg:218>
                                        //  <MCOperand Imm:4097>>
```

So it looks like that  ANDWri and ANDXri have different encodings, so we can represent both with `log*ri`. I am currently trying to figure out what kind of encoding is used on the `W` versions. 


> 1\. Can the code for `SETF8` be optimised in any way? Currently, the code outputs the following BIL for `setf8 w0`:

Are you talking about nesting extracts that could be optimized away?

ivg:
PR UQ-PAC/bap#4 will resolve the conflicts and will make the decoding closer. It looks like we need to properly cast the immediate to 32 bits to get the correct result, though I am not very sure.  Passing on the baton to you)

plugins/primus_lisp/semantics/bits.lisp
DukMastaaa: yep, using this now

DukMastaaa:
Thanks for the PR, it's merged in now.
After comparing the ISA pseudocode to the LLVM decoding code you sent, the issue was the result needs to be replicated to fill the width of the register we're operating on. I hardcoded `memory-width` as `64`, but I've modified the functions to take `register-width` as an argument now like LLVM and it works correctly.

ivg (approved):
LGTM, thanks!

ivg:
Thanks again, guys, for an awesome contribution!

PR #1460, opened 2022-04-07
Fix compatibility with Core_kernel >= 0.15

bcc32:
Core_kernel 0.15 introduces a Sys module (which is a deprecated value that refers users to core_unix.sys_unix).  Also, Result now contains an Error module (which is a monad over the Error case rather than the Ok case).

### DIFF
--- a/lib/bap/bap_self.ml
+++ b/lib/bap/bap_self.ml
@@ -7,6 +7,7 @@ open Cmdliner
 
 module Event = Bap_main_event
 module Buffer = Caml.Buffer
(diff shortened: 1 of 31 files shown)


ivg:
Thanks, @bcc32! Could you please resolve the conflicts?

bcc32:
Thanks @ivg.  I was using this PR as a place to put a patch for nixpkgs, which is currently at v2.4.0, but I should be able to upgrade that once

ivg:
thanks a lot!

PR #1470, opened 2022-04-16
Correct linking of Unix library in configure

dra27:
Silences the alert proposed in ocaml/ocaml#11198, but this PR can be merged regardless.


ivg:
Thanks, I also replicated this change to the `configure-omake` that has exactly the same invocation. Will merge as soon as the tests are passed.

PR #1489, opened 2022-05-20
Adds some missing functionality to Primus Lisp POSIX

bmourad01:
- The thread local storage structure (on SysV x86-64) had some
additional data to set up

- Added `gets` and `strchrnul` to `stdio` and `string`, respectively


ivg (approved):
awesome!

Issue #1354, opened 2021-10-10
Monad Transformer Stack


addap:
Hi,

I was wondering how to compose multiple monad transformers from your transformer library.
I'm using version 2.3.0 from opam.

The naive approach does not work
```
open Monads.Std
module R = Monad.Reader.Make(struct type t = string end)(Monad.Ident)
module SR = Monad.State.Make(struct type t = string end)(R)
```
because R is missing a definition for type t, so it does not conform to the Monad signature.

I saw in ogre.ml that you can hack around this by defining t in a new module 
https://github.com/BinaryAnalysisPlatform/bap/blob/51b4988a1ee154922a81e8a98acb8fc8661f54f1/lib/ogre/ogre.ml#L808-L813

But that is not mentioned in the docs as far as I can tell. So is that the recommended way to build a monad transformer stack using this library?

ivg:
@addap, are you satisfied with an answer? Maybe you have any suggestions on how we can improve it? Starting from documentation and ending in implementing, `MakeT` for each monad that will make compositions easier? PRs are very welcome!

PR #1510, opened 2022-06-23
adds an example on how to create a monad transformer stack (#1354)

addap:
Hi,
I forgot about issue #1354 which I opened last year, but now I finally wanted to finish it.
I tried implementing it with a higher-order functor that includes the T1 & Make modules, but I do not think that this will work.
```
module MakeMakeT (TYPE_F: functor(T: T1) -> sig ??? end) (F: (M: Monad.S) -> sig ??? end) (M: Monad.S) : Monad.S = struct
    include TYPE_F(M)
    include F(M)
end

(* in module Option/List/... *)
MakeT(M) = MakeMakeT(T1)(Make)(M)
```

For `TYPE_F` the resulting signature is mostly the same for different monads (i.e. one that includes types `t`, `m` & `e`) but for `F` the signatures are of course very different between monads since they have different functions. And iiuc you cannot pass a signature itself to a functor. So I think this approach does not work.

Therefore I just added a small explanation on how to create a transformer stack based on how I used the library in the end.

### DIFF
--- a/lib/monads/monads.mli
+++ b/lib/monads/monads.mli
@@ -230,6 +230,27 @@ open Core_kernel[@@warning "-D"]
     Each monad transformer creates a module that has functions [lift]
     and [run]. The [lift] function lifts original computation into the
     transformed one, and [run] will run the computation.
+
+    In order to use the monad transformers to build up a monad
+    transformer stack, the inner module needs both the monad types and
+    monad functions, so that it satisfies the [Monad.S] signature.
+    Therefore, it needs to include both [T] and [Make] (or [T2] and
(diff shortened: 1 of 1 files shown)


ivg:
Sounds good! And yeah, it looks like that indeed, the higher-order functors will not really work here as we don't have (a) common abstraction for all monads (other than monad itself, which is too abstract) and (b) don't have polymorphism on the functor level. So the only option that is left requires tedious and repetitive work.

ivg (approved):

ivg:
Resolves #1354

PR #1511, opened 2022-06-23
Adds more missing Thumb instructions

bmourad01:
Thanks to `--print-missing` I was able to fill these in.

This also adds the `i-shift`  helper to the `arm-bits` file, which will make it easier to deal with register-shifted operands.


ivg (approved):

ivg:
Awesome, thanks!

PR #1513, opened 2022-06-27
x86_disasm: support padd instructions

heyitsanthony:
Adds disassembly to packed binop for padd{b,w,d,q}.


heyitsanthony:
Hi,
A number of targets have been running into `paddq` and skipping the missing semantics. I've been ignoring it but I added a cross-checker that's saying it's the root cause of some bogus segfaults so I thought I'd try to fix it. Please let me know if there's somewhere I should add a test; I couldn't find one for psub.
Thanks!

ivg (approved):
Awesome, thanks!

PR #1537, opened 2022-07-12
Fixes lifting of Thumb `cbz` instruction

bmourad01:
It was checking `rn <> 0` when it needs to check `rn = 0`.


ivg:
Congrats on getting the first PR after the release the second time in a row))

PR #1545, opened 2022-08-01
KB-friendly functions in `Bap.Std`

bmourad01:
This PR proposes a KB-friendly interface for a variety of functions that operate over the old BAP 1.x data structures. Initially, this was motivated by needing to use `Program.lift` in a Knowledge computation without resorting to ugly hacks with the toplevel state. To this end we also expose the `Symtab.create` function in the user-facing API.

This change further motivated KB-friendly `Term.{map,filter_map,filter}` operations. Finally, it made sense to also KB-ify the SSA and flatten passes since they create tids and variables using the top-level KB, which may or may not play nicely when we decide to use these in the middle of a Knowledge computation.


lib/bap/bap.mli
```diff
@@ -7875,6 +7874,10 @@ module Std : sig
     (** empty symbol table  *)
     val empty : t
 
+    (** [create disasm calls] creates the symbol table given the
+        disassembly state [disasm] and callgraph [calls]. *)
```
ivg, line 7878: a missing `@since` annotation

lib/bap/bap.mli
```diff
@@ -8416,6 +8419,17 @@ module Std : sig
         not a valid position number.  *)
     val nth_exn : ('a,'b) cls -> 'a t -> int -> 'b t
 
+    module KB : sig
```
ivg, line 8422: You can put a single `@since` to the whole module.

lib/bap/bap.mli
```diff
@@ -8416,6 +8419,17 @@ module Std : sig
...
+      (** @since @2.6.0 *)
+      val map : ('a,'b) cls -> 'a t -> f:('b t -> 'b t knowledge) -> 'a t knowledge
```
ivg, line 8424: needs some minimal documentation

lib/bap/bap.mli
```diff
@@ -8636,6 +8650,11 @@ module Std : sig
       val result : t -> program term
     end
 
+    module KB : sig
```
ivg, line 8653: also, needs the module documentation together with the `since` annotation

lib/bap/bap.mli
```diff
@@ -8834,6 +8853,17 @@ module Std : sig
       val result : t -> sub term
     end
 
+    module KB : sig
+      (** @since 2.6.0 *)
+      val lift : block -> cfg -> sub term knowledge
```
ivg, line 8858: this definitely needs documentation.

lib/bap_sema/bap_sema_lift.ml
```diff
@@ -192,61 +201,66 @@ module IrBuilder = struct
...
-        ]
-      | Some (`Inter dst) ->
-        if is_call || has_call x then
-          let next = Ir_blk.create () in
-          let fall = Ir_jmp.resolved (Term.tid next) in
-          let next = insert_inter_fall dst next in [
-            next;
-            fall_if_possible fall @@
-            turn_into_call (Some fall) x
+    begin concat_map_fst_and_rev blks @@ function
```
ivg, line 207: Those begin/end blocks make the code really unreadable. It was already asking for some refactoring so could you please refactor them into functions?

ivg (changes requested):
Looks good, but could you please add the docstrings to modules and functions and get rid of the `begin/end` blocks by refactoring them into functions?

bmourad01:
I removed the behavior of `fix_names` when updating with the `sub_t` class, but I'm wondering where this should be noted in the documentation.

ivg (approved):
Thanks, that's a lot of work and I am very happy that it is finally done, it gives us so many opportunities!

PR #1547, opened 2022-08-10
Adds the `libraries` parameter to the `disassemble` command

bmourad01:
It may be useful to link two BIR programs together, especially when running dynamically linked binaries with Primus. In particular, it may be useful to execute external library code without writing Primus Lisp stubs, if the exact behavior of this code is relevant to the analysis task at hand.

The `libraries` parameter allows the user to load any libraries that may be dynamically linked at runtime. Each of these binaries are loaded as `Theory.Unit.t` objects into the Knowledge Base, disassembled, and then linked together into a final `program term`. We can then re-use the functionality of the `stub-resolver` to redirect stubs to their library implementations.

Example usage:

```
bap /bin/ls --libraries libc.so.6
```

To do this, we had to rework the `Project.t` data structure under the hood to refer to the "main" binary, as well as the "libs" that were loaded in relation to this binary. This distinction is (or should be) made explicit in the documentation as well as the naming of some new APIs in `bap.mli`.


bmourad01:
There are some questions to consider:

1. Would it be a problem if `mybinary` and `mylibrary`, have overlapping virtual address spaces? e.g. Function `foo` in `mybinary` and function `bar` in `mylibrary` both have the address `0x40000`?
2. The API/ABI resolver passes look like they have the `autorun` attribute, but they only run once when we load `mybinary` prior to linking. It seems that currently this information doesn't get resolved for the rest of the units that we lift (such as `mylibary`). Should we override this behavior so that we can run these passes again?

bmourad01:
Updated with the new approach after some discussion.

ivg (changes requested):
the first round is ready (I will meantime look deeper into the implementation)

lib/bap/bap.mli
```diff
@@ -10467,27 +10471,61 @@ module Std : sig
     *)
     val arch : t -> arch
 
+    (** [arch_of_lib project unit] reveals the architecture of
+        a loaded library, if it exists.
+
+        @since 2.6.0
+        @deprecated use [target_of_lib project unit] instead.
+    *)
+    val arch_of_lib : t -> Theory.Unit.t -> arch option
```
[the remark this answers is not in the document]
bmourad01: Done!

plugins/disassemble/disassemble_main.ml
```diff
@@ -353,7 +353,12 @@ let print_missing =
           that do not have semantics." "print-missing"
 
 let input = Extension.Command.argument
-    ~doc:"The input file" Extension.Type.("FILE" %: string =? "a.out" )
+    ~doc:"The input file" Extension.Type.("FILE" %: string =? "a.out")
+
+let link = Extension.Command.parameter
+    ~doc:"The input libraries to link with"
+    Extension.Type.("FILES" %: list string)
+    "link"
```
ivg, line 361: Let's pick the `--libraries` as the default name for the parameter, with an alias `-l`. Also, let's turn it into `parameters` (plural) so that we can repeat the option in a conventional manner. I would prefer to safe the `link` name for a possible `link` plugin (I am envisioning that at some point of time we will move stub resolver there).

plugins/disassemble/disassemble_main.ml
```diff
@@ -487,19 +492,22 @@ let save_knowledge ~had_knowledge ~update digest = function
     Knowledge.save (Toplevel.current ()) path
   | Some _ -> ()
 
-let create_and_process input outputs passes loader target update
+let create_and_process input link outputs passes loader target update
     kb print_missing ctxt =
   let uses_file_loader = Sys.file_exists loader &&
                          Fn.non Filename.is_implicit loader in
+  let link = List.dedup_and_sort link ~compare:String.compare in
```
ivg, line 499: We already are doing this in project.ml

plugins/disassemble/disassemble_main_rules.ml
```diff
@@ -172,13 +172,15 @@ module Symbols = struct
           })
       ~unmatched:(fun reason t -> match reason with
           | Non_injective_fwd (addrs,name) ->
-            info "the symbol %s has ambiguous addresses: %a@\n"
+            info "the symbol %s has ambiguous addresses: %a@"
```
ivg, line 175: `@\n` is Format's new line, why did you remove `\n`? If you think that a new line is unnecessary here from the aesthetic perspective, then please remove the `@` as well.

ivg (changes requested):
a couple of minor issues and we are good to go!

plugins/dependencies/dependencies_main.ml
```diff
@@ -367,6 +367,7 @@ let ldconfig = Extension.Command.parameter
 
 let root = Extension.Command.parameter
     Extension.Type.(some string) "root"
+    ~aliases:["R"]
```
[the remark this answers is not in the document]
bmourad01: The `-r` option was taken by `--recursive`, so for `--root` I figured `-R` would be appropriate.

plugins/disassemble/disassemble_main.ml
bmourad01: I was doing it here so that, no matter the order (or duplicates), it would lead to the same digest. But I see now that semantically the order matters so it would make sense to remove it.

plugins/dependencies/dependencies_main.ml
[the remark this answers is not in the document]
ivg: okay

lib/bap/bap_project.ml
```diff
@@ -261,7 +261,16 @@ module Input = struct
...
+  let dedup =
+    let equal = String.equal and compare = String.compare in
+    fun l ->
+      let rec loop res = function
+        | [] -> res
+        | x :: xs ->
+          let dups = List.find_all_dups (x :: xs) ~compare in
+          let res = if List.mem dups x ~equal then res else x :: res in
+          loop res xs in
+      loop [] (List.rev l)
```
ivg, line 273: Let's do something that is not quadratic :) 

```suggestion
let dedup xs =
  List.rev @@ fst @@
  List.fold xs ~init:([],String.Set.empty) ~f:(fun (xs,mems) x ->
      if Set.mem mems x then (xs,mems)
      else (x::xs,Set.add mems x))
```

lib/bap/bap_project.ml
```diff
@@ -449,7 +458,7 @@ let compute_units ?package ?state main libs =
         Term.enum sub_t prog |> Seq.fold ~init:p ~f:(Term.append sub_t),
         {unit; info} :: l) in
   set_package package >>| fun () ->
-  main, libs, prog
+  main, List.rev libs, prog
```
ivg, line 461: Not sure why it is there, you already reversed it in the dedup function.

plugins/disassemble/disassemble_main_rules.ml
```diff
@@ -232,7 +232,7 @@ module Symbols = struct
     KB.Rule.(begin
         declare "provides aliases" |>
         require Image.Spec.slot |>
-        provide Theory.Label.possible_name |>
+        provide Theory.Label.aliases |>
```
ivg, line 235: good catch!

ivg (changes requested):

lib/bap/bap_project.ml
bmourad01: The libraries actually are accumulated in reverse order when we process them one-by-one, so to preserve the original order we reverse again. Does this make sense?

ivg (approved):

PR #1552, opened 2022-09-16
Fixes parsing of `defsubst` when a syntax is specified

bmourad01:
In Primus Lisp, we can use the `defstubst` keyword to define the following:

```
(defsubst my-string :ascii "i am a string\000")
```

The Primus Lisp parser is meant to see the `:ascii` token and then desugar the string into a list of hex bytes. However, the parser currently does not drop this first token and subsequently fails to desugar the string. Furthermore, the string must be unquoted before it can be processed.

This also affects the use of the `:hex` keyword for a similar reason.


ivg (approved):
Thanks! I am surprised that ascii and hex substitutions were never used in our code base. It looks like this bug was there from the very beginning.

PR #1555, opened 2022-10-10
Fix BFM in ARM Primus Lisp semantics.

katrinafyi:
Previously, the BFM macro did not account for register widths while extracting and concatenating and also did not concatenate some parts of the old destination register.

First, we'll show the old behaviour.

```asm
bfm w3, w17, #3, #5
R3 := 31:29[R3].5:3[R17]
```
Here, r = 3 and s = 5 for the s >= r case. This should take bits from the middle of R17 and places them at the lower bits of R3. Slice of R17 is correct but slice of R3 is only 3 bits wide when it should be 29 bits wide.

```asm
bfm w3, w17, #6, #4
R3 := low:5[31:0[R17]].low:6[31:0[R3]]
```
Now, r = 6 and s = 4 for the s < r case. This should take the lower bits of R17 and inserts them into R3. Again, slice of R17 is correct but the slice of R3 is too small. This is also missing a concat with R3's upper bits which were not filled by the R17 slice.

----

This PR fixes these, matching the written description in the [A64 ISA](https://developer.arm.com/documentation/ddi0602/2022-03/Base-Instructions/BFM--Bitfield-Move-?lang=en).

```asm
bfm w3, w17, #3, #5
R3 := 31:3[R3].5:3[R17]

bfm w3, w17, #6, #4
R3 := 31:30[R3].low:5[31:0[R17]].low:25[31:0[R3]]
```

The make-eBFM macro was found to be mostly correct, except for its treatment of 32-bit registers which was fixed as well. Additionally, the special cases in the existing eBFM were also logically unreachable so we removed them.

### DIFF
--- a/plugins/arm/semantics/aarch64-logical.lisp
+++ b/plugins/arm/semantics/aarch64-logical.lisp
@@ -46,16 +46,12 @@
 (defmacro make-eBFM (set cast xd xr ir is)
   "(make-BFM  cast xd xr ir is) implements bitfield move instructions
    accepting either a W or X register, with cast being an unsigned or signed cast."
-  (let ((rs (word)))
+  (let ((rs (word-width xd)))
     (if (< is ir)
-        (if (and (/= is (- rs 1)) (= (+ is 1) ir))
-            (set xd (lshift xr (- rs ir)))
-          (set xd (lshift
-                   (cast rs (extract is 0 xr))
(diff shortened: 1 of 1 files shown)


ivg:
awesome, thanks for catching this!

ivg (approved):

PR #1556, opened 2022-10-18
Fix not propagating term attrs in `mangle_names` during lifting

bmourad01:
We mangle names in `bap_sema_lift.ml` when there are conflicting subroutine names. However, it creates a new subroutine term (to be KB friendly), and we were forgetting to preserve the old attributes dictionary in this new subroutine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ testsuite:
 	git clone https://github.com/BinaryAnalysisPlatform/bap-testsuite.git testsuite
 
 check: testsuite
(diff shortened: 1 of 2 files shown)


ivg (dismissed):

ivg:
@bmourad01 to fix the tests make a PR to [bap-testsuite][1] and after it is merged, update the revision [here][2] :)

[1]: https://github.com/BinaryAnalysisPlatform/bap-testsuite
[2]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/Makefile#L58

PR #1557, opened 2022-10-19
Fix alloc size when passing `--run-env`

bmourad01:
A typo made it so that we didn't allocate enough space for the `environ` table.


ivg (approved):

ivg:
Thanks, I remember I had problems with environment variables not working properly, but I didn't have time to investigate it then. Now it explains why they didn't work!

PR #1558, opened 2022-10-31
Corrected primus_dictionary_main.ml

A-Benlolo:
Corrected dict-length to map to Length module instead of First module.


ivg (approved):
Thanks!

PR #1559, opened 2022-11-10
Allows libraries to be allocated at non-conflicting address spaces

bmourad01:
In #1547, we added the capability for BAP to load multiple shared libraries into a single project. The problem is, we left out the behavior where these libraries should be loaded at address spaces that don't overlap with each other (or the main executable). On top of that, the memories of each library weren't being loaded by Primus, which limited the usefulness of this feature.

This PR allows BAP to, on the fly, decide where to load the libraries in memory. To do this, we extended the `Image.Loader` interface to be parameterized by the Knowledge Base, such that we can suggest a `Theory.Unit.bias` for the library file being loaded.

While we're at it, we can extend the LLVM ELF loader to look at data and `R_*_RELATIVE` relocations, such that we can fix them up in the Primus loader. The motivation for this is that undoubtedly the shared libraries will need to be relocated, so accesses to relocation symbols and so on during a Primus run will need them to be fixed in order to get the correct behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ testsuite:
 	git clone https://github.com/BinaryAnalysisPlatform/bap-testsuite.git testsuite
 
 check: testsuite
-	make REVISION=ddb5e33ce1c -C testsuite
+	make REVISION=5909f59c5c29a7b143573bd49b8101a68ccded32 -C testsuite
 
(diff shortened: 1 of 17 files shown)


ivg (approved):
Awesome work, thanks! It looks good to me to merge. 
Sorry that it take so much time to review!

PR #1564, opened 2022-12-22
Fixes the stub resolver tests

bmourad01:
The tests didn't reflect the new behavior of the stub resolver from #1547. The idea is that now we can have multiple stubs link to the same implementation, as is common when linking more than one shared library.

The algorithm is also cleaned up to use ~~`Bap_relation`~~ `Graphlib`, which ends up being simpler and produces better results.

### DIFF
--- a/plugins/stub_resolver/stub_resolver.ml
+++ b/plugins/stub_resolver/stub_resolver.ml
@@ -23,7 +23,7 @@ module G = Graphlib.Make(Regular_string)(Unit)
 
 type state = {
   graph : G.t;
-  names : (tid, string) Bap_relation.t;
+  names : Tid.Set.t String.Map.t;
   stubs : Tid.Set.t;
   units : Theory.Unit.t Tid.Map.t;
 }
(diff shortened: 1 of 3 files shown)


plugins/stub_resolver/stub_resolver.ml
```diff
@@ -2,21 +2,30 @@ open Core_kernel[@@warning "-D"]
...
+module G = Graphlib.Make(Regular_string)(Unit)
 
 type state = {
-  groups : groups;
-  names  : names;
-  next   : int;
-  stubs  : Tid.Set.t;
-  units  : Theory.Unit.t Tid.Map.t;
+  graph : G.t;
+  names : (tid, string) Bap_relation.t;
```
ivg, line 26: It's confusing why we are using Bap_relation here if we are just (ab)using it as a plain mapping. Can you please, switch to regular `string list TId.Map.t` here?

ivg (changes requested):
a small request, can you now drop bap-relation as we are no longer really using it?

ivg (approved):
Nice! Thanks a lot and Happy Holidays 🎄🎅🏻⛄!!

Issue #1561, opened 2022-11-14
OCaml 5.0 support / Dune


kit-ty-kate:
bap is not currently compatible with OCaml 5.0 due to its use of the deprecated OASIS tool.
During last release @ivg told me you were working on a switch to dune (https://github.com/ocaml/opam-repository/pull/21781#issuecomment-1184488487)

I’m opening this issue just to track the progress of this given the full release of OCaml 5.0 is currently planned for early next month.

ivg:
It is partially implemented but requires more testing and the final release. The estimated amount of work is about a week. We plan to integrate and include it in the next release. Here is the link to [the version of bap][1] with full support of dune. It is currently blocked by this [PR][2] to dune (without it we can't load plugins in top-level). 

[1]: https://github.com/ivg/bap/tree/bap-dune 
[2]: https://github.com/ocaml/dune/pull/6082

kit-ty-kate:
thanks for the update. I've pinged the dune devs to see if they could have a look at it.

greedy:
What is left blocking a new release with support for OCaml 5? Is there something where contributions would help?

Josef-Thorne-A:
Seconding the above comment. It seems the dune related requirements have been merged to dune now. If there's anything else preventing OCaml 5 support I would be happy to assist.

PR #1565, opened 2022-12-23
uses dune as the main build system

ivg:
Switches to dune as the main build system but keeps support for OASIS with ocamlbuild and omake backends.

The implementation utilizes the dune sites and plugins system, but it still supports plugins that are built with bapbuild. To ease the transition, the OASIS-built BAP is also able to load dune plugins. Later the OASIS build system will be removed from BAP but the support for bundled with bapbuild plugins will remain.

Since dune sites impose their own restrictions we had to introduce a new bap-common package that manages all the sites, starting from plugins and ending with semantic files.

related to #1561


bmourad01 (approved):
Looks good, I'm glad this is happening.

Minor request: one thing that I think would be helpful is if the README was updated with some notes on creating plugins with the new build system.

bmourad01 (approved):

bmourad01 (approved):

PR #1567, opened 2023-01-03
allows bypassing checks when storing memory, uses it for relocations

ivg:
When we perform relocations on object files, we do the linker work not the loader, therefore we do not need to check if the memory is writable or even loadable. The easiest option is to extend the Primus memory interface with primitives that perform writing without any checks.


bmourad01 (approved):
Looks good to me.

PR #1569, opened 2023-01-11
deduplicates dune plugins and ignores empty plugins

ivg:
Removes duplicating packages, provided that they have the same implementation.

When dune executes a binary (either via `dune exec` or `dune test`) it sets an internal `DUNE_DIR_LOCATIONS` environment variable, which affects the selection of plugins in bap, even if bap is installed from an indepenent repository. I am still not sure if this is a bug or a feature of dune, but to prevent double-loading we have to deduplicate plugins. We silently remove duplicating names only if all implementations that match the name are the same. We compare implementation using the sum of md5 sums of files that comprise the plugin folder. The choice of md5 sum as the implementation witness instead of parsing the META files allows us to preserve the abstraction of dune plugins, as the contents of the folder is the implementation detail of dune. This increases the probability that this code will still work when they will change the implementation.

In addition, this change workarounds another subtle bug in the dune plugin system. When a plugin is removed with `opam remove`, opam leaves an empty folder in the plugin folder, which we should ignore, otherwise, dune plugin loader will fail on it.

### DIFF
--- a/lib/bap_plugins/bap_plugins.ml
+++ b/lib/bap_plugins/bap_plugins.ml
@@ -306,9 +306,57 @@ module Plugins = struct
         | Ok p -> Some p
         | Error _ -> None)
 
+  (* returns a sorted deduplicated list of dune plugin names.
+
+     dune sites might have duplicating folders, so we have to be
+     careful that we do not try to load the same plugin twice. *)
   let list_packages () =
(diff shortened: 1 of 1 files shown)


bmourad01 (approved):
Looks good, and it fixed the issue we were having.

PR #1571, opened 2023-01-11
fixes installation from opam

ivg:
When bap installed from the testing repository it fails on frontends due to a missing dependency

```
Bap_plugins_loader_backend referenced from
/home/runner/work/bap/bap/_opam/lib/bap-plugins/bap_plugins.cmxa(Bap_common)
```

Also improves the opam workflow and adds functional checks.


ivg:
Also shall fix the docker pusher workflow.

PR #1578, opened 2023-02-07
fixes a bug in instruction encoding propagation

bmourad01:
In binaries with interworking, it is expected that the destinations of some control flow instructions will have a different encoding than said instructions (e.g. `blx` on ARM/Thumb). However, the rules in the disassembler were too restrictive, as they assumed that the destination encoding would have a join with the source encoding.

This PR relaxes the restriction by assuming at first the `Theory.Language.unknown` encoding. If no encoding can be found then we can then assume that the source encoding is to be used.


ivg (approved):

ivg:
I changed the title to reflect that it is not just a relaxation but actually a bug fix :)

PR #1580, opened 2023-02-17
Fixes a typo in the Primus Lisp IEEE754 semantics

bmourad01:
They were using `cast-sfloat/cast-float` instead of `cast-sint/cast-int`.


ivg (approved):
👍🏻

PR #1582, opened 2023-03-03
Fixes a few bugs in the Thumb semantics

bmourad01:
The helper `set-nzcv-from-registers` calls additional helpers `overflow` and `carry`, which compute their results from the addition of two integers. However, the `cmp` instruction computes the subtraction of two integers, and this wasn't reflected in the usage of this helper function.

Also, the `t2ADDrs` semantics had a typo in it.


ivg (approved):
thanks!

PR #1584, opened 2023-04-25
Corrects array-set macro for Primus Lisp.

A-Benlolo:
Corrects the array-set macro for Primus Lisp by adding the type to the call to ptr+.


ivg (approved):

ivg:
thanks!

PR #1593, opened 2023-09-08
Support linking with LLVM shared libraries

drvink:
This must be manually enabled by setting the variable `llvm-shared` to a true value in the current opam switch, similar to the existing `llvm-config` variable used by conf-bap-llvm.

* bap_llvm/config/llvm_configurator.ml: use `llvm-config --shared-mode` to get the appropriate linking mode

* bap_llvm/llvm_disasm.cpp: add missing include for Optional.h

### DIFF
--- a/bap-llvm.opam.template
+++ b/bap-llvm.opam.template
@@ -3,6 +3,7 @@ build: [
   [
   "ocaml" "tools/configure.ml"
   "--with-llvm-config=%{conf-bap-llvm:config}%"
+  "--%{llvm-shared?disable:enable}%-llvm-static"
   ]
   [
     "dune"
--- a/lib/bap_llvm/config/llvm_configurator.ml
+++ b/lib/bap_llvm/config/llvm_configurator.ml
@@ -30,9 +30,13 @@ let args = [
 
 
 let () = C.main ~args ~name:"bap-llvm" @@ fun self ->
+  let linkmode =
+    "--link-" ^
+    (String.strip @@
+     C.Process.run_capture_exn self llvm_config ["--shared-mode"]) in
   C.Flags.write_sexp "link.flags" @@ List.concat [
-    llvm self ["--link-static"; "--ldflags"];
-    llvm self (["--link-static"; "--libs"] @ llvm_components);
+    llvm self [linkmode; "--ldflags"];
+    llvm self ([linkmode; "--libs"] @ llvm_components);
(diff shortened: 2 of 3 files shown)


lib/bap_llvm/config/llvm_configurator.ml
```diff
@@ -30,9 +30,13 @@ let args = [
 
 
 let () = C.main ~args ~name:"bap-llvm" @@ fun self ->
+  let linkmode =
+    "--link-" ^
+    (String.strip @@
+     C.Process.run_capture_exn self llvm_config ["--shared-mode"]) in
```
ivg, line 36: I wonder if the `--shared-mode` command-line option is present on all llvm versions that we support? Maybe, it is better to be more defensive here (both against previous and future versions) and anticipate possible failures by sticking with the static mode by default.

bap-llvm.opam.template
```diff
@@ -3,6 +3,7 @@ build: [
   [
   "ocaml" "tools/configure.ml"
   "--with-llvm-config=%{conf-bap-llvm:config}%"
+  "--%{llvm-shared?disable:enable}%-llvm-static"
```
ivg, line 6: wow, didn't know about this syntax, is it new? (I mean will it work with opam 2.0)

ivg (commented):
Some questions in the code. And the main question: it is not clear to me that the static mode is the default (which I would like to have). Is that right? We need statically linked and self-contained bap. Of course, having an ability to dynamically link with llvm would be nice (especially since some distributions do not distribute statically linked llvm) as we do have with the OASIS and OMake build systems.

bap-llvm.opam.template
drvink: Yes, this was added in opam 1.2.1.

lib/bap_llvm/config/llvm_configurator.ml
drvink: It exists at least as far back as llvm 4.0. I don't know if there's a great solution here in general, since the installed version of llvm could change since bap was built, and then we might have a mismatch of the linking mode anyway.

drvink:
Is anything preventing this from being merged?

ivg (approved):

ivg:
> Is anything preventing this from being merged?

No, not really. There was an issue with the build action, but it looks like it wasn't the problem of this PR, I will look at it later! Thanks for the contribution!

PR #1595, opened 2023-11-20
Fix compatibility with LLVM 16

katrinafyi:
LLVM 16 uses std::optional in place of llvm::Optional. Since these have very similar interfaces, we use auto to maintain backwards compatibility.

I haven't tested much, but it compiles successfully with LLVM 16 and LLVM 15.


ivg:
covered by #1597

PR #1597, opened 2024-02-15
Fixes for LLVM 17 (and 16)

bmourad01:
This also covers #1595.

There was a really tricky segfault to track down in the call to `MCInstrDesc::mayAffectControlFlow`, see my comment as well as [here](https://github.com/llvm/llvm-project/blob/release/17.x/llvm/include/llvm/MC/MCInstrDesc.h#L200). This comment is also present in [LLVM 16](https://github.com/llvm/llvm-project/blob/release/16.x/llvm/include/llvm/MC/MCInstrDesc.h#L200).

### DIFF
--- a/lib/bap_llvm/llvm_disasm.cpp
+++ b/lib/bap_llvm/llvm_disasm.cpp
@@ -30,6 +30,9 @@
 #include <limits>
 #include <typeinfo>
 #include <iostream>
+#if LLVM_VERSION_MAJOR >= 16
+#include <optional>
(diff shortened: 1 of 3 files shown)


ivg:
Thanks! The build on linux is failing with `No space left on device`  on z3. I will take a look at it tomorrow to see if we can squeeze anything out of the runner. However, the build on macOS fails with llvm, it looks like an issue with the packager, will try addressing it in a separate PR.

Issue #1589, opened 2023-06-23
monads: `core_kernel.rope` no longer exists in v0.16 


bmourad01:
It seems this functionality has been separated into the `jane_rope` package. Would we need to break compatibility with previous versions of janestreet libraries to upgrade, or hard stop at v0.15?

ivg:
Do you mean by a hard stop that we shouldn't support versions *below* `v0.15` or *above*?

XVilka:
It's better to support newer Core/Base versions in my opinion, since they will benefit from all the OCaml 5.x multicore and effects work in the future, which would fit BAP quite well for variety of reasons.

greedy:
FWIW it is possible to support both using a disjunctive dependency formula and dune's [alternative dependency](https://dune.readthedocs.io/en/stable/reference/library-dependencies.html#alternative-dependencies) mechanism.

For example, I checked with

    ( "core_kernel" {>= "v0.14" & < "v0.16"}
    | ("core_kernel" {>= "v0.16" & < "v0.17"} & "jane_rope" {>= "v0.16" & < "v0.17"}) )

in monads.opam.

Then with

```
(library
  (name monads)
  (public_name monads)
  (wrapped false)
  (preprocess (pps ppx_bap))
  (libraries core_kernel
             (select rope_witness.ml from
                     (core_kernel.rope -> rope_witness.core_kernel.ml)
                     (jane_rope -> rope_witness.jane_rope.ml)))
  (private_modules monads_monad monads_monoid monads_types rope_witness))
```

in lib/monads/dune I could build the monads library against either core_kernel 0.15 or core_kernel 0.16 + rope 0.16.

The rope_witness.*.ml files don't have any code in them, they just are there to make the dune facility work. I put comments in them so that I could see in the build tree which file was selected for rope_witness.ml.

Not sure if it is really worth the complication though :)

PR #1602, opened 2024-04-13
Upgrades BAP to janestreet v0.16.0 libraries

bmourad01:
Related: #1589 

This is here as a placeholder for now in case we want to upgrade to v0.16.0 version of `base`, `core`, `core_kernel`, and other related libraries.

Several interfaces have been changed/broken as far as I can see, namely:

- `X.Map`/`X.Table` have been gutted of various functions that are present in `Map` and `Hashtbl`, respectively
- `Binable.Of_binable`, `Binable.Of_stringable`, and `Bin_prot.Make_binable` are gone and now have `{with,without}_uuid` and other variants. The `without_uuid` is marked as "legacy".
- `Caml` is now deprecated in favor of `Stdlib`

Right now some of the PowerPC unit tests are failing, so I will try to investigate later when I have time.


bmourad01:
@ivg pinging just to make you aware of this

ivg (approved):
I went through all the changes, they look pretty syntactical. I wonder what could get wrong that ppc tests fail.

XVilka:
@bmourad01 do you plan to update the PR?

bmourad01:
> @bmourad01 do you plan to update/finish the PR?

I believe it should be ready now. I put extra constraints on the versions of the libraries (we don't want to allow `v0.17` out of the gate, as it will exclude OCaml 4. Later in the future we can have more discussion about what needs to be done to move BAP to be ready for OCaml 5 (but it seems that this would be a one way street).

ivg:
> Later in the future we can have more discussion about what needs to be done to move BAP to be ready for OCaml 5 (but it seems that this would mean abandoning support for OCaml 4

I wouldn't mind this, if we can jump forward to OCaml 5 and forfeit any support for OCaml 4.x.

ivg:
@bmourad01 if you will have time, can you please rebase this on the master branch, I pushed some fixes to CI so that at least some of them should now start passing.

ivg:
Okay, the full build passes and this is good news! I will update the CI, but before merging there is an internal blocker on our side. We are feeding bap directly from the master branch, so I have to update our code base to v0.16. I will schedule it during this month.

PR #1615, opened 2025-02-27
relaxes x86 operands deconstructor to address llvm changes

ivg:
Occasionally, llvm refines their instruction definitions and change the number of operands, for example this [commit][1] added an extra destination register operand. For the old lifters we used a pretty strict rules for operands deconstructing and for the bts, btc, and btr instructions we were expecting two operands (now there are three of them).

I decided to relax the deconstructor instead of changing the definition of instructions, first to preserve backward compatibility with the older versions of llvm and second because many other instructions could be affected by the same changes from the llvm side, and I don't see any harm if we will accept instructions with one more argument, knowing that destinations are always prepended.

[1]: https://github.com/llvm/llvm-project/commit/fe96ff73983e01c2fcc7efa23d4513f7c5290611

### DIFF
--- a/.github/workflows/alpine.yml
+++ b/.github/workflows/alpine.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Build on Alpine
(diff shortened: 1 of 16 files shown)


ivg:
I am going to merge this and later investigate the issue with failing opam installation.

PR #1617, opened 2025-03-04
makes bap-extra packages optional to facilitate local development

ivg:
Eases local dune-base development, so that you can easily start local development without too much external dependencies hassle. Allows you to disable bap-extra packages and just do `dune build` and everything should work.

For the context, Dune doesn't provide nice options for multi-project monorepos. The dune files represent the static information about the project and it is presumed that there is an external tool that generates them. Cf., tezos manifest that generates dune files from the OCaml EDSL or whatever they use in Janestreet (jenga?). Originally, the plan was to implement the same mechanism, e.g., generate the dune-project file during the configuration phase. But right now I don't have enough time for that, so I am pushing a simple workaround. All packages in bap-extra are now marked as allow_empty, indicating that they can produce now artifacts. Their corresponding libraries are now optional (if a dependency is not met, then the library is silently not built). The only special case is ghidra that doesn't have dependencies trackable by Dune (i.e., OCaml libraries, besides base and bap), so to prevent it from building I added a small check that ghidra library is installed.


ivg:
Okay, the pipelines now look fixed, modulo macOS, which had never picked up a runner, so I had to cancel it.

PR #1618, opened 2025-03-05
attempts to fix the nightly and weekly tests

ivg:
as per a workaround suggested in https://github.com/ocaml/setup-ocaml/issues/678


ivg:
it didn't work though :(

PR #1622, opened 2025-03-12
adds semantics and supporting code for several x86 instructions

ivg:
Adds the following instructions:
 - shufps
 - pmuldq
 - [v]packusdw
 - [v]packuswb
 - [v]packssdw
 - [v]packsswb
 - vzeroupper
 - movntd
 - xchgb

The generated code for signed and unsigned saturation (the pack* family of instructions) is thoroughly optimized to minimize branching and comparison, which negatively affects analysis performance. Both signed and unsigned saturation are implemented as a lisp primitive that generates pure code with a single ite instruction and any comparisons are made with zero, i.e., no inequalities.

To following extra changes support the added instructions:

1) [NEW] Extend x86 CPU family definition with 8-bit registers, e.g.,
   AL,BL,CL,DL,AH,BH,CH,DH,SIL,DIL,BPL, and SPL.
2) [NEW] Added EFLAGS and RFLAGS aliases to the set of flag registers. 3) [NEW] It is now possible to use bool registers in the aliasing
   specifications.
4) [NEW] Added system and control register flags.
5) [FIX] Fixed a bug with incorrect reification of higher parts of
   aliased registers.
6) [NEW] Added a `dolist` macro to Primus Lisp.
7) [NEW] Added the `unquote` primitive to Primus Lisp. 8) [NEW] Added the `cast-saturate` primitive to Primus Lisp. 9) [NEW] Improved BIL code generation in case of empty if statements.


.github/workflows/main.yml
```diff
@@ -9,17 +9,13 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - macos-11
```
mjkoo, line 12: Was this intended to be replaced with `macos-latest` or is removal correct?

lib/bap_core_theory/bap_core_theory.mli
```diff
@@ -2218,7 +2230,7 @@ module Theory : sig
         register. *)
     val hi : ('a,sub) t -> int
 
-    (** [lo origin] returns the inclusive upper bound of the origin register
+    (** [lo origin] returns the inclusive lowe bound of the origin register
```
mjkoo, line 2233:
```suggestion
    (** [lo origin] returns the inclusive lower bound of the origin register
```

mjkoo (dismissed):

.github/workflows/main.yml
ivg: I finally gave up on building with macos, due to weird dependency issues in the 3rd party packages, so in the end I just temporary dropped macos from the CI/CD pipelines. Though, with the latest I managed to get further that with the 11-th version, which didn't event start a build.

PR #1632, opened 2025-07-03
Make compatible with OCaml 5 and Ghidra 11.4

PerfectLaugh:
This PR attempts to be an ongoing effort of https://github.com/BinaryAnalysisPlatform/bap/pull/1621

`core` library should have version tagged with it but I currently leave it blank.


PerfectLaugh:
This PR still contains several blocking issues:
1. Several PowerPC 32 tests not passed in `dune test` routine. (Probably LLVM disassembler problem)
2. RISCV-64 and PowerPC32 tests not passed in testsuite. (LLVM problem also)
3. Ghidra disassembler is somehow incomplete. For example, can make `aarch64-linux-gnu-echo` in `run` testsuite not exiting.

Rot127:
cc @ivg

Rot127:
@ivg Do you think you will have time to take a look?

PR #1635, opened 2026-01-03
fix incorect varible types after block flattening

Angs12:
Blk.flatten produces new variables with incorrect type after flattening expressions with extract or concat expressions.

On extract expressions the new variable gets the type of the expression inside the extract and not the size (high - low  + 1).

On concat expressions the new variable gets the type of the second expression and not the added size of the two expressions,
after the changes the type is the sum of sizes of the types of the expressions if the expressions are of type Imm otherwise if any of the expressions are of Unknown type the new variable type is also unknown. Lastly, i don't know how to define concat on memories so i call failwith.

### DIFF
--- a/lib/bap_sema/bap_sema_flatten.ml
+++ b/lib/bap_sema/bap_sema_flatten.ml
@@ -94,14 +94,28 @@ let flatten_exp
       Bil.Var var, insert blk def
     | Bil.Extract (n, p, x) ->
       aux x blk >>= fun (x, blk) ->
-      new_var (Var.sort_of_typ (get_direct_typ x)) >>= fun var ->
+      let res_typ = Type.Imm (n - p + 1) in
+      new_var (Var.sort_of_typ (res_typ)) >>= fun var ->
       let e = Bil.Extract (n, p, x) in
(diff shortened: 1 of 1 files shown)


lib/bap_sema/bap_sema_flatten.ml
```diff
@@ -94,14 +94,28 @@ let flatten_exp
...
+      let res_typ = Type.Imm (n - p + 1) in
+      new_var (Var.sort_of_typ (res_typ)) >>= fun var ->
       let e = Bil.Extract (n, p, x) in
       new_def var e >>| fun def ->
       Bil.Var var, insert blk def
     | Bil.Concat (x, y) ->
       aux x blk >>= fun (x, blk) ->
       aux y blk >>= fun (y, blk) ->
-      new_var (Var.sort_of_typ (get_direct_typ y)) >>= fun var ->
+      let exp_size (e : exp) : int option =
```
ivg, line 105: You can use `Type.infer_exn` here, together with `Type.sort`, but that also works for me

ivg (approved):

ivg:
Thanks, sorry for missing the PR!

